PR #14, opened 2017-06-02
Add various metrics to CRG

brian-brazil:
Given the recent discussions around CRG performance on FB, I figured it might help to add some data to the conversation.

This change adds all the standard Java/process metrics, some http metrics and basic websocket metrics.

Performance impact should be minimal, incrementing a counter/gauge is on the order of 10-15ns.

I would have added some more fun stuff like the current score, but I couldn't quite figure out where it's actually stored.


yetzt:
👍

jaredquinn:
Love it.  Great idea.
Merged into dev branch for wider testing.

PR #17, opened 2017-10-09
Black background around the SB mainDiv

riffreff:
moved style attribute of the 4 divs sourrounding the mainDiv to CSS.
background black instead of grey


AdamSmasherDerby:
This is a good change. Can we pull this?

PR #19, opened 2017-11-08
Updating README to remove old content and point to revised wiki.

rogerharkavy:
I revised the README so that the out-of-date instructions were removed and it includes a link to the current project wiki.

### DIFF
--- a/README
+++ b/README
@@ -1,3 +1,125 @@
-Instructions for installing the scoreboard software and getting started are available at the official wiki:
+# INSTALLING THE SCOREBOARD SOFTWARE
 
-https://github.com/rollerderby/scoreboard/wiki
+These are instructions for getting the software installed and running on a standalone computer to provide a functioning scoreboard. Once you complete them, you can go to the Quick Start Guide to Operating the Scoreboard below or browse the topics on the [Scoreboard Wiki Main Page](https://github.com/rollerderby/scoreboard/wiki/) for more detailed information about scoreboard procedures.
+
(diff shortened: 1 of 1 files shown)


AdamSmasherDerby:
I don't think the README should be THIS empty.  Even though it wouldn't dynamically update, maybe the contents of the "Installing the Scoreboard" page on the Wiki would be good to have here?

brian-brazil:
Avoiding duplication is generally best, but it does seem a bit bare. Maybe point to the facebook groups, and include a screenshot so users know they've found the right project?

AdamSmasherDerby:
I think we ought to have at least the basics of "how to get this running," included with the download package.  A lot of the time, this is being run by the bench coach's brother-in-law who isn't super comfortable with computers, and they're at the West Shelbyville Rollerama where the internet access isn't great.  I think having a little more than just links would probably be the kind thing to do - it's not as though the basics of "how to start CRG" change all that often.

xrobau:
I agree, I think we need better docs.  However, I'm happy to write them. How about we merge this, and I'll write some docs now?

AdamSmasherDerby:
Just don't feel like you have to rewrite them from scratch!  The wiki may not be perfect, but there's quite a bit of stuff there to be used.

rogerharkavy:
I'll work on this, I agree with the idea of duplicating the quick start guide because the GitHub wiki doesn't appear when you're viewing it on a mobile device. We'll just have to be diligent about updating the content in both places.

brian-brazil:
How about storing it in markdown in this repo so github can render it, and also including that with releases?

rogerharkavy:
Okay, updated readme with markdown text from the wiki, let me know what you think.

PR #20, opened 2017-11-08
Removed link to old mailing list. Old information about graphics will…

rogerharkavy:
… be moved to the wiki.

### DIFF
--- a/html/guide.html
+++ b/html/guide.html
@@ -11,40 +11,7 @@
 </a>
 
 <h2>ScoreBoard Wiki</h2>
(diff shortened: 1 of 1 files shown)


AdamSmasherDerby:
In a similar vein from my comment on the readme, I think pasting the "quick start" guide in here might be useful.  A lot of scoreboard situations occur in places without simple internet access from the computer.

rogerharkavy:
I think having the content in three places is pushing it. Let me work on this, though, I have an idea.

rogerharkavy:
Updated the guide page with more information, let me know what you think.

PR #22, opened 2017-12-15
Improve Timeout and Clock handling:

frank-weinberg:
- Distinguish between Official Timeout and unspecified Timeout (declare OTO by
re-pressing the "Timeout" button)
- Go from timeout to lineup with the "Stop Jam/TO" button
- Restore type of Timeout when returning to Timeout with "Un-*" button
- Highlight the Team-T/O  resp. OR buttons when the respective type of
timeout is active
- Remove the Start, Stop, Set, and Reset buttons from the individual
clocks as they should not normally be used.

### DIFF
--- a/html/controls/mobile.html
+++ b/html/controls/mobile.html
@@ -119,7 +119,7 @@
 				<button class="StartJam" data-role="button" data-icon="check" data-iconpos="right">Start Jam</button>
 			</div>
 			<div class="ui-block-b">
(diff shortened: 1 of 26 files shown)


jaredquinn:
I want to do some testing on this one before I merge into the release for the new penalty codes.
Is there anything I should know?

frank-weinberg:
Any behaviour change not mentioned in the commit messages is probably unintentional.

PR #24, opened 2018-01-20
Update penalty codes to match 2018 verbal cues

official-sounding:
prompts for the codes should fully cover all standard verbal cues.


AdamSmasherDerby:
We should pull this sooner, rather than later, and probably put out the word that we've done so, since one implication of the "D" code is that this is now the *only* penalty tracking software in existence.

jaredquinn:
Merged in ready for a release before end of February to support the changes.

jaredquinn:
Ideally we'd pull these out of code and move them to a configuration; ideally the ruleset editor.

PR #35, opened 2018-03-19
Rename README to README.markdown

rogerharkavy:
Since the content for this file is in markdown, the filename extension should reflect it.


brian-brazil:
Isn't it usually `.md` ?

rogerharkavy:
I researched it and there really isn't an agreed-upon extension for markdown, but .markdown is one that many editors and file managers (and GitHub) will recognize. Plus, .md is sometimes used by another file format in Linux.

brian-brazil:
I've personally never seen .markdown, it's always been .md on github and elsewhere.

rogerharkavy:
You can use either one (see Wikipedia), but on Linux machines, .md can also be a machine description file. I chose .markdown because it was less ambiguous and would be recognized by editors and GitHub. If there's a consensus to use .md instead, I'm ok with changing it.

rogerharkavy:
Accidentally closed this, sorry. Can it be reopened?

jaredquinn:
I can't re-open because the repository it was pulled from has been closed.

rogerharkavy:
@jaredquinn I think there was more than one pull request that renamed README to README.markdown

Issue #27, opened 2018-02-26
Outdated text on the start page.


rogerharkavy:
If the scoreboard script isn't running, start.html shows out-of-date information.

rogerharkavy:
I'll test this in 3.9.3.

rogerharkavy:
I tried it and when the script is running, the page has out-of-date information and a link to the old SourceForge site. I will fix it.

The release notes page (release.html) also needs to be updated, I defer to the folks writing the code but I can assist with that if needed.

rogerharkavy:
Ok, I updated guide.html and index.html, forgot to associate them with this issue when saving, though. Sorry. Hope that doesn't screw anything up.

PR #36, opened 2018-03-19
Made consistent with index.html.

rogerharkavy:
Just reformatted intro text. Addresses #27

### DIFF
--- a/html/guide.html
+++ b/html/guide.html
@@ -10,7 +10,7 @@
 </center>
 </a>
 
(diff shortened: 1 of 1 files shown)


yetzt:
why would you replace a semantic heading with a deprecated-bold text in a paragraph?

rogerharkavy:
Force of habit. Feel free to change it if you think it should be done differently.

rogerharkavy:
Changed the HTML as per the suggestion.

Issue #48, opened 2018-05-02
REQUEST: Add button to main interface to restart period clock and lineup clock after an OTO, where there are <:30s left on period clock.


voidptr:
In prior releases, it was possible to automate this process by binding the old stop TO, start lineup, start Period clock buttons to a single key. Now that these buttons were removed, this is no longer possible.

This is a business rule that is required by the rules of roller derby, and it happens often enough that we need a proper button for it. 

Maybe locate it somewhere nearby to where we put the new endTO/startLineup button?

frank-weinberg:
Instead of adding a new button it should be possible to add a detection for this situation to the logic applied for the end TO button. This should probably be configurable in the rules editor.

As a workaround making the start/stop buttonsfor the individual clocks accessible again would solve the immediate problem. But these buttons should not be needed during normal gameplay.

AdamSmasherDerby:
The requirement to restart the period clock again after a timeout is ONLY necessary when there are 30 seconds remaining on the period clock and there would *not* have been another jam.  If there *would* have been another jam, there's no necessity to restart the period clock.

So the lineup clock is irrelevant here - the period clock controls in all cases.

However - we DO need a way to simultaneously end the timeout and restart the period clock.

AdamSmasherDerby:
For now, best practice would seem to be 
1. Click the timeout clock.
2. Clock Stop
3. Click the period clock.
4. Click start.

This is something that should be fixed for sure, but at least there's the steps for making it happen as it stands now.

JeneralPain:
The team in the development crew have discussed this at length, given the potential impacts specifically to high level sanctioned play where this is a requirement, we have tagged this as an urgent issue. We will do our best to release a patch/update to this as quickly as possible.

Thanks

JP - Issues Management Team

voidptr:
The lineup clock aspect of this is actually VERY IMPORTANT. Consider the following pair of scenarios:

Scenario 1: Jam ends, there's enough time for another jam (35s on the period clock). Period clock ticks down to 15, lineup clock is at 20. JT is getting ready to say 5s. HR blows OTO. Time passes. OTO ends, period clock restarts (because there are less than 30s), lineup-clock does not continue from where it was. Everyone is confused, because there are now less than 30s to start the next jam. 

[...]

PR #49, opened 2018-05-03
Add switch to show/hide start and stop buttons for individual clocks.

frank-weinberg:
Workaround for Issue #48 (and probably other unreported issues)

### DIFF
--- a/html/controls/operator.js
+++ b/html/controls/operator.js
@@ -166,6 +166,15 @@ function createMetaControlTable() {
 			$("tr.SpeedScore").toggleClass("Show", this.checked);
 		});
 
+	$("<label>").text("Show Start/Stop Buttons").attr("for", "ShowClockControlsButton")
+		.appendTo(buttonsTd);
+	$("<input type='checkbox'>").attr("id", "ShowClockControlsButton")
+		.appendTo(buttonsTd)
+		.button()
+		.click(function() {
+			$("tr.Control").toggleClass("Show", this.checked);
+		});
+
 	$("<button>").attr("id", "GameControl")
 		.text("Start New Game")
 		.addClass('clickMe')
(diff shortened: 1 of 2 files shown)


brian-brazil:
Code itself looks fine. I know that the start/stop buttons are for advanced users and have caused confusion for new users, but I'm not sure it's worth another toggle in the top set.

brian-brazil:
This makes the team colour control row on the Teams tab also controlled by the new toggle. I propose we keep it simple, always show the start/stop, and don't add the new toggle.

frank-weinberg:
Fix is in. I also fixed the other show/hide controls, which would also have controlled elements of the same name outside of the tab (if any had been added).

html/controls/operator.js
```diff
@@ -166,6 +166,15 @@ function createMetaControlTable() {
 			$("tr.SpeedScore").toggleClass("Show", this.checked);
 		});
 
+	$("<label>").text("Show Start/Stop Buttons").attr("for", "ShowClockControlsButton")
+		.appendTo(buttonsTd);
+	$("<input type='checkbox'>").attr("id", "ShowClockControlsButton")
+		.appendTo(buttonsTd)
+		.button()
+		.click(function() {
+			$("tr.Control").toggleClass("Show", this.checked);
```
brian-brazil: Can this not affect tr.Controls on other tabs?

frank-weinberg: Quick check shows that it does.
Since I copied this code from the other show/hide buttons, the problem extends to all of them. (Though the other ones affect elements with less generic class names.)

frank-weinberg: Restricted the search to the #TeamTime tab for all three buttons.

(At my current rate of learning, I can add jquery and css to the list of skills in my CV before the end of the year.)

PR #55, opened 2018-05-12
fix month in game filename

official-sounding:
months in js are set/get in c-style (eg 0-11), so previous filenames have an off-by-one error when saving.  Having the correct month in filenames will likely be important when resuming existing games.


yetzt:
whats wrong with `date.toISOString()`? it even has the time zone.

brian-brazil:
We probably need to redesign the way this entire JSON file works, however this is a simple bugfix so :+1: from me.

Issue #57, opened 2018-05-20
Increasing lag by the end of the game.


AdamSmasherDerby:
Multiple WFTDA tournaments have reported that the clocks have been lagging to a greater and greater degree over the course of a game.  By the end of the second half, lag can accumulate up to as much as several seconds.  This is obviously a critical issue for a piece of timing software.  I'll attach reports below from several tournaments.

It appears that throwing CPU speed at the problem is presently the only thing that improves the situation.

### Udder Chaos

Running CRG on an i5 laptop, scoreboard operator is on the same computer.  This computer is dedicated to CRG - it has nothing else installed (running Win 10).

The overlay computer and the scoreboard projection computer are both remote and show lag.

Also have 4 computers showing the electronic whiteboard.  

Restarting CRG before each game, clearing the chrome cache at the same time as the restart.

The CRG scoreboard is lagging.
It doesn't lose time on the game clock, but it doesn't start the lineup clock until after it finishes lagging.

This is adding from 2-11 seconds to lineup times.

-

First half nearly over with CRG on Linux Mint, issue is still present. What we are seeing that is the most concerning is that after a jam ends, it takes progressively longer for the lineup clock to start. We've been keeping track, and it's between 45 second and 2 minutes of game time that is run off the clock that shouldn't be - this could definitely have an effect on the outcome of a game. 

Previous game we tried only loading in the two teams we were using, no change in the behavior.

-

The last game on Windows and this game on Linux both lost 45 seconds of game time due to the lineup clock not starting - so Mint definitely isn't the magic bullet.

-

For our first game today we used a Linux Mint live USB to boot a core i7 box with 24 gb ram (retail price US$1899 for this machine). Things were pretty good until the 20 to 25 minute mark of the second half, where we started losing about 2 seconds after each jam for 75% of jams. 

Still not great, but at least we're not losing 2 minutes a game.

Also have projector, overlay, and operators computers all running separately (plus the 3 whiteboards).

-

One more note, on the i7 box we are seeing CPU usage spike to ~60% as a jam ends, when the lineup clock is supposed to start (on the core2 duo we tried first it would spike to 90%). After the CPU usage drops back down, the clocks are good.

### Euroclash

Setup: CRG 3.9.3a running on Linux Mint 18.3, i7 8gb ram laptop, connected by ethernet to the private network, and to the projector. There are digital whiteboard displays in both team benches, the penalty box, THOs table, and announcer table.

Issues:
[...]

PR #59, opened 2018-05-21
Fix lag and persist penalties

brian-brazil:
This is only a partial solution. To complete it the penalty logic should be implemented in the same way everything else is and all `update` calls in `game/*.java` removed, with the relevant penalty and penalty update logic moving to the `DefaultSkaterModel`.

This PR also persists penalties.


brian-brazil:
Fixes #57

frank-weinberg (commented):
I'm not convinced your removal of the argument to sendUpdates is the correct fix.

If I understand the code correctly, there are three key sets:
1. The set of all keys, aka state.keySet()
2. The set of updated keys, aka the removed argument keys
3. The set of keys that the client has registered, aka Conn.paths

The original code checked 1 against 2, which is bad. The new code checks 1 against 3, which is better.
The correct solution would be to check 2 against 3.

Also this change has the potential to uncover previously hidden bugs in the clients: Unless there is another filter down the line, the old code sent all updates to all clients, while the new code only sends updates the client has registered. If a client forgets to register a path it needs, this would be hidden by the old code but result in missing updates with the new code.

brian-brazil:
> The correct solution would be to check 2 against 3.

That may be better, but it requires a rewrite of how this entire data flow works as 2 doesn't exist currently. Keep in mind that `path` was a prefix, not an exact path. This is something that could be done in the future, but we have good confidence the current proposed code is good enough and it has been tested.

> If a client forgets to register a path it needs, this would be hidden by the old code but result in missing updates with the new code.

This is a possibility, however I didn't spot this in any of the code in the screens and none of the alpha testers reported issues.

frank-weinberg:
2 does exist. It is in the removed argument `keys`. `WS.updateState()` (Sidenote: That method is also horribly inefficient.) makes sure that every element of `state` that is modified is added to that set before passing it to the connections. There should be no need to send subkeys of updated keys as they will be unchanged.

That being said, I am on board with getting this improvement shipped (assuming there are no severe issues with the later commits - I haven't reviewed them yet). We just should not consider the issue completely fixed yet and should have the possible uncovering of bugs in mind in case any reports come in.

brian-brazil:
That change is just one part of the fix, the other is massively reducing the amount of data this code is handling in the first place. Between them it should be pretty hard for the performance to regress here.

frank-weinberg (commented):
One more minor issue: Where the old code is indented with 1 tab per level, newly added code is indented with 2 spaces per level.

In terms of content everything should be good.

frank-weinberg:
And I agree that reducing the size of updates removes the impact in terms of efficiency. But having analyzed the code, I now have a strong urge to make it  a. more efficient and b. easier to follow. I'll probably put in some effort to that end after the release is out.

brian-brazil:
The existing code style is quite hard to work with, given the breadth of the changes I gave up trying to match it. I'm planning on running a formatter over it at some point, the number of brace-less if/for statements is a risk.

jaredquinn:
@brian-brazil  yeah some of the core code is 10+ years old.  There's a plan to tackle that but no time.

PR #60, opened 2018-05-29
Overhaul Clocks and move Policies to Rulesets

frank-weinberg:
Bugfixes:
- Correctly handle timeouts close to (issue #48 ) and after period end.
- Settings in the Policies were reset on restart and when starting a new game. This has been fixed by moving them into the rulesets. Some of these would not work properly, when the direction of a clock was switched. This should be fixed as well.

(Immediately) user visible changes:
- Operator Screen:
  - The Un-Start, Un-Stop and Un-Timeout button have been replaced by a common Undo button that can also undo starting Lineup for an overtime Jam. The label on the button indicates, what can currently be undone.
  - When pressing Un-Timeout the number of Team Timeouts and Official Reviews is reverted to the value from before starting the Timeout.
  - The label of the Jam Control buttons (Start Jam, Stop Jam/TO, and Timeout) also changes based on the respective button's current function.
  - The label on the buttons that start/select a Team Timeout has been changed from "Timeout" to "Team TO"
  - The policy Tab has been removed. All settings from the Policies that were still relevant (som affected code that has been removed in earlier versions) can now be found in the rulesets. (Quite a few of them were already there.)

- Ruleset editor screen:
  - Show the description of a rule as a tooltip when the mouse is over the rule's name.
  - Make sections foldable and fold them by default.
  - Order sections and rules within section as they appear in the code. Sort them in the code such that more commonly changed rules appear near the top.
  - Some new settings have been introduced in addition to those moved from Policies:
     - A setting to control which clock is shown between the end of a timeout and the start of the next Jam.
     - Settings that control the number of Team Timeouts and Official Reviews per period or per game.
     - Settings that allow to stop the period clock only on some types of timeouts (or none at all) or to stop it only, if a timeout lasts longer than a given time.
     - Settings that control the behaviour of the Undo button (see below for details).

Internal changes and changes that will only be user visible after a configuration change:
- There is an Undo buffer for user induced clock changes. The size of the buffer can be configured via rulesets with a default of 1 (matching the behaviour of previous versions).
- The Undo button will be deactivated after a time configurable via rulesets. The default value is 1 hour (effectively matching the behaviour of previous versions).
- All code that affects the clocks is now in the class DefaultScoreBoardModel and this code has been overhauled for readability.
- The ClockModel has been given convenience functions to deal with elapsed time indepent of the direction of the clock (count up or down) and to increase number, resetTime and start in order to serve the aforementioned readability.
- The Settings class also got convenience methods that return settings converted to a specific data type.
- All code dealing with Policies has been removed.

Rationales:
- Changing the way Undos are handled was done because the code for the Undo functions was getting severely complicated and thus very hard to maintain in order to handle all the special cases. 
- This triggered a general overhaul of the clock code with the aim of read- and maintainability.
- Moving clock related Policies was done as part of cleaning up the clock handling code.
- Moving all Policies was then done as there was only one left that was not dealing with clocks.
- Making the labels on the Jam control buttons dynamic was done in order to address concerns that buttons with context sensitive functionality can be confusing to users. This problem would have been especially big with the newly introduced Undo button.


frank-weinberg:
Following discussions in the Facebook Dev group, the Undo behaviour has been modified. I have updated the description above accordingly.

PR #64, opened 2018-05-31
Create Black and White teams by default

### LINKED ISSUE #61: Releases should include empty Black and White team
With "Start New Game" becoming effectively required, having to go and create two new teams to get going is annoying - particularly if you don't know how to use the scoreboard already.

I suggest that two empty teams be included out of the box (Black and White, as that's common for scrimmage and unranked games) to make this easier for users who are getting started.

brian-brazil:
If there's no autosave directory, start out with a
black and white team to make it easier to get
a game going from a fresh install.

Fixes #61

### DIFF
--- a/config/default/15-teams.xml
+++ b/config/default/15-teams.xml
@@ -68,5 +68,15 @@
       </Position>
     </Team>
   </ScoreBoard>
+  <Teams>
+    <Team Id="Black">
+      <Name><![CDATA[Black]]></Name>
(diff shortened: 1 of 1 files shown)


frank-weinberg (approved):
Quick test shows that the fix works. Only side effect is that there are two extra teams on the dropdown. Maybe we should add a sentence to the documentation that they can be safely removed after the user has entered their own teams.

PR #71, opened 2018-06-06
Fix errors from merging incompatible branches

frank-weinberg:
This fixes the compile errors in the java code. I did not have time to check if there are other errors, especially in the frontend code.


brian-brazil:
I've fixed up the broken merges, so this is obsolete now.

PR #73, opened 2018-06-08
Display expulsion on whiteboard as red line, same as 7+ penalties

brian-brazil:


official-sounding:
Looks good to me.

brian-brazil:
Thanks!

PR #77, opened 2018-06-15
Set background to plain black by default

frank-weinberg:
(That huge list of commits is a side effect of my attempt to rebase my local repository to the rebased origin.)


brian-brazil:
Nothing a squash can't handle!

Issue #72, opened 2018-06-06
Add unittest infrastructure


brian-brazil:
We currently have no unittests, and there's a variety of places (e.g. around clock start/stop) that we would benefit from them. It seems that Junit works with Ant, so getting that working would be a good place to start: https://ant.apache.org/manual/Tasks/junit.html

PR #78, opened 2018-06-17
Fix Box button not working after Star Pass

### LINKED ISSUE #5: Star Pass + Penalty causes Box button to not reset
Reported by RH on Facebook:

Jammer and Pivot selected from drop downs, star pass occurs, the pivot (now jammer) goes to the box, box button is pressed for the pivot. Jam ends, pivot is "promoted" to jammer and the red background on the box button follows as expected, after the new jam begins the jammer is released and the box button pushed to clear them from the box - and here's the bug - the red highlight/background on the box button for the jammer doesn't clear.

If the jammer is set to "no jammer" it'll clear the red highlight on the box button. I've also verified that pushing the box button clears the skater from the box on the scoreboard assistant screen (even though the red highlight remains on the operator screen.)

official-sounding:
When the skater changes positiosn from "Pivot" to "Jammer", the state is not correctly sync-ed between the PositionModel and the SkaterModel.  At Jam end, the Jammer position's PenaltyBox status is cleared, but then not re-set based on the SkaterModel.

This is now "fixed", but this PositionModel/SkaterModel nonsense is a mess, and in an ideal world I would just maintain the state on the SkaterModel and fire Position updates if the skater's status changes.  I would be more comfortable making those changes once unit tests are in place though (#72).

Closes #5


brian-brazil:
Thanks for figuring this out, though I've not been fully through this bit of the code. The duplication of PositionModel/SkaterModel does sound messy.

I'm wondering if it's worth making this our first unittest.

official-sounding:
I went ahead and added some "unit" tests covering base PositionModel functionality, and this scenario in particular.   I also introduced Mockito as a test dependency.

Despite the mocking, I end up having to make heavy use of TeamModel and SkaterModel in them to really test the position model functionality correctly, but it will at least start getting coverage to eventually allow for some separation.

src/com/carolinarollergirls/scoreboard/defaults/DefaultPenaltyModel.java
```diff
@@ -0,0 +1,27 @@
+package com.carolinarollergirls.scoreboard.defaults;
+
+import com.carolinarollergirls.scoreboard.Skater.Penalty;
+import com.carolinarollergirls.scoreboard.model.SkaterModel.PenaltyModel;
+
+public class DefaultPenaltyModel extends DefaultScoreBoardEventProvider implements PenaltyModel
```
brian-brazil: Why did you move this out of the SkaterModel? Given PenaltyModel is a subinterface of SkaterModel I'd expect that to stay where it is.

tests/com/carolinarollergirls/scoreboard/DefaultSkaterModelTests.java
```diff
@@ -0,0 +1,14 @@
...
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+
+public class DefaultSkaterModelTests {
+
+	@Test
+	public void test() {
+		fail("Not yet implemented");
```
brian-brazil: Will this cause the tests to fail?

brian-brazil (commented):
This looks generally good.

One thing I want to do at some point is hook things up so we can test the event state is right too.

official-sounding:
Sorry, didn't mean to commit the skater model test file, obviously wasn't completed yet.  ditto splitting skatermodel/penaltymodel, and I'm cool keeping it where it is anyways.

after fixing those, I realized there were a few issues with build.xml, so I fixed those as well.

brian-brazil:
Great, thanks!

PR #79, opened 2018-06-17
Add Unit Test infrastucture

### LINKED ISSUE #72: Add unittest infrastructure

official-sounding:
Add ability to run jUnit 4.x tests to both the ant file and the eclipse project.  Also add a simple example unit test that ensures that crg objects can be accessed.

Test compilation goes to it's own top-level bin directory so that the test classes don't get included in the main release, and I didn't set the test task as on the required path for an ant build or release.  Once we have meaningful unit tests, we can decide if passing tests are a requirement for PR merging/releases/etc.

Closes #72


.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,7 +1,12 @@
 eclipse.preferences.version=1
 org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.methodParameters=do not generate
```
brian-brazil: Are these changes related to this PR?

.project
```diff
@@ -26,4 +26,15 @@
 		<nature>org.eclipse.jdt.core.javanature</nature>
 		<nature>org.eclipse.wst.jsdt.core.jsNature</nature>
 	</natures>
+	<filteredResources>
```
brian-brazil, line 29: Are these changes related to this PR?

brian-brazil (commented):
This is great! I'd make tests required for at least release, so we don't forget to do it later.

.project
official-sounding: Yes, this filters the test-bin directory from view in eclipse

official-sounding:
removed the changes to the file in .settings with a force-push.  I think they were added inadvertently while debugging on my other PR.

build.xml
```diff
@@ -37,6 +39,13 @@
...
 
+	
+	 <path id="classpath.test">
+	    <pathelement location="test-libs/junit-4.11.jar"/>
+	    <pathelement location="test-libs/hamcrest-core-1.3.jar"/>
+	    <pathelement location="${dest.dir}"/>
+	  </path>
+	
 	<!-- Show help (the different options for this build.xml) -->
 	<target name="help">
```
brian-brazil: This needs updating too

brian-brazil (commented):
Can you make this a dependency of the release and we can get this in?

official-sounding:
cleaned up the tabs/spaces issues I introduced, made the test task a pre-req of the "zip" step that is used to make releases

brian-brazil:
Thanks!

PR #80, opened 2018-06-19
Restrict key controls to the Team/Time tab

frank-weinberg:
The commit that made hidden key controls work, unintentionally also enabled them when a different tab is selected (e.g. entering teams). This fixes the unwanted behaviour and restricts the key controls to the Team/Time tab again.


brian-brazil:
Thanks!

PR #81, opened 2018-06-22
Add "OR" Button to Mobile Jam Control, Add Timeout Editor Page to Mobile Controls

### LINKED ISSUE #41: REQUEST: Add Official Review to Mobile Jam Control
Currently the when control the Clocks from the Mobile Jam Control, you can't enter Official Review for either team

Recommendation is to add an additional pair of buttons under the existing Time Out buttons for each team, where the red boxes are in the screenshot below

screenshot_27

official-sounding:
Added OR Buttons to the Mobile Jam Control.  Also added a new view to the Mobile Controls that allows the mobile user to edit the number of timeouts and official reviews each team has, without having to open the operator screen.  

This view is very rudimentary, mostly because jquery-ui grid layouts remain a profound mystery to me.  Someone more layout-oriented could probably make this look a bit better.  But for now i think it will work.

Closes #41

### DIFF
--- a/html/controls/mobile.html
+++ b/html/controls/mobile.html
@@ -33,9 +33,6 @@
 			<li><a href="#JamControlPage">Jam Control</a></li>
 			<li><a href="#PeriodTimePage">Period Time</a></li>
 			<li><a href="#TeamScorePage">Team Score</a></li>
-<!-- hide link for next release, then work on using on-server clocks
-			<li><a href="#PenaltyPage">Penalty Timing</a></li>
--->
 		</ul>
 	</div>
 </div>
@@ -130,7 +127,7 @@
 				<span class="Team1 AlternateName"></span>
 				
 			</button></div>
-			<div class="ui-block-b"><button class="Official">Official</button></div>
+			<div class="ui-block-b"><button class="Official">TO Official</button></div>
 			<div class="ui-block-c"><button class="Team2">
 			TO 
 				<span class="Team2 Name">Team 2</span>
--- a/html/controls/mobile.html
+++ b/html/controls/mobile.html
@@ -171,14 +171,58 @@
 
 	<div data-role="content">
 		<p class="Time"><a class="Time"></a></p>
-		<p><hr/></p>
+		<hr/>
 		<p>
 			<button class="TimeDown" data-inline="true" value="-1000">-1 sec</button>
 			<button class="TimeUp" data-inline="true" value="1000">+1 sec</button>
 		</p>
-		<p><hr/></p>
+		<hr/>
 		<p><input type="text" class="SetTime"/></p>
 		<p><button class="SetTime" data-inline="true">Set Time</button></p>
+	</div>
+</div>
+
+<div data-role="page" id="TeamTimeoutPage">
+	<div data-role="header">
+		<a href="#Main" data-direction="reverse" data-icon="arrow-l">Main</a>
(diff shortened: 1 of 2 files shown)


html/controls/mobile.js
```diff
@@ -15,7 +15,8 @@ $sb(function() {
 	setupJamControlPage();
 	setupPeriodTimePage();
 	setupTeamScorePage();
-	setupPenaltyTimePage();
+	//setupPenaltyTimePage();
```
brian-brazil: Why is this commented out?

html/controls/mobile.html
```diff
@@ -125,20 +125,41 @@ <h1>Jam Control</h1>
 		<button class="Timeout" data-role="button">Timeout</button>
 		<div class="ui-grid-b Timeout">
 			<div class="ui-block-a"><button class="Team1">
+				TO 
 				<span class="Team1 Name">Team 1</span>
 				<span class="Team1 AlternateName"></span>
+				
 			</button></div>
 			<div class="ui-block-b"><button class="Official">Official</button></div>
```
brian-brazil: Can it be made clearer this is an OTO rather than a OR?

html/controls/mobile.js
official-sounding: the UI side of the penalty Timing page is already commented out, wanted to prevent any of the js side of initializing as well.  I'm happy to remove the line instead of commenting out if you'd prefer.

brian-brazil: I'd prefer it removed, we can always get it back from source control.

html/controls/mobile.html
```diff
@@ -125,20 +122,41 @@ <h1>Jam Control</h1>
 		<button class="Timeout" data-role="button">Timeout</button>
 		<div class="ui-grid-b Timeout">
 			<div class="ui-block-a"><button class="Team1">
+				TO 
 				<span class="Team1 Name">Team 1</span>
 				<span class="Team1 AlternateName"></span>
+				
 			</button></div>
-			<div class="ui-block-b"><button class="Official">Official</button></div>
+			<div class="ui-block-b"><button class="Official">TO Official</button></div>
```
brian-brazil, line 130: I've already heard this as "official time out" rather than "time out official". Is this one of those things that varies across leagues?

official-sounding: Official Timeout is definitely the preferred nomenclature, but using "TO Official" was more about keeping the buttons consistent with "TO Team 1" and "TO Team 2".  Also when the buttons are small, the text is truncated, so I wanted to make it clearer it was a TO instead of an OR (so it shows up as "TO Off..." instead of "Offici..." or the like)

brian-brazil:
Thanks!

html/controls/mobile.html
```diff
@@ -150,17 +168,61 @@ <h1>Period Time</h1>
...
 			<button class="TimeUp" data-inline="true" value="1000">+1 sec</button>
 		</p>
-		<p><hr/></p>
+		<hr/>
 		<p><input type="text" class="SetTime"/></p>
 		<p><button class="SetTime" data-inline="true">Set Time</button></p>
 	</div>
 </div>
 
+<div data-role="page" id="TeamTimeoutPage">
```
brian-brazil, line 182: Just testing this now, and I noticed that this doesn't appear on the mobile control index.

PR #83, opened 2018-06-27
Add File-based Runtime Loaded peanlty file

### LINKED ISSUE #53: Move penalty definitions into a runtime-loaded file
This issue tracks leftover work from #24 and #45/#52.  I am planning on working on this, but just wanted to give it visibility.

Rather than having penalty information hard-coded into ScoreBoardJSONListener, there should be a separate file that contains penalty definition information, like codes, verbal cues, etc.  This file should load when the scoreboard starts up, so that if/when verbal cues change, the file can be updated by operators without waiting for and testing a new release.

I also sort of want to rework the way penalties data is loaded into the PT sheet, but may leave it as a part of the WebSocket firehose for now.

official-sounding:
Uses a Ruleset definition to point to a file in the config directory that defines a list of penalties.  Adds an endpoint in the JSON Servlet that supplies that file as a string to the frontend, and overhauled the pt & pt (color) views to use a direct ajax call (easily chachable later on) to load the file.  Defaults to the 2018 penalty file, and includes the previous version for posterity.  Added a few small unit tests around the SettingsModel.

Rationale for approach vs text string, etc:

1. The WS-based delivery of penalty codes is over-complicated, the front-end needs to handle receiving repeated codes, etc.  hitting an endpoint to download a file is simpler and allows for potential caching
2. using a JSON file to define the penalty codes makes it easier to edit than a text delimited string, especially when considering defining multiple penalty verbal cues per code, whether a penalty is expel-able, etc.
3. Distributing an update to penalty codes/verbal cues becomes easier, as a user just needs to drop the file and repoint, instead of updating a potentially error-prone encoded text string in the ruleset editor.  My sense is that OffCom may make changes to verbal cues on a more frequent basis.
4. Allows users to maintain multiple sets of penalties (say, during beta testing of 2019 rules) more efficiently, and potentially allows WFTDA or other governing bodies to distribute new penalty codes more efficiently.

Closes #53


frank-weinberg:
I find that rationale convincing (and hope to have time to review over the weekend).

brian-brazil:
> The WS-based delivery of penalty codes is over-complicated, the front-end needs to handle receiving repeated codes, etc. hitting an endpoint to download a file is simpler and allows for potential caching

Adding an entire 2nd way in which configuration gets to some pages is also complicated. I'd much prefer we stick to exactly one way of getting data to WS endpoints. We already have enough data transfer mechanisms without adding yet another one. The data sizes involved are also tiny, so caching is not relevant.

I also suspect this implementation is vulnerable to a directory traversal attack.

> whether a penalty is expel-able,

For simplicity all penalties should be considered expelable. The rules have a pretty broad interpretation, so best to avoid getting stuck in that debate.

> Distributing an update to penalty codes/verbal cues becomes easier, as a user just needs to drop the file and repoint, instead of updating a potentially error-prone encoded text string in the ruleset editor. 

I'm not convinced by this. Our users seem to have difficulty uploading files as-is for team logos and ads. Now they need to figure out how to do that and change the ruleset anyway.

> Allows users to maintain multiple sets of penalties (say, during beta testing of 2019 rules) more efficiently, and potentially allows WFTDA or other governing bodies to distribute new penalty codes more efficiently.

I don't see how this changes this. It's a new ruleset one way or the other.

frank-weinberg:
> Our users seem to have difficulty uploading files as-is for team logos and ads. Now they need to figure out how to do that and change the ruleset anyway.

Users can grab logos and ads from literally anywhere. New sets of penalty codes would have to be distributed from someone who knows how to prepare them and can thus include instructions how to install them. Plus we can add multiple sets of codes with our distribution and the user can choose from those without difficulty.

> I also suspect this implementation is vulnerable to a directory traversal attack.

Giving the file name as a string is also prone to files not being found due to typos. The ruleset editor should only allow the user to select from the detected set of files (presumably through a dropdown).
However, we currently don't support rules that allow the choice from a fixed set of options, except for binary choices. So we would need to add this type of rule (which would probably be useful in other places as well).

(I'm keeping out of the debate about data transmission between backend and frontend as this is not my area of expertise.)

brian-brazil:
> Plus we can add multiple sets of codes with our distribution and the user can choose from those without difficulty.

That argument makes sense, and applies for any approach we come up with.

> The ruleset editor should only allow the user to select from the detected set of files

It's important that this sort of thing be a backend check. Security is not that important for the scoreboard currently, but allowing anyone with access to the scoreboard to poke around the files of what is likely a personal laptop seems unwise. The safest way to handle this would be to use something that filters out/ignores any ../ in paths.

frank-weinberg:
> That argument makes sense, and applies for any approach we come up with.

It's at the very least a lot harder to do, if the set of penalties is stored as one or more StringRules. 

> The safest way to handle this would be to use something that filters out/ignores any ../ in paths.

I would use the more generic approach of having a rule type that presents a set of options and checks any selction from the frontend against these options. (And the set of options is obviously populated in the backend.)

official-sounding:
I agree that file uploads pose difficulties to users, but I agree with frank-weinberg that using some format stored in a StringRule is going to be harder to update in practice for the set of folks who would be interested in updating penalty codes.   

Using a format like JSON over something custom gets us the benefit of being able to use external editors for validating syntax, and using a file reference instead of the contents of a string rule means that we avoid a class of errors (copy-and-paste mishaps leading to corrupted penalty definitions), let users/ruleset maintainers store the penalties in an easier to edit file instead of a one-line string, and (more weakly) allows for more consistency between installs.  Updating a file location is (in my opinion) easier for someone to update than putting all the definitions directly into a ruleset item.  Users could also tell much more simply what set of penalties a given ruleset is actually using, especially when there are subtle differences; thinking here of JRDA, which uses all of the WFTDA penalties, plus one additional one.  It might be hard to see if that penalty is present from the ruleset editor if the penalties are all in a string.

From there, actually delivering the definition to the front-end, I understand the trepidation pulling this out of the WS feed, but I think keeping them in the WS feed adds significantly more complexity than delivering a standalone file.  

We'd have to add backend logic to parse the file contents, put everything into the WS feed, then pull it back out on the front-end, potentially introducing issues, for example, the current line format used by the websocket for penalties will break if a verbal cue has a '-' in it.   I have a WS-based approach in 816cc96 on this PR, and it requires a lot of extra "work" on the backend to deliver the penalties. The added complexity of the ajax call is outweighed by the simplification of the rest of the process.

I agree with the presence of a directory traversal attack, and I will make some updates to account for that.  Additionally, I agree that it'll just be simpler to allow all penalties to be expellable.

brian-brazil:
> that using some format stored in a StringRule is going to be harder to update in practice for the set of folks who would be interested in updating penalty codes.

JSON is pretty finicky, in particular around trailing commas. I wouldn't presume this for non-technical users.

> one-line string,

It doesn't have to be one line.

> It might be hard to see if that penalty is present from the ruleset editor if the penalties are all in a string.

It'd be far harder if it's in a file, which wouldn't provide a list of the penalites in the ruleset editor at all.

> From there, actually delivering the definition to the front-end, I understand the trepidation pulling this out of the WS feed, but I think keeping them in the WS feed adds significantly more complexity than delivering a standalone file.

This may be easier here, but I think is worse for the scoreboard overall. It's an entirely new data flow to consider, and we've already seen the fun that inventing new independent dataflows has caused in areas such as penalties. I'd rather not see the same mistake made again.

I believe this should follow our existing standard ways of doing things. That means that all state lives  inside the XML, and that copying the XML autosave around is sufficient to continue on a game. In addition we should be passing scoreboard state via the existing XML/WS methods rather than adding a brand new one - and one in particular that hasn't a way to update it inflight that XML/WS has. If we go for a new approach all that work needs to be reinvented and debugged, compared to taking advantage of the architecture that already exists.

Long term I'd like to consolidate to having the WS be the sole API for both internal and external users.

>  the current line format used by the websocket for penalties will break if a verbal cue has a '-' in it.

That sounds like a bug we should be fixing, team and skater names can contain hyphens.

> I have a WS-based approach in 816cc96 on this PR, and it requires a lot of extra "work" on the backend to deliver the penalties. 

That seems to be a similar number of lines of code as what you're proposing here. Rewriting the penalty code handling frontend code isn't free.

official-sounding:
Okay, I've updated the approach here somewhat:

1. Penalty codes are delivered via WS, instead of ajax
2. Penalty codes are still stored in a json file, but are parsed on the backend and delivered directly as JS objects via WS (I initially did this on accident, actually surprised this worked as well as it did)
3. removed the regex parsing of penalty types, actually simplifying the front-end instead of just changing it.

This resolves the "different special dataflow" issues, and I think should also resolve security issues of blindly delivering a file to the front-end.

> JSON is pretty finicky, in particular around trailing commas. I wouldn't presume this for non-technical users.
> It doesn't have to be one line.

I agree JSON can be finicky, I'm still of the opinion that being able to edit a json file in an editor that has syntax highlighting is going to be easier than trying to edit it directly in the ruleset editor. 

I think a JSON file is a good middle ground between no runtime penalties and a proper structured editor for penalty codes in the ruleset editor.  I envision that as being a fair bit off though, based on what I saw in the current ruleset code.  I am not prepared to add even multi-line configs to the current ruleset code/ruleset editor page without a lot more unit test coverage.

> That sounds like a bug we should be fixing, team and skater names can contain hyphens.

This is a specific bug with verbal cues, hypens were used to separate different verbal cues under the same code.

.classpath
```diff
@@ -37,5 +37,8 @@
 	<classpathentry kind="lib" path="test-libs/junit-4.11.jar"/>
 	<classpathentry kind="lib" path="test-libs/hamcrest-core-1.3.jar"/>
 	<classpathentry kind="lib" path="test-libs/mockito-all-1.9.5.jar"/>
+	<classpathentry kind="lib" path="lib/deps/jackson-annotations-2.9.5.jar"/>
```
brian-brazil, line 40: We should use only one json library, there's already another one 4 lines up.

config/penalties/wftda2016.json
```diff
@@ -0,0 +1,20 @@
+{
+	"penalties": [
+	{ "code": "B", "verbalCues": ["Back Block"]},
+{ "code": "A", "verbalCues": ["High Block"]},
```
brian-brazil: Pretty print this please.

src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
```diff
@@ -280,6 +281,25 @@ private void processPosition(String path, Position p, boolean remove) {
...
 	}
+	
+	private void processPenalties(Settings s) {
+		PenaltiesManager pm = new PenaltiesManager(s);
+
+        try {
+            PenaltiesDefinition penalties = pm.loadFromJSON();
+            for(PenaltyCode p : penalties.getPenalties()) {
+                
+                updateMap.put("PenaltyCode.("+p.getCode()+")", p);
```
brian-brazil: While you're changing all this, would you mind moving this under `ScoreBoard.`? That way a WS client will (eventually) be able to register `Scoreboard.` and get everything.

src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
```diff
@@ -280,6 +281,25 @@ private void processPosition(String path, Position p, boolean remove) {
 		updateMap.put(path + "." + Position.EVENT_SKATER, p.getSkater() == null ? null : p.getSkater().getId());
 		updateMap.put(path + "." + Position.EVENT_PENALTY_BOX, p.getPenaltyBox());
 	}
+	
+	private void processPenalties(Settings s) {
+		PenaltiesManager pm = new PenaltiesManager(s);
+
+        try {
+            PenaltiesDefinition penalties = pm.loadFromJSON();
+            for(PenaltyCode p : penalties.getPenalties()) {
```
brian-brazil: This needs to be broken down into multiple string entries, rather than one complex object that would require custom handling of some WS state entries by clients.

src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
```diff
@@ -280,6 +281,25 @@ private void processPosition(String path, Position p, boolean remove) {
 		updateMap.put(path + "." + Position.EVENT_SKATER, p.getSkater() == null ? null : p.getSkater().getId());
 		updateMap.put(path + "." + Position.EVENT_PENALTY_BOX, p.getPenaltyBox());
 	}
+	
+	private void processPenalties(Settings s) {
```
brian-brazil: PenaltyCodes

This could be confused with the actual penalties.

src/com/carolinarollergirls/scoreboard/penalties/PenaltiesManager.java
```diff
@@ -0,0 +1,36 @@
+package com.carolinarollergirls.scoreboard.penalties;
+
+import java.io.File;
+
+import com.carolinarollergirls.scoreboard.ScoreBoardManager;
+import com.carolinarollergirls.scoreboard.Settings;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class PenaltiesManager {
```
brian-brazil: PenaltyCode

src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
```diff
@@ -289,38 +309,12 @@ private void initialize(ScoreBoard sb) {
...
-		updateMap.put("PenaltyCode.FO_EXP(EXP-N)", "Expulsion-Interference");
-		updateMap.put("PenaltyCode.FO_EXP(EXP-G)", "Expulsion-Misconduct");
-		updateMap.put("PenaltyCode.FO_EXP(EXP-?)", "Expulsion-Unknown");
-
+		
+		
 		// Process Settings
 		processSettings("ScoreBoard", sb.getSettings());
+		
+		processPenalties(sb.getSettings());
```
brian-brazil: This code will only get called once at startup as far as I can see, can the codes actually be changed during a run/game?

.classpath
official-sounding: I agree with this wholeheartedly, and my next project is to strip out org.json (both the seemingly unused vendored source code and the jar & uses in the scoreboard.  org.json has a weird license and is much lower level than we should really be using.

That said, I think removing org.json is a bit of an undertaking that is outside of the scope of this PR.  if you'd prefer me to rewrite PenaltyCodeManager to use org.json constructs for now I can, but I would rather not do that just to undo it in a few weeks.

Also, after doing a bit of reading (and realizing that I had added a 1.3MB dependency), I swapped jackson-databind for jackson-jr, that is much lighter weight but still gives us a much nicer org.json replacement.

official-sounding:
fixed style/naming convention issues from code review, and added some code to actually correctly handle updating which penalty definition file is used while the scoreboard is running.

.classpath
brian-brazil: It's not the best of libraries alright.

> I would rather not do that just to undo it in a few weeks.

I'd hold off on that until I've finished removing the Game directory, as that's going to have json-y impacts. Does jackson-jr permit building up json by hand like the currently library does?

src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
```diff
@@ -280,6 +285,26 @@ private void processPosition(String path, Position p, boolean remove) {
 		updateMap.put(path + "." + Position.EVENT_SKATER, p.getSkater() == null ? null : p.getSkater().getId());
 		updateMap.put(path + "." + Position.EVENT_PENALTY_BOX, p.getPenaltyBox());
 	}
+	
+	private void processPenaltyCodes(Settings s) {
+		PenaltyCodesManager pm = new PenaltyCodesManager(s);
+		updateMap.put("ScoreBoard.PenaltyCode", null);
```
brian-brazil: I don't think you can both clear a subtree like this and give it entries in one go, as the updateMap can be processes in any order as maps are unordered.

html/controls/pt/index.js
```diff
@@ -213,20 +213,23 @@ function submitPenalty() {
...
-	var code = match[2];
+	var code = k.split('.').pop();
 
-	var div = $('.Codes .' + type + '[code="' + code + '"]');
-	if (div.length > 0) {
-		if (v == null) {
-			div.detach();
-			return;
+
+		if(code !== 'FO') {
```
brian-brazil: Maybe we should do the FO code handling entirely in the frontend? It feels odd to list it as a penalty code.

src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
official-sounding: Since updateMap is a LinkedHashMap, the iteration order is stable in insertion order.

brian-brazil: There's two problems with that. Firstly a entry could be nulled and later have a value put in, and the subtree wouldn't get cleared. Secondly we don't iterate over the updates in WS.updateState, we get the set of keys and iterate over that.

We'd need to change that code somehow to make this work correctly in general.

official-sounding: > Firstly a entry could be nulled and later have a value put in, and the subtree wouldn't get cleared.
I agree on this in the general case, I think the subtree deletion 

> Secondly we don't iterate over the updates in WS.updateState, we get the set of keys and iterate over that.

official-sounding: > Firstly a entry could be nulled and later have a value put in, and the subtree wouldn't get cleared.

I agree on this in the general case, I think the subtree state deletion basically only works on accident.  In this case though I'm never putting anything into the top-level key.  I'd probably rather update WS to provide an explicit "removeState/deleteState" function to give us more explicit control over what is in the state object.  

If you have an alternative approach in mind, we can break out that discussion to the FB or slack.  Unless you think it's too big of a change for this PR, I can incorporate it with these changes.

> Secondly we don't iterate over the updates in WS.updateState, we get the set of keys and iterate over that.

It's original-key-insertion-order specifically.  That's definitely more nuanced than just insertion order, but since the updateMap is emptied after handing off to WS.updateState, that doesn't become a problem in practice all that much.

brian-brazil: Pondering this more, this design as-is is just not safe in general. A hash map is the wrong data structure, as it loses ordering. This may not be an issue for a single deletion and then adding it back, however there's the notion of batching so you could have delete+add+delete+add which would end up with a mixture of the two adds rather than only the last add being present.

I don't believe existing code is affected, as skaters don't change attributes from update to update; and the set of policies is fixed. By contrast penalty codes can be added and removed.

Without getting into a full-transactional design, I think switching the updateMap to a list should do the trick. That way the updates always happen in order, and will still get squashed down before being sent out to clients.

.classpath
official-sounding: yes, jackson-jr provides a fluent api that works very similarly to how org.json manipulates json objects.

official-sounding:
Updated the updates to use a List instead of a Map, moved the FO stuff to the front-end, and made it slightly clearer whether the user is editing an expulsion code vs a regular code.

src/com/carolinarollergirls/scoreboard/jetty/WS.java
```diff
@@ -28,15 +28,17 @@
...
 import org.json.JSONObject;
 
-import com.carolinarollergirls.scoreboard.ScoreBoard;
 import com.carolinarollergirls.scoreboard.ScoreBoardManager;
 import com.carolinarollergirls.scoreboard.json.ScoreBoardJSONListener;
+import com.carolinarollergirls.scoreboard.json.WSUpdate;
 import com.carolinarollergirls.scoreboard.model.ScoreBoardModel;
 
 public class WS extends WebSocketServlet {
+	private static final long serialVersionUID = 1L;
```
brian-brazil: This is unused

src/com/carolinarollergirls/scoreboard/jetty/WS.java
```diff
@@ -62,43 +64,35 @@ protected static void unregister(Conn source) {
...
+							s.value = null;
 						}
 					}
 				} else {
-					State s = state.get(k);
-					if (s == null) {
-						if (v != null)
-							state.put(k, new State(stateID, v));
-					} else if (!v.equals(s.value)) {
+					State s= state.get(update.getKey());
```
brian-brazil: s =

brian-brazil (commented):
This looks good, just 2 nits.

src/com/carolinarollergirls/scoreboard/jetty/WS.java
frank-weinberg: Not having it in a serializable class will generate a warning, even if we don't serialize.

brian-brazil: We don't have this in any other class. Where are you seeing these warnings?

Also we shouldn't be using Java serialisation anywhere.

frank-weinberg: Eclipse gives the warning for 17 files that all implement java.io.Serializable (inherited through some other interface).

brian-brazil: I don't think we should be cluttering the code due to false-positive warnings from one editor.

frank-weinberg: The regular javac used by ant also gives this warning (if you don't configure ant to hide all warnings).

brian-brazil: I see no warnings from the default build, Java 1.8.0_162-8u162-b12-0ubuntu0.16.04.2-b12. If there are such warnings we should disable them.

frank-weinberg: Setting nowarn to false and xlint:all in build.xml gives 99 warnings with javac 1.8.0_172
And I'm against disabling warnings. If these classes don't need to be serializable, don't implement that interface. But if you implement it, do it properly. (That being said: using 1 as UID is also improper. )

brian-brazil: So it's not a warning, it's a lint. If we're going to enable those we should do so as a separate PR, not piecemeal. We should also make sure that any warnings are actually issues, and not false positives like this one.

Serialisation is a very old java feature that noone should be using these days, I don't think our code should suffer due to its legacy.

brian-brazil: And looks like it's easy to disable that particular lint warning.


I'm not against useful warnings, but I also like reasonably self-contained changes.

PR #84, opened 2018-06-27
auto-size PT grid to screen

### LINKED ISSUE #67: PLT Tracking Page - Auto Sizing Functions
From BrewHaHa this weekend...

4) Can the rows autoheight so they expand to be bigger if there's fewer skaters?

I would expand this to adjust the view depending on the display size also and the number of skaters.

official-sounding:
This is very easy thanks to "vh" units in css.  Only apply them in the non-mobile view so that more penalties can be seen at once in that view (based on understanding of how that view is supposed to work).  I can change this if desired though.

Closes #67

### DIFF
--- a/html/controls/pt/index.css
+++ b/html/controls/pt/index.css
@@ -36,11 +36,6 @@ tr:nth-child(odd) {
 .PenaltyEditor .Codes>div .Description { font-size: 0.75em; }
 
 /* responsive vertical layout on phones and tablet portrait */
-@media only screen and (min-width: 960px) {
-	.Team {
-		height: 100vh;
-	}
-}
 @media only screen and (max-width: 960px) { 
    .Team{
 	position: relative;
--- a/html/controls/pt/index.css
+++ b/html/controls/pt/index.css
@@ -36,6 +36,11 @@ tr:nth-child(odd) {
 .PenaltyEditor .Codes>div .Description { font-size: 0.75em; }
 
 /* responsive vertical layout on phones and tablet portrait */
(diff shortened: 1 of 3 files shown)


brian-brazil:
@riffreff Would you mind reviewing this?

riffreff:
Sorry for replying on the wrong issue.

I can add this change in the next days. I would like to hear a few more opinions, if that is really the desired effect. 
I personally am not against it. However, the issue I see is, that if you either have vastly different skater numbers on teams you have a very weird layout. And also if the numbers vary just by one or two skaters the rows will be slightly off, which might be uncomfortable.

official-sounding:
based on feedback from the FB group, I went ahead and put this behind a Ruleset toggle, defaulting it to "All Rows the Same Height".

After some thought though, I agree with Brian's comment in #93 however that this is not really a "rule" per se and Ruleset is not the best place for it.  It's honestly is probably closer to a per-operator preference, and maybe should be an extension of that side (though adding a "login" to the PT screen also seems sort of weird)

I'm happy to let this PR sit pending more feedback and rework it once we figure out a better place to put the setting.

brian-brazil:
This doesn't belong in a ruleset as that's for game-level things, and would not allow a PT to change it to their preferences as it'd be too late after a game starts. The current standard way to do this is a URL parameter, or new screen (see Minimal Animations, and our various overlay/PT duplicates).

official-sounding:
I moved the setting from a Ruleset to a URL parameter.  I'll add a link to it when I remove the legacy PT stuff in a different PR later this week.

PR #86, opened 2018-07-01
Recreate skater rows on number change for PT screens

### LINKED ISSUE #85: Penatly view not updated if skater number changed
If a skater's number is changed the penalty view (controls/pt/ptcolor.html) is not updated with this, but does show with a refresh. The dialog that pops when you click is updated, so this is likely a js issue.

brian-brazil:
Fixes #85


official-sounding (approved):
This looks good.

PR #87, opened 2018-07-01
Add a 30s heartbeat to WS client code

brian-brazil:
Part of the lag fix was no longer sending
updates which clients hadn't requested. Due
to this it is now possible that a client would have
no communication for some time, such as if there was
a long official timeout or during the intermission.

The WS code on the server side will kill such connections
after 5m of no activity. The client code should autoreconnect,
however under some circumstances which remain unclear this
does not happen. To mitigate this, add a heartbeat.
This also adds protection against falling out of NAT connection
tables etc. if someone is using it on a network where that matters.

### DIFF
--- a/html/json/WS.js
+++ b/html/json/WS.js
@@ -6,6 +6,7 @@ var WS = {
 	callbacks: new Array(),
 	Connected: false,
 	state: { },
+	heartbeat: null,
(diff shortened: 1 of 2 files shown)


official-sounding (approved):
This looks good to me.

PR #90, opened 2018-07-04
Fix issues when Jam Clock is counting Up

### LINKED ISSUE #89: Jam doesn't stop automatically at 2 minutes when counting up
When setting the jam clock to count up, the jam does not end automatically when the clock reaches 2:00.  the stop jam/start jam buttons work as expected, but the lineup time does not automatically start.

official-sounding:
There are a few bugs in ScoreBoardModel related to checking if a clock is at the minimum time, which causes problems when the clock is instead set to count up.  I've updated these instances (and the other instances that are correct, to simplify & avoid errors in the future) to use the isTimeAtStart/isTimeAtEnd functions instead.

Additional changes:

1. Add a bunch of unit tests around the logic in the DefaultClockModel.  The actual nitty-gritty of start/stop and ticking on the Runnable still needs to be tested, but that will probably need some refactoring to be unit testable correctly.

2. Clean up the warnings in DefaultScoreBoardModel and DefaultClockModel.  These are cosmetic changes, mostly.

3. split the "period End" logic and "jam End" logic into two separate event handlers.  I think this split makes things a bit clearer.

Next steps:
1. A bunch of logic exists in Policies around clocks.  It's all correct, but I didn't touch it because those Policies will change into Rules (hopefully) in the near future.

2. More Unit Tests, both around DefaultClockModel and the actual "business rules" in the ScoreBoardModel related to clocks.

3. Expose a getElapsedTime()/getRemainingTime() method on the Clock object.  This will make it easier to simplify a lot of other instances that we are looking at times, which seems very error prone.

4. Clean up the apis related to Rulesets slightly to reduce the reliance on statics and improve testability.

Closes #89

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
+++ b/src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
@@ -116,6 +116,7 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 	protected void addInPeriodListeners() {
 		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_PERIOD, "Running", Boolean.TRUE, periodStartListener));
 		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_PERIOD, "Running", Boolean.FALSE, periodEndListener));
+		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_JAM, "Running", false, jamEndListener));
 		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_JAM, "Running", Boolean.FALSE, periodEndListener));
 		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_JAM, "Running", Boolean.TRUE, jamStartListener));
 		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_TIMEOUT, "Running", Boolean.FALSE, periodEndListener));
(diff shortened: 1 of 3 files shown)


brian-brazil:
> Expose a getElapsedTime()/getRemainingTime() method on the Clock object. This will make it easier to simplify a lot of other instances that we are looking at times, which seems very error prone.

Don't getTime/getInvertedTime cover this already?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -116,6 +116,7 @@ public void setInPeriod(boolean p) {
 	protected void addInPeriodListeners() {
 		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_PERIOD, "Running", Boolean.TRUE, periodStartListener));
 		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_PERIOD, "Running", Boolean.FALSE, periodEndListener));
+		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_JAM, "Running", false, jamEndListener));
```
brian-brazil: Boolean.FALSE for consistency

official-sounding:
> Don't getTime/getInvertedTime cover this already?

Not exactly.  For example, at jam start, if the jam clock is counting down, getTime will return 120000, and getTimeInverted will return 0.  if the clock is counting up however, getTime will return 0, and getTimeInverted will return 120000.

a getElapsedTime method would always return 0, and getRemainingTime would always return 12000 at jam start, regardless of clock direction.

brian-brazil:
Ugh, comments in the interface will be required I think.

PR #92, opened 2018-07-04
Make it possible to upload images if directory is empty

brian-brazil:
The default release has no sponsor/team images, so you have
to place them on the filesystem by hand. With this change
you can do this entirely from the UI now.

Another small quality of life improvement.


src/com/carolinarollergirls/scoreboard/xml/MediaXmlDocumentManager.java
```diff
@@ -98,6 +98,10 @@ protected void initializeType(File typeDir) {
 		Iterator<File> files = Arrays.asList(typeDir.listFiles((FileFilter)mediaFileFilter)).iterator();
 		while (files.hasNext())
 			addMediaElement(typeDir.getName(), files.next());
+		// Ensure we have an entry, even if dir is empty.
+		synchronized (updateLock) {
```
frank-weinberg: There should be brackets on the preceding while in order to make clear that the new code is intended to not be part of the loop.

brian-brazil: Done.

The codebase is littered with these, I'm planning on doing a big style PR once there's no other changes pending.

PR #93, opened 2018-07-04
Update handling of settings (part 1)

frank-weinberg:
This is replaying some of the uncontroversial changes from #60:

- Moving policies to rulesets
- Touching up the ruleset editor a bit
- Adding convenience methods to the Clock Interfaces that abstract from the direction of the clock

@official-sounding This PR covers/affects several of your future items mentioned in #90. 

The next part I'm looking into regarding rules/settings will be splitting all of them into four groups:

1. Operator settings: Influence only the operator interface. Can be set there and are stored for individual operators (like the key bindings).
2. Scoreboard Settings: Influence the appearance or non-game-relevant behaviour (running direction of clocks, display of per jam scores, etc.) of the scoreboard. Can be set in the scorebaord (many of those are already on the View Tab) and overridden via a rule set (i.e. if the ruleset contains a specific value that value is set at game start, otherwise the value from the scoreboard is left unchanged.)
3. Rule settings: Influence the game-relevant behaviour of the scoreboard (duration of jams/periods/intermission; number of timeouts etc.). Have to be set in a ruleset (i.e. if the ruleset doesn't specify an explicit value, the value from the default ruleset is set at game start.) A way to change them during the game may be added at a later point.
4. Internal settings that a user shouldn't be able to change. (e.g. min/max number/time of clocks) Some of these may be affected by other rules (e.g. I would like to have a setting "Rule.Period.Duration" that internally sets the max time of the period clock but is a lot more intuitive than "Clocks.Period.MaximumTime")


brian-brazil:
I think we can merge these groups a bit.

> Scoreboard Settings ... (i.e. if the ruleset contains a specific value that value is set at game start, otherwise the value from the scoreboard is left unchanged.)

If these are non-game-relevant then they shouldn't come from the ruleset, as they're not about rules.

> Internal settings that a user shouldn't be able to change. (e.g. min/max number/time of clocks)

Do sanctioned games still require a 15m intermission? These sound like ruleset things.

html/controls/rulesets.js
```diff
@@ -71,10 +74,14 @@ function loadDefinitions() {
 				if (s.attr("group") == def.group && s.attr("subgroup") == def.subgroup) {
 					section = s;
 				}
-				if (s.attr("group") > def.group || (s.attr("group") == def.group && s.attr("subgroup") > def.subgroup)) {
+				/*
```
brian-brazil: Don't leave commented out code lying about, it causes confusion and leads to bugs.

src/com/carolinarollergirls/scoreboard/Clock.java
```diff
@@ -24,6 +24,8 @@
 
 	public long getTime();
 	public long getInvertedTime();
+	public long getTimeElapsed();
```
brian-brazil: These 4 methods need comments

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -131,40 +129,34 @@ public void setInOvertime(boolean o) {
...
-			getClockModel(Clock.ID_INTERMISSION).stop();
-			try {
-				OvertimeLineupTimePolicy p = (OvertimeLineupTimePolicy)getPolicy(OvertimeLineupTimePolicy.ID);
-				if (null != p)
-					p.startOvertime();
-			} catch ( ClassCastException ccE ) {
-				ScoreBoardManager.printMessage("Internal Error: invalid OvertimeLineupTimePolicy : "+ccE.getMessage());
+			ic.stop();
+			long otLineupTime = settings.getLong("Clock." + Clock.ID_LINEUP + ".OvertimeTime");
+			if (lc.getMaximumTime() < otLineupTime) {
```
brian-brazil, line 156: I'd expect this to override one way or the other

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -188,25 +180,10 @@ public void startJam() {
 				ClockModel ic = getClockModel(Clock.ID_INTERMISSION);
 				lineupClockWasRunning = getClockModel(Clock.ID_LINEUP).isRunning();
```
brian-brazil, line 182: This wasn't used

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -188,25 +180,10 @@ public void startJam() {
...
 
-				// If intermission clock has almost run down, end intermission
```
brian-brazil, line 210: This change is notable enough to be mentioned in the commit description

Is jam start during intermission undoable?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -579,59 +522,121 @@ protected void createTeamModel(String id) {
...
+			requestBatchEnd();
+		}
+	};
+	protected ScoreBoardListener intermissionEndListener = new ScoreBoardListener() {
+		public void scoreBoardChange(ScoreBoardEvent event) {
+			ClockModel ic = getClockModel(Clock.ID_INTERMISSION);
+			ClockModel pc = getClockModel(Clock.ID_PERIOD);
+			ClockModel jc = getClockModel(Clock.ID_JAM);
+			if (ic.getTimeElapsed() > ic.getTimeRemaining() && pc.getNumber() < pc.getMaximumNumber()) {
+				//if more than half of the intermission has elapsed and there is another period,
```
brian-brazil: This seems a bit magic

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -579,59 +522,121 @@ protected void createTeamModel(String id) {
...
+			if (ic.getTimeElapsed() > ic.getTimeRemaining() && pc.getNumber() < pc.getMaximumNumber()) {
+				//if more than half of the intermission has elapsed and there is another period, 
+				// start the next period. Otherwise extend the previous period.
 				requestBatchStart();
-				setTimeoutOwner("");
-				setOfficialReview(false);
-				getClockModel(Clock.ID_TIMEOUT).changeNumber(1);
+				pc.changeNumber(1);
+				pc.resetTime();
+				if (settings.getBoolean("ScoreBoard." + Clock.ID_JAM + ".Number")) {
```
brian-brazil: There's no such setting, you want MinimumNumber and to set it rather than 1

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -579,59 +522,121 @@ protected void createTeamModel(String id) {
...
+					jc.setNumber(1);
+				} else {
+					jc.changeNumber(1);
```
brian-brazil: What's the purpose of this?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -579,59 +522,121 @@ protected void createTeamModel(String id) {
...
 			}
-		};
-
+		}
+	};
+	protected ScoreBoardListener lineupClockListener = new ScoreBoardListener() {
+		public void scoreBoardChange(ScoreBoardEvent event) {
+			if (settings.getBoolean("ScoreBoard." + Clock.ID_LINEUP + ".AutoStart")) {
+				ClockModel lc = getClockModel(Clock.ID_LINEUP);
+				long bufferTime = settings.getLong("ScoreBoard." + Clock.ID_LINEUP + ".AutoStartBuffer");
```
brian-brazil, line 617: This is a new feature, I suggest leaving it to a separate PR. Particularly as I believe this may have unintended consequences and is not in line with the rules.

src/com/carolinarollergirls/scoreboard/jetty/XmlScoreBoardServlet.java
```diff
@@ -8,30 +8,26 @@
...
-import org.apache.commons.fileupload.servlet.*;
+import com.carolinarollergirls.scoreboard.ScoreBoard;
+import com.carolinarollergirls.scoreboard.ScoreBoardManager;
 
 public class XmlScoreBoardServlet extends AbstractXmlServlet
 {
+	/**
+	 * 
+	 */
+	private static final long serialVersionUID = 7299658481660605015L;
```
brian-brazil: Please don't clutter up the code with unused variables.

src/com/carolinarollergirls/scoreboard/model/ClockModel.java
```diff
@@ -35,6 +35,7 @@
 
 	public void setTime(long ms);
 	public void changeTime(long ms);
+	public void elapseTime(long ms);
```
brian-brazil: How is this different from changeTime?

src/com/carolinarollergirls/scoreboard/xml/ScoreBoardXmlListener.java
```diff
@@ -100,33 +106,26 @@ public void scoreBoardChange(ScoreBoardEvent event) {
...
-					editor.removeElement(getScoreBoardElement(), "Policy", ((Policy)event.getValue()).getId());
-				else
-					editor.setRemovePI(converter.toElement(getScoreBoardElement(), (Policy)event.getValue()));
 			} else {
 				editor.setElement(getScoreBoardElement(), prop, null, v);
 			}
 		} else if (p.getProviderName().equals("Team")) {
 			if (prop.equals(Team.EVENT_ADD_ALTERNATE_NAME)) {
-				Element e = converter.toElement(getTeamElement((Team)p), (Team.AlternateName)event.getValue());
+				//Element e = converter.toElement(getTeamElement((Team)p), (Team.AlternateName)event.getValue());
```
brian-brazil: Why is this and other lines commented out?

brian-brazil (commented):
There are changes here beyond moving the policies around, some of which would be best done in another PR.

This is touching lots of key internals and is bound to be breaking something. This should all have unittests.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: Undo code is a mess and I will replace all of it by a snapshot model, but that is beyond the scope of this PR.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: I don't understand this comment.

brian-brazil: Why is there an if here rather than being unconditional?

src/com/carolinarollergirls/scoreboard/model/ClockModel.java
frank-weinberg: It takes into account the direction the clock is running. On a countdwon clock a positive value will reduce the time with elapseTime.

src/com/carolinarollergirls/scoreboard/xml/ScoreBoardXmlListener.java
frank-weinberg: The variables are unused currently. But they will probably be needed once someone implements this functionality.

src/com/carolinarollergirls/scoreboard/model/ClockModel.java
brian-brazil: Can you add comments to these methods explaining this?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: Because we don't want the maximum time to be reduced. For starters it would break AutoStart, which needs a maximum time of at least lineup time + buffer time.

src/com/carolinarollergirls/scoreboard/xml/ScoreBoardXmlListener.java
brian-brazil: I think toElement has a side-effect, and this breaks functionality.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: No more magic than having the tipping point at 1 minute before intermission end as it was before.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: There is. See line 41. and Ruleset.java line 57.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: I think it's best to avoid making business logic changes in a PR like this, there's enough risk of breakage without mixing in unrelated changes.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: It allows to have jam numbers to continue counting through all periods instead of restarting at 1 each period.  (This functionality is moved over from policies.)

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: This is not new. It has been in policies for ages.

brian-brazil: Ah, I see it now

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: That setting naming is a bit confusing then, as it overlaps with the WS event name for the jam number. Can you choose a different setting name?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: That setting naming is a bit confusing then, as it overlaps with the WS event name for the jam number. Can you choose a different setting name?

You should still be using MinimumNumber

frank-weinberg:
This commit should address all comments not yet resolved.

brian-brazil (commented):
The code itself looks okay, but I think a few minimal tests around the various changes in DefaultScoreBoardModel would be a good idea. That logic is all fairly hairy.

frank-weinberg:
What tests do you have in mind?
Given that the changes are primarily in the event listeners, any meaningful test for those will require quite a big setup, that's probably worthy of it's own PR.

brian-brazil:
The state changes in startOvertime shouldn't be hard to test.

For the listener stuff I think the main challenge is that ScoreBoardEventProviderManager is a singleton.

frank-weinberg:
Status update: I'm working on unit tests for DefaultScoreboardModel as a separate PR.

brian-brazil:
So will we get this one in and then do the tests separately?  My next PR will also have a fair bit of tests, as I'll be moving jam stats into the model.

frank-weinberg:
I'm fine with merging this one before the tests are completed.

PR #96, opened 2018-07-05
Enable warnings from javac.

brian-brazil:
Bump to Java 1.6 to remove warning about 1.5 being deprecated. It's
been 12 years, we're probably safe.
Supress a warning about bootstrap class path not set in conjunction with
-source 1.6.

This keeps a standard build clean of warnings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ The scoreboard should be unzipped into a folder on the local machine. The user r
 
 ### Java
 
-Java is required for providing a Java Runtime Environment (JRE) version 1.5.0 or newer. Installing the latest version of Oracle's Java is recommended.
+Java is required for providing a Java Runtime Environment (JRE) version 1.7.0 or newer. Installing the latest version of Oracle's Java is recommended.
 
 * Windows users can install the standard Java for Windows package that is available when clicking on Free Java Download from [Oracle’s Java site](https://java.com/).
 
--- a/build.xml
+++ b/build.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0"?>
 
 <project name="Carolina Rollergirls Scoreboard" default="compile">
-	<property name="ant.build.javac.source" value="1.5"/>
-	<property name="ant.build.javac.target" value="1.5"/>
+	<property name="ant.build.javac.source" value="1.7"/>
+	<property name="ant.build.javac.target" value="1.7"/>
 
 	<property name="src.dir" value="src"/>
 	<property name="dest.dir" value="bin"/>
@@ -17,7 +17,7 @@
 
 	<property name="compile.deprecation" value="true"/>
 	<property name="compile.debug" value="true"/>
-	<property name="compile.nowarn" value="true"/>
+	<property name="compile.nowarn" value="false"/>
 	<property name="compile.verbose" value="false"/>
 
 	<property name="jar.dir" value="lib"/>
@@ -101,6 +101,7 @@
 
 			<compilerarg value="-Xlint:all"/>
 			<compilerarg value="-Xlint:-serial"/>
+			<compilerarg value="-Xlint:-options"/>
 			<classpath>
 				<fileset dir="${jar.dir}" includes="deps/*.jar" />
 			</classpath>
--- a/json/JSONObject.java
+++ b/json/JSONObject.java
@@ -298,7 +298,7 @@ public class JSONObject {
      */
     public JSONObject(Object object, String names[]) {
         this();
-        Class c = object.getClass();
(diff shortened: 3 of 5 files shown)


frank-weinberg:
Why 1.6? Even JDK 1.7 has been EOLed more than 3 years ago. And without the bootstrap cp there is no check that the generated code will even run on 1.6.

brian-brazil:
1.6 is enough to stop the warning.

EOL doesn't mean that noone is using it, and some lifecycles are overly agressive. Java 8 is only 4 years old, that's quite new as these things go. I wouldn't presume that all of our users are on 8, and even 8 is going to be EOLed in 6 months.

official-sounding:
Should we consider jumping straight from 5 to 7? 

Based on my reading of the [JEP 182](http://openjdk.java.net/jeps/182) deprecation process, it looks like 6 will be [completely removed with JDK 12](https://bugs.openjdk.java.net/browse/JDK-8028563), roughly [March 2019](http://openjdk.java.net/projects/jdk/12/spec/).  Support will still be in the LTS 11 version, but then folks will have to keep around a v11 JDK to work on the scoreboard, which will increase friction.

If we upgrade to 7, support will be deprecated in 12, but will likely not be removed until at least 17 (so 2024 or so, I think?).  This will get us "off the treadmill" a bit on version upgrades, and not require developers to keep an LTS JDK around.

Java 7 is still 7 years old, and while I bet there are some scoreboards out there running on Java 6, I would also bet they are not likely to be upgrading their scoreboard software either.

Also Java 7 introduces some major quality of life improvements in language features. That + wins on contributor friction point to dropping support for 6 in my mind.

frank-weinberg:
^This. Plus I object to suppressing the warning about the bootstrap cp. At least let developers see that there is no check being performed against the old runtime.

brian-brazil:
I don't think it's something we need to decide right now. It mightn't hurt to do an informal survey of our users first, there are quite a few trying to run on older kit that may not have a recent jdk.

> If we upgrade to 7, support will be deprecated in 12,

That's not my reading, 7 will be dropped in 11. 11 will be out in August 2018 as they're moving to a 6 month release cycle. Which is only a month away.

> Plus I object to suppressing the warning about the bootstrap cp.

I object to warnings that developers can do nothing about. Warnings only work if we can keep them clean, otherwise they're just noise.

Unless we're going to build out CI infrastructure that sets this correctly, I think it'd be better to lose this warning but gain others.

brian-brazil:
> That's not my reading, 7 will be dropped in 11.

Looks like they made an exception: http://mail.openjdk.java.net/pipermail/jdk-dev/2018-May/001190.html

It's still all very short timelines, that's only 7 months before it breaks.

official-sounding:
A poll is fine by me.

> 7 will be dropped in 11

While the original plan appears to be removing 6 in JDK10, this JDK bug tracker item has 6 support removed in version 12 (the "Fix Version/s" field):
https://bugs.openjdk.java.net/browse/JDK-8028563

7 being deprecated in version 12 (the "Fix Version/s" field):
https://bugs.openjdk.java.net/browse/JDK-8173606

7 being removed in at least version 13
https://bugs.openjdk.java.net/browse/JDK-8173605

frank-weinberg:
According to the thread following on the mail @brian-brazil linked, 7 will likely only be removed after the next LTS version, which would be three more years down the road. And even if not, since 11 is LTS, there's a good chance that it will remain the default for quite some time. (Like it is currently the case with 8.)

brian-brazil:
> 7 being removed in at least version 13

That gives us about a year and a half. Likely all this will have settled down by then. Let's see how the survey goes, and how many stragglers we have.

brian-brazil:
The survey indicates that almost everyone is on 8, with one on each of 7 and 10. Accordingly I've bumped to 7.

brian-brazil:
I've also fixed some new warnings missed by #97. There's still 3 "overrides equals, but neither it nor any superclass overrides hashCode method".

official-sounding:
I'm not the biggest fan of updating the vendored library code, but hopefully it's not long for this world anyways, so I'm fine with those changes.

Beyond that, this PR looks good.

brian-brazil:
Yeah, I'm unsure why this is like this rather than a jar in the first place, possibly there are other modifications already: https://github.com/rollerderby/scoreboard/commit/eff3e077ba0dfa2c185461fc0bd6d4a3b0f5943c

As you say it'll be gone soon-ish.

rogerharkavy:
Just a reminder that when you make a change in the README, please make the corresponding change on the wiki page. Thank you.

PR #98, opened 2018-07-06
Fix DDTV overlay to get team data from the right place. Cleanup.

brian-brazil:
Some Game->Scoreboard were missed in
https://github.com/rollerderby/scoreboard/commit/0de1a43e6026ae4bde750e52d057ebd3e8fe6d9d

Also remove dead code around "PPJ", as that's one less place Game. is
mentioned.

This brings us down to one overlay that `Game.` WS updates are used in. Once those are gone `WS` can be made non-static, and we can start writing unittests around it.

### DIFF
--- a/html/views/overlays/ddtv/admin.html
+++ b/html/views/overlays/ddtv/admin.html
@@ -7,55 +7,6 @@
 	</head>
 
 	<body>
-<!--
-	<div class="controls">
-
-	<p id="Panel">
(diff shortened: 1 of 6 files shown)


brian-brazil:
Any objections to me merging this?

frank-weinberg:
I think we should have a second set of eyes looking at each PR before merging.

frank-weinberg (approved):

PR #100, opened 2018-07-07
Get rid of Game.Team WS, it's already in ScoreBoard.Team.

brian-brazil:
Simplify TeamInfo to get the information it needs as it needs it,
rather than trying to maintain state.

The next step is harder, all the per-jam stats need to be moved under the scoreboard object. Plus persisted in the XML.

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/game/TeamInfo.java
+++ b/src/com/carolinarollergirls/scoreboard/game/TeamInfo.java
@@ -25,136 +25,42 @@ public class TeamInfo extends Updater {
 	}
 
 	public void snapshot() {
-		Team t = game.getTeam(team);
-		name = t.getName();
-		logo = t.getLogo();
-
-		// Add/Update/Remove Alternate Names
-		List<Team.AlternateName> t_an = t.getAlternateNames();
-		for (Team.AlternateName an : t_an) {
-			if (an != null && an.getId() != null && an.getName() != null)
-				alternateNames.put(an.getId(), an.getName());
-		}
-		Object[] keys = alternateNames.keySet().toArray();
-		for (Object oid : keys) {
-			String id = (String)oid;
-			boolean found = false;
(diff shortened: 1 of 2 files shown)


official-sounding (commented):
Removing everything but toJSON from TeamInfo is a bit strange to me. 

I would prefer moving the JSON piece to Team/DefaultTeamModel, and deleting TeamInfo entirely.  I'm assuming that is coming in a future PR, though?

brian-brazil:
Deleting all of `game` is coming as is removing all this json marshalling, but one step at a time.

src/com/carolinarollergirls/scoreboard/game/TeamInfo.java
```diff
@@ -25,136 +25,42 @@ public String getUpdaterBase() {
...
 		json.put("team", team);
-		json.put("name", name);
-		json.put("logo", logo);
-		json.put("color", alternateNames.get("operator"));
+		json.put("name", t.getName());
+		json.put("logo", t.getLogo());
+		Team.AlternateName operatorName = t.getAlternateName("operator");
+		if (operatorName != null) {
+			json.put("color", operatorName.getName());
+		}
```
frank-weinberg, line 39: Assuming that the operator name (if set) will be the team color is bound to break at some point. We should be looking to make this attribute accessible directly (and then add an operator option to use it as operator name, if set).
(I am aware that this is not in the scope of this PR.)

brian-brazil: This data is already available directly, and when I replace all of this this special casing will go away.

brian-brazil:
Any objections to me merging this?

I'm trying where possible to go with many smaller PRs rather than one massive one that's harder to review.

frank-weinberg:
Two people have reviewed without substantial objections, so this one should be fine. And then I can just do the click myself.

PR #106, opened 2018-07-10
Fix broken test on DefaultSettingsModel

official-sounding:
I don't know if I just didn't run this test before adding it or what, but set() doesn't actually do what I expected to the setting dictionary -- it makes an existing key '' instead of null.

I've updated the test to reflect that functionality and all the tests pass.

### DIFF
--- a/tests/com/carolinarollergirls/scoreboard/DefaultSettingsModelTests.java
+++ b/tests/com/carolinarollergirls/scoreboard/DefaultSettingsModelTests.java
@@ -1,5 +1,6 @@
 package com.carolinarollergirls.scoreboard;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 
@@ -56,8 +57,8 @@ public class DefaultSettingsModelTests {
 		
 		settingsModel.reset();
 		
-		assertNull(settingsModel.get("Example.Team1"));
-		assertNull(settingsModel.get("Example"));
+		assertEquals("",settingsModel.get("Example.Team1"));
+		assertEquals(null,settingsModel.get("Example"));
 	}
 
 }
--- a/build.xml
+++ b/build.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 
-<project name="Carolina Rollergirls Scoreboard" default="compile">
+<project name="Carolina Rollergirls Scoreboard" default="test">
 	<property name="ant.build.javac.source" value="1.7"/>
 	<property name="ant.build.javac.target" value="1.7"/>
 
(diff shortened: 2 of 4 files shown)


tests/com/carolinarollergirls/scoreboard/DefaultSettingsModelTests.java
```diff
@@ -1,7 +1,6 @@
 package com.carolinarollergirls.scoreboard;
 
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertSame;
+import static org.junit.Assert.*;
```
brian-brazil: You and @frank-weinberg are changing these in different directions. Can ye decide which ye're going with?

brian-brazil (commented):
Let's enable tests on every build, so things like this don't slip through.

frank-weinberg:
Can we also agree to sync the packet structure of src/ and tests/, i.e. move this one and DefaultPositionModelTests to the defaults/ package?

brian-brazil:
@frank-weinberg That makes sense to me, that's the standard way to do it.

frank-weinberg:
Enabling tests on every build increases build time and dependencies.
But we should make sure that we only apporve/merge changes on the java code that pass all tests.

brian-brazil:
The increase in build time should be negligible, and I'd expect unittests to always be run by default unless explicitly requested not to. I've already asked @jaredquinn to enable CircleCI so I can setup test status on PRs.

frank-weinberg:
The tests I already have for DefaultScoreboardModel currently take 24s. A clean build takes 3s.

official-sounding:
I've standardized the imports, moved everything under the defaults package, and made "test" the default ant task.

>I've already asked @jaredquinn to enable CircleCI

I'd recommend travis-ci over CircleCI, as travis-ci seems to have better out-of-the-box support for ant, the [.travis.yml file](https://github.com/official-sounding/scoreboard/blob/feature/travis-ci/.travis.yml) I put together is one line, and works with our ant file pretty much perfectly, running tests. Though CircleCI may be better if the ambition is beyond just validating build/test results on PRs

brian-brazil:
They both will more than support what we need. I proposed circle as they happen to have an office in my city.

official-sounding:
@brian-brazil  are you waiting on any additional changes to merge this PR? Not sure if I am missing some additional change.

tests/com/carolinarollergirls/scoreboard/DefaultSettingsModelTests.java
brian-brazil: Did ye two decide which way to go on these?

official-sounding: I'm fine with @frank-weinberg's approach with specifically-declared imports instead of `*` imports, I've updated this PR to match.

PR #108, opened 2018-07-15
PT Screen toggle table height and team color from operator

riffreff:
Table height can toggle between full screen height and normal height.
Team color is taken from operator instead of overlay.

2018-07-15 13_50_43-penalty tracker
2018-07-15 13_51_13-penalty tracker

### DIFF
--- a/html/controls/pt/ptcolor.css
+++ b/html/controls/pt/ptcolor.css
@@ -1,10 +1,13 @@
-body {	
+body {	margin:0;padding:0;
 	font-size: 11pt;
-	font-family: -apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,sans-serif;"; 
+	font-family: -apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,sans-serif;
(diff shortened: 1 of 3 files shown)


brian-brazil:
You've got some conflicts there. @official-sounding has already implemented this, and linking it from the index will be in his next PR. We can still take the overlay->operator change though.

riffreff:
According to what I read, he solved it by adding a URL parameter. That means we have two links to the index screen?
I made an actual toggle button, so the user can decide on the PT screen, what layout they prefer.

brian-brazil:
Yes, there'd be two links. We already do this for the scoreboard view.

riffreff:
I find my solution more user friendly, this way penalty trackers can decide on their tablet what view they prefer (especially when crews switch) without going back to the index screen.

The index with all the legacy links is horrible already. Too many options!

brian-brazil:
Your suggestion is an extra button on a view which is already rather cluttered with buttons, so there's a good chance they'd hit it by accident and get confused in a middle of a game. It's also state which would be lost on refresh. A URL parameter has neither of these problems, and is more consistent with the rest of the scoreboard.

PR #109, opened 2018-07-20
Replace Java 1.8 code

frank-weinberg:
String.join() is not available in Java 1.7.
(With this change applied, the server starts up again on 1.7)


brian-brazil:
Thanks!

PR #110, opened 2018-07-21
Fix occasional FileNotFoundException when loading penalty data

official-sounding:
the root exception message is `Caused by: java.io.FileNotFoundException: . (Access is denied)`, which suggests to me that this is a threading-related issue.  I added a synchronized block around the file access portion.  I tried to recreate the exception by rapidly starting new games (which had triggered it in the past) and was unable to.

I haven't observed this causing a functional problem, the codes are always still delivered.  I suspect this is maybe related to a lot of change handlers being triggered by the *Stats classes in the game folder, but that's just a suspicion.

Additionally, since I was editing this code, I went ahead and updated the pattern to a Java 7's try-with-resources (I promise this is actually in Java 7, unlike String.join())

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
+++ b/src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
@@ -319,21 +319,19 @@ public class ScoreBoardJSONListener implements ScoreBoardListener
 	}
 	
 	private void processPenaltyCodes(Settings s) {
-		PenaltyCodesManager pm = new PenaltyCodesManager(s);
 		updates.add(new WSUpdate("ScoreBoard.PenaltyCode", null));
-		
-        try {
-            PenaltyCodesDefinition penalties = pm.loadFromJSON();
-            for(PenaltyCode p : penalties.getPenalties()) {
-                
-                updates.add(new WSUpdate("ScoreBoard.PenaltyCode."+p.getCode(), p.CuesForWS(p)));
-            }
-            
-            
-            updates.add(new WSUpdate("ScoreBoard.PenaltyCode.?","Unknown"));
-        } catch (Exception e) {
-            throw new RuntimeException(e);
-        }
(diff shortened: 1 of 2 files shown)


brian-brazil:
This fix doesn't feel right, this method isn't doing any mutations. If there's an issue here it's in the `settings.get` I'd say. Maybe we're missing a default somewhere at startup?

official-sounding:
You're right, I did a bit more investigation and it does appear that the setting for the penalty file is momentarily an empty string occasionally. It seems to be "usually" fixed before the setting is actually used, and very frustratingly, is never blank when I put a debugger on `settings.get`.  everything appears to be properly synchronized in the DefaultSettingsModel, so if it's a concurrent data access issue I am deeply mystified.

I think my intuition about something in the game folder potentially causing intermittent resets() is probably still right? but for now I've updated the fix to instead just not try to read a penalty file if the settings value in question is null or blank.

src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
```diff
@@ -276,21 +276,19 @@ private void processPosition(String path, Position p, boolean remove) {
...
-        }
+		String file = s.get(PenaltyCodesManager.PenaltiesFileSetting);
+		if(file != null && !file.isEmpty()) {
+			try {
+				PenaltyCodesDefinition penalties = pm.loadFromJSON(file);
+				for(PenaltyCode p : penalties.getPenalties()) {
+					updates.add(new WSUpdate("ScoreBoard.PenaltyCode."+p.getCode(), p.CuesForWS(p)));
+				}
+				updates.add(new WSUpdate("ScoreBoard.PenaltyCode.?","Unknown"));
+			} catch (Exception e) {
```
brian-brazil: What exception is this catching? loadFromJSON doesn't have any checked exceptions.

official-sounding:
removed the extraneous catch block, and rebased up to the top of head for good measure

brian-brazil:
Thanks!

Issue #66, opened 2018-06-02
Add option to whiteboard display to change text color of penalties in the second half


GorramRob:
Feature Request -- add an option to change the font color of penalties incurred in the second half of a game on the whiteboard display.

official-sounding:
a few other style/highlighting tweaks to the PT sheet, from feedback in the FB group:

* Highlight Penalties from the current jam (through to the start of the next jam)
* Highlight Skaters currently in the jam, as selected from the LT/Operator screen
* Make the background of the PT sheet black

AdamSmasherDerby:
As a matter of style, how should we be allowing users to select selectable options?  Ruleset editor? Policies? Config File? Switch on the PT entry page itself?  The actual implementation of the color change doesn't seem that difficult, but the UI design is something that should have some thought applied to it.

bullseye555:
Can I add one to this - an option in Ruleset (or policies or config, whichever path it goes down) to be able to set # penalties in a game. Then when you're running a 20 min half game with reduced number of penalties, the PT and IWB screens will have their colour highlights based on that (eg: set 5 penalties instead of 7, and it will yellow at 3, orange at 4 and red at 5 etc)(can also open another ticket, if deemed too far from the original to be included in it) 

Personally, I'd be adding something like this to the ruleset editor, however it would depend on what other functions are made available, and something like the original request in config

admdikramr:
Visual indication of "Period 1" penalties as different from "Period 2" came up for all three CHNSOs (including myself) at the WFTDA Continental Cup in Kalamazoo last weekend.

Using a different font color is one option. Greying the background is another option (more similar to paper PT).

brian-brazil:
This was done in #112, they're now bold/not-bold.

PR #112, opened 2018-07-21
Penalty Tracking Page Enhancements

### LINKED ISSUE #91: Penalties are recorded on the PT sheet ordered by entry time, not jam number
As reported in the Derby Scoreboard FB Group:

If I enter a penalty for jam 4, then realize that I missed a penalty for jam 3, the jam 3 penalty will show up after the jam 4 penalty.

Ideally, when saving the penalty, they should sort by period/jam number

eg instead of this:
image

it should be this:
image

### LINKED ISSUE #101: Remove the "legacy" PT screen
the actual differences from the "Penalty Tracking" sheet are basically nil at this point (the main difference is the Penalty Tracking Sheet supports colors), so the only purpose the legacy Penalty Tracker screen seems to serve is an opportunity to introduce bugs.

I would like to remove the legacy Penalty Tracking sheet, and make the ptcolor version the "default".  Does any one see a reason to keeping the legacy screen around?

official-sounding:
User-facing changes to the Penalty Tracking page:

1. Use the Operator Team Name & Color, and name/color updates when updated on the WS, not when a skater is updated
2. Make Background dark
3. Penalties committed in the current jam are _underlined_
4. Skaters in the current jam are highlighted green (similar to elsewhere)
5. Sort penalties when added/updated
6. Remove Legacy Penalty Tracking page, add link to "auto-height" version
7. De-emphasize penalties in prior periods (penalties in the current period are *bold*)

Additionally, I've made the following set of backend/developer changes

1. Wrap PT code in an IIFE, enabled strict mode on the PT page
2. fix some conflicts between global and function-local variables are had the same name
3. Added some unit tests to DefaultSkaterModel related to penalty adding/removing/editing

This doesn't technically do what #66 asks, in that it's not an option and it's not color, but I am pretty much alright with something opinionated for now.  I am not totally sure that coloring the prior periods penalties justifies adding a set of operator settings/login functionality/etc on it's own.  Maybe as we build up a few more settings for PT we can re-address.

I had to make a bunch of whitespace changes to index.js to keep it formatted properly which sort of screws up the unified diff, unfortunately.

Closes #91 
Closes #101


tests/com/carolinarollergirls/scoreboard/defaults/DefaultSkaterModelTests.java
```diff
@@ -0,0 +1,134 @@
...
+	private DefaultSkaterModel model;
+	private TeamModel teamMock;
+	private UUID skaterId;
+	
+	@Before
+	public void setup() {
+		teamMock = Mockito.mock(TeamModel.class);
+		skaterId = UUID.randomUUID();
+		
+		model = new DefaultSkaterModel(teamMock, skaterId.toString(), "Test Skater","123","");
```
brian-brazil: Spaces after comma is inconsistent here

src/com/carolinarollergirls/scoreboard/defaults/DefaultSkaterModel.java
```diff
@@ -151,6 +152,7 @@ public void AddPenaltyModel(String id, boolean foulout_explusion, int period, in
         if (penalties.size() < 9) {
           DefaultPenaltyModel dpm = new DefaultPenaltyModel(id, period, jam, code);
           penalties.add(dpm);
+          sortPenalties();
```
brian-brazil: The EVENT_PENALTY is no longer just about that individual penalty, as it could reorder other penalties. Thus I think the event should be considered to be more generic and not pass dpm to make that clearer. This doesn't require code changes in xml/json listeners as they currently replace the entire skater.

brian-brazil (commented):
I'm totally okay with keeping this simple and unconfigurable.

Do you have a screenshot?

official-sounding:
Here's a screenshot of the various UI updates, showing two skaters on the track, with penalties in both the prior period, a prior jam in the same period, and the current jam.

I will push some fixes momentarily to address the code review issues, plus a few bugs related to reloading the penalty tracking page I discovered while making screenshots.
image

src/com/carolinarollergirls/scoreboard/defaults/DefaultSkaterModel.java
official-sounding: I agree that a single penalty object is not ideal, but I also want to avoid the potential pitfalls of not sending anything at all in as the event value (as that would violate the convention that null is a removal action of some sort).  For now I've updated it to be the skater's list of penalties, but I can update it to be something else (like total count of penalties, a string/enum indicating the action type, etc) if you think that's more appropriate.

brian-brazil: sgtm

brian-brazil:
Thanks!

PR #119, opened 2018-07-24
Fix clock sync.

### LINKED ISSUE #113: Clock sync is broken
At current master clock sync is not working, with the period/jam/lineup clocks not advancing on the same second.

~~In addition there was other clock weirdness. In one case at EuroDerbyCon the period clock expired, then the jam clock, then rather than going to intermission the lineup clock started. We also had two cases where it seems the lineup clock either was running during a jam, or something else that caused it to be over a minute just at the end of the jam. The other two issues were intermission related. I'm suspecting that this is related to the clock sync not working, and wondering if the scoreboard is even usable without clock sync.~~ Edit, this is a different bug. See #118

brian-brazil:
Fixes #113 
@frank-weinberg

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
+++ b/src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
@@ -328,7 +328,7 @@ public class DefaultClockModel extends DefaultScoreBoardEventProvider implements
 	}
 
 	protected boolean isSyncTime() {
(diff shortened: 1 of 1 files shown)


frank-weinberg (approved):
My first guess was a wrong reference to a setting. We should probably add logging to the Settings class that is triggered when an unknown setting is referenced. (And for the Java code we should introduce constants to reference them, like it is done for IDs or use enums for that purpose.)

Issue #114, opened 2018-07-24
WS Scoreboard doesn't support team colours


brian-brazil:
As with the XML Scoreboard, team colours should be supported.

We may not wish to support the scoreboard_dots for the sake of simplicity.

PR #121, opened 2018-07-25
Add team colors to WS Scoreboard

AdamSmasherDerby:
Adds support for team colors to the WS version of the scoreboard, closing issue #114 . Due to the somewhat awkward implementation of the "official review retained" symbol, it uses a somewhat awkward implementation of the coloration for said symbol, but it should work.


brian-brazil:
When testing this, the circle is still black when you have an OR available.

AdamSmasherDerby:
Fixed. If you stand *right next* to the screen with a magnifying glass after a retained OR, you might be able to see the seams between the parent box gradient and the gradient behind the OR cross, but it's pretty close.

brian-brazil:
Thanks! The background colour is displayed a bit differently, but I don't think that's really important.

PR #122, opened 2018-07-26
Add unit tests for game state

frank-weinberg:
These tests verify the behaviour of the game state as displayed on the main
scoreboard (except Jammer names).

Not included:
- Undo. This will be added together with the switch to a snapshot model
in a separate PR.
- Game Statistics (Penalties, Lineups)
- Settings other than WFTDA sanctioned (with a few exceptions).

In addition the PR contains fixes for a few minor bugs uncovered by the tests.

PR #123, opened 2018-07-27
Don't set period number to 2 after Time to Derby

frank-weinberg:


src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -595,7 +595,7 @@ public void scoreBoardChange(ScoreBoardEvent event) {
 				//If less than one minute of intermission is left and there is another period, 
 				// start the next period. Otherwise extend the previous period.
 				requestBatchStart();
-				pc.changeNumber(1);
+				pc.setNumber(ic.getNumber()+1);
```
brian-brazil, line 598: Are there any known issues with ic Number being wrong? I don't think we've ever depended on it before.

Do your proposed unittests already cover this?

frank-weinberg: On intermission start ic Number is set to period number. And the display ("Time to Derby", "Intermission", "Unofficial Score"/"Final") is based on intermission number. So a wrong intermission number would be very visible.

And the tests in #122 cover the period number after intermission ends after period 1 and 2. I just didn't check it for Intermission end before period 1, but that's easy to add..

brian-brazil: Okay, let's add that test over there then to keep things straight. I doubt this is all the cause of our problems, but it should fix some ones caused by operator error.

PR #125, opened 2018-07-29
Use snapshots for undo

frank-weinberg:
Whenever one of the clock control buttons is pressed (Start Jam, Stop
Jam/TO, Timeout), a snapshot of the scoreboard state is taken. On undo
the last snapshot is restored and the running clocks are advanced by the
time that has passed since the snapshot was taken.
Un-Start, Un-Stop and Un-Timeout are wrappers for the general undo
function that check if the current snapshot is of the approriate type
and do nothing if it isn't.

This prepares the backend for a single undo button but doesn't add it to
the frontend. The latter is planned as an option, after we have a
better way of handling operator settings than a long row of buttons.

(This is identical to #124 but rebased to current dev HEAD)


brian-brazil:
The general approach seems good, but something feels off about having all the snapshot state hidden off on the side. I feel like it should be more innate to the relevant classes.

It's also inconsistent that the skaters are unbenched, whereas other state is snapshotted.

frank-weinberg:
So you're suggesting having the snapshot distributed across all the objects involved (i.e each Clock stores a ClockSnapshot etc.)? That would introduce the risk of getting inconsistent snapshots, if for some reason the snapshot in one object isn't properly updated. And it will get more complex if we start to add more elements to the snapshot. Especially, if some of those elements or their connections are not persistent (Skater-/PositionModel-Links; maybe we will add penalty clocks in the future.)

Not including the SkaterModel right now was a conscious decision as that class is very much in flow as you are migrating the game data stuff. Once this is somewhat settled, we can look at what data should be part of the snapshot and add that.

brian-brazil:
One thing that stands out to me is that these are in separate files, so as new features are added updating these might be missed. It also seems a bit off that our snapshots are in clear structs, but our day-to-day state on which it's based is not. That also makes me wonder how this might interact with better serialisation in the future (and does this state need to be serialised?), having all Scoreboard state in an easily marshallable form might make it easier to replace XML with JSON.

> Not including the SkaterModel right now was a conscious decision as that class is very much in flow as you are migrating the game data stuff. 

That should be orthogonal, I'm using the Skater interface. I'd more worry about @official-sounding changing how Skaters/Positions work.

frank-weinberg:
So you suggest including the snapshot classes as children of the class they snapshot?

And regarding using the snapshot classes to track current state (and renaming them accordingly): I can see benefits of that but I'd say that's definitely beyond the scope of this PR.

And having re-checked I now notice that you are still keeping the stats as aparallel structure, just closer to the rest of the scoreboard. I'm not sure that's ideal when thinking about problems like tracking per-trip-scores and fixing Jam/Period-Numbers, but that's a discussion for another place and time. I guess I can add the SkaterModel into the snapshots then.

brian-brazil:
Yes, at the least they should be child classes.

> I'm not sure that's ideal when thinking about problems like tracking per-trip-scores and fixing Jam/Period-Numbers, but that's a discussion for another place and time.

Part of that is just getting rid of Game, but I expect it'll also work for SK/LT. Ultimately I want all state under ScoreBoard somewhere, rather than the mix we currently have.

frank-weinberg:
Yay, the first commit that passed circleCI

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -319,74 +313,58 @@ public void timeout(TeamModel team, boolean review) {
...
-				getClockModel(Clock.ID_LINEUP).unstop();
-			if (timeoutClockWasRunning) {
-				setTimeoutOwner(lastTimeoutOwner);
-				setOfficialReview(wasOfficialReview);
-				getClockModel(Clock.ID_TIMEOUT).unstop();
+	protected void createSnapshot(String type) {
+		snapshot = new ScoreBoardSnapshot(this, type);
+	}
+	protected void restoreSnapshot() {
+		relapseTime = System.currentTimeMillis() - snapshot.getSnapshotTime();
```
brian-brazil: How are you unittesting this given that System.currentTimeMillis isn't stubbed out?

brian-brazil (commented):
This generally seems fine.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: I'm doing a plausibilty check on `relapseTime` (>0) and verify that this is the time that is being applied to the clocks. I didn't see a way to make sure that this is the actual elapsed wall lock time.

brian-brazil: https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Gauge.java#L139 is one way to handle it.

frank-weinberg: Wrapping System.currentTimeMillis() into a class? The idea then being that we stub  this class for tests?
That would work but it's a lot of overhead just to test two simple lines of code.

brian-brazil: Yes. That's the standard way to do it, every approach I've seen boils down to pretty much that.

frank-weinberg: Turns out we already have a class that fits this description sufficiently close and is instrumented for unit tests: `UpdateClockTimerTask`. (It returns ms since startup instead of system time but as we only want to measure a difference, this is perfectly fine.)

frank-weinberg:
That last test failure appears to be another manifestation of the random clock weirdness bug. I didn't manage to reproduce it in ~50 runs of the tests on my machine.
Making the listener thread safe was however still a good idea, as before that fix I had random missed events about once in two or three runs.

src/com/carolinarollergirls/scoreboard/model/SkaterModel.java
```diff
@@ -17,17 +17,34 @@
...
-  // A null code removes the penalty.  
-  public void AddPenaltyModel(String id, boolean foulout_explusion, int period, int jam, String code);
+	// A null code removes the penalty.  
+	public void AddPenaltyModel(String id, boolean foulout_explusion, int period, int jam, String code);
 
 	public static interface PenaltyModel extends Penalty {
 	}
+	
+	public class SkaterSnapshot	{
+		public SkaterSnapshot(SkaterModel skater) {
```
brian-brazil: Models shouldn't contain any code, they're just interfaces so should only have methods.

src/com/carolinarollergirls/scoreboard/model/TeamModel.java
```diff
@@ -91,4 +92,45 @@
 
 		public TeamModel getTeamModel();
 	}
+	
+	public class TeamSnapshot {
+		public TeamSnapshot(TeamModel team) {
+			id = team.getId();
```
brian-brazil: Same here, no code in models.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -319,74 +315,58 @@ public void timeout(TeamModel team, boolean review) {
...
+		setTimeoutOwner(snapshot.getTimeoutOwner());
+		setOfficialReview(snapshot.isOfficialReview());
+		setInOvertime(snapshot.inOvertime());
+		setInPeriod(snapshot.inPeriod());
+		snapshot = null;
+	}
+	protected void relapseTime() {
+		for (ClockModel clock : getClockModels()) {
+			if (clock.isRunning()) {
+				clock.elapseTime(relapseTime);
```
brian-brazil: Passing replaseTime between these functions via a field of the object is a bit weird, it's creating an implicit dependency between restoreSnapshot and relapseTime. I think it'd be better to either have one call the other, or to merge the functions.

frank-weinberg: I intentionally seperated them as this will allow implementing a replace feature down the road (e.g. replace "Stop Jam" by "Timeout" when you figure out it was an injury call-off).

src/com/carolinarollergirls/scoreboard/model/SkaterModel.java
frank-weinberg: Putting the Snapshot classes in the implementation would require to include the implementation from the interface, which is even worse. So it's either this or making them seperate, which you disapproved in the first review.

brian-brazil: There's no need to have the implementation in the interface - and none of the other views or models do this. It's also fine to have an inner class that's not mentioned in the Model, if it's not doing to be part of the Model interface.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: replaseTime being a global-ish variable that's passing state behind the scenes is still weird, the dependency should be explicit.

src/com/carolinarollergirls/scoreboard/model/SkaterModel.java
frank-weinberg: It needs to be known to the Model as it's a parameter to `restoreSnapshot`.
Now we could have XxSnapshotInterface in the Model and implement it in the class. But given the simplicty of the snapshot class that's basically copying over the code. Ew. Then I'd rather go back to having the snapshot classes separate. And thinking about it some more, I don't think your argument of "it might be overlooked when updating the class" really applies. There still is `restoreSnapshot` in the class, which tells the programmer that there are snapshots to be considered. (And someone who doesn't notice that would likely also have missed an inner class.)

brian-brazil: Okay, then keep it in the model but remove the implementation from the model. The implementation should go in the default class file.

frank-weinberg: I just wrote that I'd rather not do this and gave a reason. Now just demanding I do it without addressing my reasoning at all is not going to bring us towards a consensus.

brian-brazil: My reasoning is that nowhere in this codebase do interfaces contain implementation, and I'm asking you to keep with that for consistency. Look at PenaltyModel just above this for example.

Your points about the class hierarchy are orthogonal, I'm talking about something else.

frank-weinberg: I got that first part. My point wass that duplicating 70% of the class in the interface is ugly and annoying to maintain. I didn't consider the option of having an Interface that doesn't contain the actual interface. 
So I'll implement the latter when I get home, though I don't consider it a good way of doing it but just less bad than the other options.

brian-brazil: I understand this, there are many aspects of this codebase that are less than optimal. I say we work with what we have and try to keep things consistent, there's already enough sedimentary layers in this codebase that we don't need to add more.

frank-weinberg: ...except that a methodless interface won't work because then the implementation of `restoreSnapshot`, which also gets the interface as parameter, won't be able to access the getter methods (unless we cast to the assumed implementation, which violates type safety).

So I'll revert to my initial implementation of having the Snapshot classes separate. While this means adding a new parameter that should be part of the snapshot requires editing two classes across two files, the other option would require editing two classes and an interface across two files, which is worse.

brian-brazil: It's not a problem to have mehods on the interface. Penalty has methods, but they're in the view interface rather than the model and you can do the same here.

Once again, whether they are inner classes or classes in their own right is completely orthogonal.

frank-weinberg: It's not a problem in the sense that it wouldn't work or violate any design principles. But it would give us the downsides that originally caused the request to have the snapshots as inner classes. So there's no point in having them as inner classes.

brian-brazil: Even as standalone classes, the current design pattern would still require a view interface and a model interface for those classes.

The snapshots are quite tightly bound the the things their snapshotting in, so it feels more natural to me to have them as inner classes. It'd be weird if Penalty and TeamColor is an inner class and Snapshot isn't.

frank-weinberg: As we have seen here, having them as inner classes would require duplicating ~70% of the code or storing a single snapshot distributed over dozens of objects. So the approach currently in this PR is the least bad IMO.
(And the View-Model-Implementation pattern isn't consistently followed anyway.)

brian-brazil: I'm not following you, how does them being inner classes affect code size?

> (And the View-Model-Implementation pattern isn't consistently followed anyway.)

Where is it not being followed? I'll have to add it to my todo list.

frank-weinberg: > how does them being inner classes affect code size?

Ok, once more: We can't have the snapshot as an inner class without declaring its full interface in the corresponding Model. Given the nature of the snapshot classes consisting of just a construcotr and a set of getters, these interfaces will then repeat a large part of the code of the actual class.

> Where is it not being followed? I'll have to add it to my todo list.

- Ruleset and rules
- The event system
- The jetty, json, viewer, and xml packages
- The metrics collector
- The policy package (huh, I thought I removed that o.o)
- ...

Good luck with your todo list.

brian-brazil: > We can't have the snapshot as an inner class without declaring its full interface in the corresponding Model.

Ah, that's what you're talking about. That's normal for this design pattern, it's just something you have to do. I'm surprised Eclipse doesn't have tooling to handle it for you, though I've always done it by hand.

> Good luck with your todo list.

The MVC model is only followed within the ScoreBoard class and its components, as it is the M and V. The rest is the C around the MV. Of those you list only rulesets needs doing, which I've already on my todo list. Teams also need it.

frank-weinberg: But if you view the scoreboard as MVC, the snapshots are not part of the outward face of the model. They are internal to the model (though not to a single class of the model). So they don't need to be part of the model interface.

(I previously thought you were referring to X - XModel - DefaultXModel as the design pattern, labelling X a view, because it contains the readonly part of the interface.)

brian-brazil: > So they don't need to be part of the model interface.

If you can make it work with them not being exposed as part of the X/XModel I'm cool with that, though I'm not sure that's practical.

I propose that you go back to the approach in https://github.com/rollerderby/scoreboard/pull/125/commits/1f549869e758f7683a7dcb1feede2fbdd4a0de2a, but with the implementation of all those methods living to the DefafultXModel.java rather than in X or XModel.java consistent with the existing codebase.

One of the things I'm currently slowly working on with the scoreboard is getting all state to live under ScoreBoard somewhere, getting it into the WS, and very long term getting rid of XML and all that manual handling. Penalties are done and Stats is currently what I'm working on.
These snapshots are state, so they should live with all the other state as they are both used by and use the current state classes. The circular dependency in this snapshot package is also a bit of a smell, these aren't pure utility classes like say ClockConversion.

frank-weinberg: What you want is one package that contains all scoreboard state and has a coherent outward facing interface. But that's a long way from what we have today, with state split across several packages (defalt, model, top level, ...) and each class having a seperate Interface that mixes outward and inward facing methods. Unless we attack that underlying problem there won't be a good solution for the snapshots. And which of the bad solutions is least bad is a matter of taste more than anything else. I don't see why your taste should be more relevant than mine here.

And BTW: This discussion is blocking a possible fix for the random clock weirdnesses and a fix for #48, which is marked as urgent, while we are less than 2 weeks from the first continental cup.

brian-brazil: > What you want is one package that contains all scoreboard state and has a coherent outward facing interface. But that's a long way from what we have today, with state split across several packages (defalt, model, top level, ...) and each class having a seperate Interface that mixes outward and inward facing methods

More precisely, I want all our data under the ScoreBoard interface and following its design patterns. I don't want data living only in the XML (e.g. keybindings) or other ad-hoc designs (rulesets). Once that massive task is done, we can get rid of XML and more importantly all the manual handling that goes with it.

I don't see a reason to change from the default/model/top MVC pattern, and in fact plan to move a lot more of CRG to it. Sure it's a little bit verbose, but that's a price worth paying to have a coherent, clean, and maintainable design.

> Unless we attack that underlying problem there won't be a good solution for the snapshots. 

I am actively attacking the problem. I've done it with penalties, I'm at the tail end of doing it with stats, and after that I'm planning on doing rulesets. Then it's all the state that lives only in XML.

My issue is that your PR takes us further from this goal by adding a brand new design pattern, rather than working with the default/model/top level layout that we already have for the majority of scoreboard state.

I'm not asking you to go all the way and hook this into XML/autosaves/WS (I think we can get away without it for this specific feature), but I do ask you make to this look like Teams, Skaters and Positions do.

> while we are less than 2 weeks from the first continental cup.

We already ran out of time there. The build I've recommended to WFTDA is at the clock sync fix.

In relation to #48, I'm not sure why that's considered urgent. We've survived 9 years without that corner case in the rules being handled, and the traditional workaround everyone has (presumably) been using for years still works.

The clock weirdness I do care about, but any changes to that are high-risk and we simply don't have time to test.

frank-weinberg: The thing is: Your vision of how the code should be structured is at this point just one developers vision. There has never been a discussion in the dev group or anywhere else that I am aware of on this topic.
If you want people to follow that vision, write it up and propose it in the Dev group, so we can have a proper discussion about it with all interested parties. Hopefully we will reach a consensus there that we can then point to in future PRs.
And I propose that this PR is merged, so development can continue. It's not as if adapting the snapshots to any agreed upon code structure will be made significantly harder by this step.

brian-brazil: You're proposing a new design pattern where by we start spreading state around tightly-coupled but distinct packages. As you are the one proposing a change the onus is on you gain support for it, not on me.

> It's not as if adapting the snapshots to any agreed upon code structure will be made significantly harder by this step.

Your proposed PR makes the code harder to understand and maintain in my opinion, and it is ultimately one more special case that'll need to be made not special later on. I'd much prefer we got it right first time around, especially as the previous iteration was quite close and this is a relatively small and self-contained change.

If you're not willing to change the code to be in line with how the existing code is structured, in order to resolve this I'm willing to build off your commits and do it for you.

frank-weinberg: I do believe that your proposed way is harder to maintain. So I am not willing to implement that.

PR #126, opened 2018-07-31
Enable CircleCI

brian-brazil:
Just testing this out atm.

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,0 +1,13 @@
+# We need an image with jdk7 and ant, which CircleCI no longer provides.
+
+FROM openjdk:7-jdk
+
+# From https://github.com/circleci/circleci-images/blob/master/openjdk/generate-images
+RUN curl --silent --show-error --location --fail --retry 3 --output /tmp/apache-ant.tar.gz \
+    https://archive.apache.org/dist/ant/binaries/apache-ant-1.9.6-bin.tar.gz \
+  && tar xf /tmp/apache-ant.tar.gz -C /opt/ \
+  && ln -s /opt/apache-ant-* /opt/apache-ant \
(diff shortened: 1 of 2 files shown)


brian-brazil:
This is all working. I have confirmed this breaks on Java 8 code.

README.md
```diff
@@ -1,3 +1,5 @@
+[![CircleCI](https://circleci.com/gh/rollerderby/scoreboard.svg?style=svg)](https://circleci.com/gh/rollerderby/scoreboard)
```
frank-weinberg: Why is this included here? This file is directed at end users and most end users presumably have no idea what this is about.

brian-brazil: I'm used to the various build indicators being included at the top of a project's readme, as an indication that all is healthy. https://github.com/prometheus/prometheus#prometheus- for example

brian-brazil: What would you like to do here?

This is the fashion and a common way to indicate that a project cares about code quality, but I'm not wedded to it.

frank-weinberg: Given that this file is primarily directed at new users of CRG, many of whom come in with a fear of not being able to install and run it on their own, and that one goal of the file is to help them overcome this fear, I am worried that if the first thing they see in the file is something they don't understand, this will work against this goal. 

But I am not super invested here either. Maybe we should bring this to the dev group to get a few more opinions.

brian-brazil: The current readme duplicates content in the wiki, so overall I'd be inclined to have a big thing at the start pointing users to the wiki and then dev stuff from there in the readme.

frank-weinberg: The readme intentionally duplicates that content as not everybody has internet access in their venues.

brian-brazil: I'm just going to remove this for now.

Depending on place to keep two sets of docs in place is not likely to be very robust, using a single source would be better.

PR #127, opened 2018-07-31
Make WS not static

brian-brazil:
This is will make future tests and changes easier.


brian-brazil:
I've had this commit lying around a while, but had to remove all the state updates from Game first before it'll actually work.

PR #129, opened 2018-08-07
Adjust #125 to better follow existing patterns

brian-brazil:
@frank-weinberg This is more in line with what we already do.

I also had to rebase and fix a bad merge. Yay for unittests catching that.
While testing I discovered our tests are pretty flaky due to races, so put in a small tweak for that too. I haven't seen a failure since.


brian-brazil:
Thanks!

I believe my next change will include the further optimisations you were talking about to the WS handling, as I need to rework all that for other reasons.

PR #130, opened 2018-08-07
Move Clock handling code out of event handlers

### LINKED ISSUE #48: REQUEST: Add button to main interface to restart period clock and lineup clock after an OTO, where there are <:30s left on period clock.

frank-weinberg:
Visible changes:
- Before the start of a period the scoreboard will now display Jam 0 instead of Jam 1. This makes the behaviour consistent with any other time where no Jam is on and the number of the previous Jam is displayed.
- Proper fix to #48

Internal changes:
- The code from the button handling functions and event listeners has
been factored into a private set of methods that clearly spell out what each of
them does and run serialized. Only event listeners that trigger
independent of any button presses have remained and those have been
reduced to a single method call.
- With the exception of the period clock all clocks are reset and the
number increased as they are started. A convenience method has been
added to ClockModel to do this in one call. (Period clock is special as
we reset it when the intermission ends, but only start it at the first
Jam.)
- ScoreBoardModel.stopJam() has been renamed to stopJamTO() to better reflect
its current function.
- ScoreBoardModel.timeout() with parameter(s) has been renamed setTimeoutType() and the code that handles the clocks has been separated from the code that handles ownership as far as currently possible. (Since the "Timeout" button currently doubles to mark a timeout as OTO, some ownership code currently has to reside in the clock handling function. Fixing this will require a UI change.)

Closes #48

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
+++ b/src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
@@ -315,9 +315,11 @@ public class DefaultClockModel extends DefaultScoreBoardEventProvider implements
 	}
 	
 	public void startNext() {
+		requestBatchStart();
 		changeNumber(1);
 		resetTime();
 		start();
+		requestBatchEnd();
 	}
 
 	protected void timerTick(long delta) {
--- a/src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
+++ b/src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
@@ -313,6 +313,12 @@ public class DefaultClockModel extends DefaultScoreBoardEventProvider implements
 			scoreBoardChange(new ScoreBoardEvent(this, EVENT_RUNNING, Boolean.FALSE, Boolean.TRUE));
 		}
 	}
+	
+	public void startNext() {
+		changeNumber(1);
+		resetTime();
+		start();
+	}
 
 	protected void timerTick(long delta) {
 		if (!isRunning())
--- a/src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
+++ b/src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
@@ -140,11 +140,10 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 	}
 	public void startOvertime() {
 		synchronized (runLock) {
-			requestBatchStart();
 			ClockModel pc = getClockModel(Clock.ID_PERIOD);
 			ClockModel jc = getClockModel(Clock.ID_JAM);
 			ClockModel lc = getClockModel(Clock.ID_LINEUP);
-			ClockModel tc = getClockModel(Clock.ID_TIMEOUT);
+
 			if (pc.isRunning() || jc.isRunning())
 				return;
 			if (pc.getNumber() < pc.getMaximumNumber())
@@ -153,10 +152,9 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 				return;
 			createSnapshot(ACTION_OVERTIME);
 			
+			requestBatchStart();
 			setInOvertime(true);
-			if (tc.isRunning()) {
-				_endTimeout();
-			}
+			_endTimeout();
 			long otLineupTime = settings.getLong("Clock." + Clock.ID_LINEUP + ".OvertimeTime");
 			if (lc.getMaximumTime() < otLineupTime) {
 				lc.setMaximumTime(otLineupTime);
@@ -206,11 +204,7 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 	public void timeout() { 
 		synchronized (runLock) {
 			createSnapshot(ACTION_TIMEOUT);
-
-			requestBatchStart();
 			_startTimeout();
-			requestBatchEnd();
-
 			ScoreBoardManager.gameSnapshot();
 		}
 	}
@@ -223,7 +217,7 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 		}
 		setTimeoutOwner(owner);
 		setOfficialReview(review);
-		if (owner != "" && owner != "O") {
+		if (owner != TIMEOUT_OWNER_NONE && owner != TIMEOUT_OWNER_OTO) {
 			restartPcAfterTimeout = false;
 		}
 		requestBatchEnd();
@@ -271,8 +265,8 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 		pc.start();
 		jc.startNext();
 
-		getTeamModel("1").startJam();
-		getTeamModel("2").startJam();
+		getTeamModel(Team.ID_1).startJam();
+		getTeamModel(Team.ID_2).startJam();
 		requestBatchEnd();
 	}
 	private void _endJam(boolean force) {
@@ -283,10 +277,11 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 		
 		requestBatchStart();
 		jc.stop();
-		getTeamModel("1").stopJam();
-		getTeamModel("2").stopJam();
+		getTeamModel(Team.ID_1).stopJam();
+		getTeamModel(Team.ID_2).stopJam();
 		setInOvertime(false);
 
+		//TODO: Make this value configurable in the ruleset.
 		if (pc.getTimeRemaining() < 30000) {
 			restartPcAfterTimeout = true;
 		}
@@ -320,11 +315,12 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 		requestBatchStart();
 		if (tc.isRunning()) {
 			//TODO: Make Official Timeout its own button that calls setTimeoutType()
-			if (getTimeoutOwner()=="") {
-				setTimeoutOwner("O");
+			if (getTimeoutOwner()==TIMEOUT_OWNER_NONE) {
+				setTimeoutOwner(TIMEOUT_OWNER_OTO);
 			} else {
-				setTimeoutOwner("");
+				setTimeoutOwner(TIMEOUT_OWNER_NONE);
 			}
+			requestBatchEnd();
 			return; 
 		}
 		
@@ -334,6 +330,7 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 		_endIntermission(false);
 		setInPeriod(true);
 		tc.startNext();
+		requestBatchEnd();
 	}
 	private void _endTimeout() {
 		ClockModel tc = getClockModel(Clock.ID_TIMEOUT);
@@ -343,7 +340,7 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 		
 		requestBatchStart();
 		tc.stop();
-		setTimeoutOwner("");
+		setTimeoutOwner(TIMEOUT_OWNER_NONE);
 		setOfficialReview(false);
 		if (pc.isTimeAtEnd()) {
 			_possiblyEndPeriod();
@@ -381,7 +378,7 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 		}
 		requestBatchEnd();
 	}
-	private void _checkAutostart() {
+	private void _possiblyAutostart() {
 		ClockModel pc = getClockModel(Clock.ID_PERIOD);
 		ClockModel jc = getClockModel(Clock.ID_JAM);
 		ClockModel lc = getClockModel(Clock.ID_LINEUP);
@@ -422,6 +419,7 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 		setOfficialReview(snapshot.isOfficialReview());
 		setInOvertime(snapshot.inOvertime());
 		setInPeriod(snapshot.inPeriod());
+		restartPcAfterTimeout = snapshot.restartPcAfterTo();
 		snapshot = null;
 		return relapseTime;
 	}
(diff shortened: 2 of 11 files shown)


src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -489,6 +573,7 @@ protected void createTeamModel(String id) {
 	protected Object timeoutOwnerLock = new Object();
 	protected boolean officialReview;
 	protected Object officialReviewLock = new Object();
+	protected boolean restartPcAfterTimeout;
```
brian-brazil, line 574: Does this need to be snapshotted?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -185,144 +179,234 @@ public void startJam() {
...
+	}
+	private void _startTimeout() {
+		ClockModel pc = getClockModel(Clock.ID_PERIOD);
+		ClockModel tc = getClockModel(Clock.ID_TIMEOUT);
+
+		requestBatchStart();
+		if (tc.isRunning()) {
+			//TODO: Make Official Timeout its own button that calls setTimeoutType()
+			if (getTimeoutOwner()=="") {
 				setTimeoutOwner("O");
```
brian-brazil: This convention should be documented somewhere.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -185,144 +179,234 @@ public void startJam() {
...
-	private void _stopTimeout() {
-		synchronized (runLock) {
-			createSnapshot(ACTION_STOP_TO);
-			
-			ClockModel tc = getClockModel(Clock.ID_TIMEOUT);
-			ClockModel lc = getClockModel(Clock.ID_LINEUP);
-			ClockModel pc = getClockModel(Clock.ID_PERIOD);
-			ClockModel ic = getClockModel(Clock.ID_INTERMISSION);
-			
+	public void setTimeoutType(String owner, boolean review) {
```
brian-brazil: This name implies that this method is setting the type of a timeout that has already started, not actually starting a timeout.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -157,17 +153,15 @@ public void startOvertime() {
 				return;
 			createSnapshot(ACTION_OVERTIME);
 			
-			pc.setTime(1000); //TODO: Should not be needed
-			setInPeriod(true);
 			setInOvertime(true);
-			tc.stop();
-			ic.stop();
+			if (tc.isRunning()) {
```
brian-brazil: _endTimeout already checks for this, which you rely on elsewhere

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -103,34 +103,31 @@ public void reset() {
...
-		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_TIMEOUT, "Running", Boolean.TRUE, timeoutStartListener));
-		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_TIMEOUT, "Running", Boolean.FALSE, timeoutEndListener));
 		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_INTERMISSION, "Running", Boolean.FALSE, intermissionEndListener));
-		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_INTERMISSION, "Time", lineupClockListener));
+		addScoreBoardListener(new ConditionalScoreBoardListener(Clock.class, Clock.ID_LINEUP, "Time", lineupClockListener));
 	}
 
 	public boolean isInOvertime() { return inOvertime; }
 	public void setInOvertime(boolean o) {
+		if (o == inOvertime) { return; }
```
brian-brazil, line 130: Do we need to worry about the lc.setMaximumTime getting undone if there's an undo?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -185,144 +179,234 @@ public void startJam() {
...
+		requestBatchStart();
+		ic.stop();
+		if (ic.getTimeRemaining() < 60000 && pc.getNumber() < pc.getMaximumNumber()) {
+			//If less than one minute of intermission is left and there is another period, 
+			// go to the next period. Otherwise extend the previous period.
+			_preparePeriod();
+		}
+		requestBatchEnd();
+	}
+	private void _checkAutostart() {
```
brian-brazil: _possiblyAutostartJam would be clearer and consistent with _possiblyEndPeriod. "Check" implies that this function is stateless.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -185,144 +179,234 @@ public void startJam() {
...
-			if (null==team && tc.isRunning() && getTimeoutOwner() == "") {
+		requestBatchStart();
+		lc.stop();
+		requestBatchEnd();
+	}
+	private void _startTimeout() {
+		ClockModel pc = getClockModel(Clock.ID_PERIOD);
+		ClockModel tc = getClockModel(Clock.ID_TIMEOUT);
+
+		requestBatchStart();
```
brian-brazil, line 315: You don't close this batch, and the early return will mess this up too. This would break WS at the least.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -185,144 +179,234 @@ public void startJam() {
...
 		synchronized (runLock) {
-			ScoreBoardManager.gameSnapshot(true);
-			createSnapshot(ACTION_STOP_JAM);
-
-			ClockModel pc = getClockModel(Clock.ID_PERIOD);
-			ClockModel jc = getClockModel(Clock.ID_JAM);
-			ClockModel lc = getClockModel(Clock.ID_LINEUP);
+			createSnapshot(ACTION_TIMEOUT);
 
 			requestBatchStart();
```
brian-brazil: You use a batch here for a single call that also starts a batch, but not in stopJamTO which does similar.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -185,144 +179,234 @@ public void startJam() {
...
+		jc.stop();
+		getTeamModel("1").stopJam();
+		getTeamModel("2").stopJam();
+		setInOvertime(false);
 
-			requestBatchStart();
-			lc.resetTime();
-			lc.start();
-			requestBatchEnd();
+		if (pc.getTimeRemaining() < 30000) {
```
brian-brazil: Should we use the normal lineup time here rather than hardcoding?

brian-brazil (commented):
This is quite a nice improvement, much clearer what's going on.

How does the "Timeout before period end" function work with all this?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: Which is what it is used for 99% of the time and what is decribed on the Wiki. That the function also starts a timeout if none is running is a convenience feature.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: It's descriebd in the Wiki since 3.9.3.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: This is done by setInOvertime(), which restoreSnapshot calls.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: The "Jam"-suffix however is too specific, as it might also start a timeout.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: In the WFTDA rules it is hardcoded to 30s, not duration of lineup.
And the proper way would be to make it configurable in the ruleset. I'll add a TODO note for when I add the other items I have on my list for the rulesets.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: I don't understand how a function you have just created can be documented on the wiki. This name is still a bit confusing.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: Can you point me to where the time out owner can have the value "O" on the wiki? This is an implementation detail I'd expect to be mentioned in a comment, as otherwise I'd expect it to be "1" or "2".

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: No. Undoing a period preparation (if possible) or a Jam end will always bring us in a state where it will be set again before we can get into a timeout. Changing the timout type is not snapshotted. And no other action sets this variable.
It will however not do any damage to include it and that may save people time wondering if it should be.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: The duration of a lineup is hardcoded to 30s too, so if it's different we're already off the beaten track. Not a problem in practice.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: That's what I suspected, mightn't hurt to do it for consistency.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: Oh, you mean the fact that "O" is OTO. That is indeed not on the Wiki AFAIK.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: This function is triggered when you press one of the Team Timeout or Official Review Buttons on the operator tab. And their function is documented.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: The duration of a lineup can be set with the "Clock.Lineup.Time" rule.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: It's not in the WS (that comes out as bools), so it probably shouldn't be on the wiki as it's internals only of interest to a few of us. A quick comment somewhere, or appropriately named consts, would do the trick I think.

frank-weinberg: It's in the WS in ScoreBoard.TimeoutOwner and is used in the WS display.

frank-weinberg:
That CircleCI failure is interesting. It indicates that the clock is running despite being set to paused in the setup routine.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: Ah, I missed that. That's a todo for me then.

frank-weinberg:
@brian-brazil As far as I am aware, I adressed all your comments. Did I miss anything, you are waiting for or did you just not get around to review again?

brian-brazil:
I didn't see that you'd pushed a commit, GitHub doesn't send notifications about that. I'll get to this in the next day.

src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
```diff
@@ -313,6 +313,12 @@ public void stop() {
 			scoreBoardChange(new ScoreBoardEvent(this, EVENT_RUNNING, Boolean.FALSE, Boolean.TRUE));
 		}
 	}
+	
+	public void startNext() {
+		changeNumber(1);
```
brian-brazil: This should probably have a batch around it

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: Those buttons are documented, this function is not. My point is that this function name on its own is confusing. I'd expect a function called setTimeoutType to have no effect if no timeout was already in progress as various other functions don't do anything if the scoreboard isn't in the right state.

Please choose a better name, like startTimeoutType.

brian-brazil (commented):
Looks like #135 does in fact conflict with this, so I'll wait for this to go in first.

Did you have any other PRs in the works, or could I get the style change in after this?

frank-weinberg:
I have addressed the remaining comments and also found and fixed the source for the failure on the unit tests that looked like clocks were running: When the ClockTimerTask ran for more than 200ms before being paused, clock sync caused the clocks to be modified when started.

brian-brazil:
Hmm, I wonder if we might miss stuff long term if the tests aren't using clock sync - but we can deal with that as is happens.

PR #132, opened 2018-08-07
Improve test runs.

brian-brazil:
Display line numbers for assertion failures, by including debug symbols.
Have clean also remove test artifacts.
Use the same lint settings as for main compiles, this removes a spurious
warning.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -118,26 +118,33 @@
 		</one-jar>
 	</target>
 	
-	 <target name="test-compile" depends="compile">
-	    <mkdir dir="${test.build.dir}"/>
-	    <javac srcdir="${test.src.dir}" destdir="${test.build.dir}" includeantruntime="false">
-	        <classpath refid="classpath.test"/>
-	    </javac>
-	  </target>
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -119,8 +119,14 @@
 	</target>
 	
 	 <target name="test-compile" depends="compile">
-	    <mkdir dir="${test.build.dir}"/>
-	    <javac srcdir="${test.src.dir}" destdir="${test.build.dir}" includeantruntime="false">
+			<mkdir dir="${test.build.dir}"/>
```
frank-weinberg: Can you fix the indentation of this section please?
(And while you're at it maybe the following section as well.)

brian-brazil: That should be all fixed now, and the rest of the file looks okay.

frank-weinberg: The complete block from <mkdir ... /> to < / javac> (ll.122-133) is still indented by one tab too many. The line above it has an erroneous space in front of the tag, which may be throwing your editor off.

brian-brazil: Is this all fine now?

PR #133, opened 2018-08-15
Use the "default" operator by default.

brian-brazil:
Multiple sets of keybindings is a fairly advanced feature,
and asking new users to login as their first experience of CRG
can be a bit overwhelming. Instead use the default user, which
there's a good change they were using anyway as that's what you
get if you hit the X on the login dialog.


frank-weinberg:
Based on the responses in the dev group, I'm not merging this PR at this time.

PR #135, opened 2018-08-16
Fix lockup when overtime is called during a jam.

brian-brazil:
This was due to mismatched batches, so add
general testing for that. Fix one other place
where this was an issue revealed by the new test.


brian-brazil:
This wasn't ready to be merged, my debugging so far indicates that the change to having a Jam 0 state broke the stats code which I'm figuring out how to handle now.

frank-weinberg:
Aw crap. The one time I don't check it out and rerun the tests...
(The earlier version seemed to only fail on unrelated old issues tat have been fixed in HEAD.)

PR #136, opened 2018-08-18
Tcrd foreground fix

AdamSmasherDerby:
Updates the TCRD overlay and whiteboard screens so that the foreground text color uses the "overlay_fg" color.


html/views/wb/tcdg2016wb.js
```diff
@@ -32,50 +31,55 @@ function initialize() {
...
 var penaltyRegex = /Penalty\(([^\)]+)\)/;
-function skaterUpdate(t, k, v) { //arguments: team number, skater id, jam number 
-	match = k.match(skaterIdRegex); //match = skater id cleaned up
+function skaterUpdate(t, k, v) { 
+	match = k.match(skaterIdRegex); 
 	if (match == null || match.length == 0)
 		return;
 	var id = match[1]; // id = skater id
 	var prefix = 'ScoreBoard.Team(' + t + ').Skater(' + id + ')';  //Example: prefix = ScoreBoard.Team('1').Skater('id')
-	if (k == prefix + '.Number') { //Example if skater id == ScoreBoard.Team('team').Skater('id').Number
```
brian-brazil, line 57: This doesn't hurt to keep

html/views/wb/tcdg2016wb.js
```diff
@@ -91,8 +95,9 @@ function displayPenalty(t, s, p) { // team skater penalty#
 		penaltyBoxd.html("&nbsp;");
 	}
 
-	var cnt = 0; // Change row colors for skaters on 5 or more penalties, or explusion.
-  var fo_exp = ($($('.Team' + t + ' .Skater.Penalty[id=' + s + '] .BoxFO_EXP')[0]).data("id") != null);
+	// Change row colors for skaters on 5 or more penalties, or who have been expelled.
+	var cnt = 0; 
+    var fo_exp = ($($('.Team' + t + ' .Skater.Penalty[id=' + s + '] .BoxFO_EXP')[0]).data("id") != null);
```
brian-brazil: indentation is off here

html/views/wb/tcdg2016wb.js
```diff
@@ -123,10 +128,10 @@ function makeSkaterRows(t, id, number) { //team, id, number
 	pd.append($('<div width=10%>').addClass('Number').text(number));
 	pd.append($('<div width=50%>').addClass('Name').text(WS.state['ScoreBoard.Team(' + t + ').Skater(' + id + ').Name']));
 	$.each([1, 2, 3, 4, 5, 6, 7, 8, 9], function(idx, c) {
-		pd.append($('<div width=10%>').addClass('Box Box' + c).html('&nbsp;'));
+		pd.append($('<div width=10%>').addClass('Box Box' + c).html('&nbsp;').css('color',teamFColor));
 	});
-	pd.append($('<div>').addClass('BoxFO_EXP').html('&nbsp;'));
-	pd.append($('<div>').addClass('Total').text('0'));
+	pd.append($('<div>').addClass('BoxFO_EXP').html('&nbsp;').css('color',teamFColor));
+	pd.append($('<div>').addClass('Total').text('0').css('color',teamFColor));
```
brian-brazil, line 134: Is this the best way to handle this?

The existing colour handling code is up around line 38, and we should try to keep with that.

AdamSmasherDerby: The existing color handling code at line 38 just assigns the colors to a custom class.  Attempting to assign the custom class to the partial row where the penalty data lives breaks the appearance pretty badly.  I'm open to other suggestions, but I've not been able to figure out another way to do this that doesn't look *terrible*

html/views/wb/tcdg2016wb.js
AdamSmasherDerby: It's wrong, though.  I pulled all my old comments because they were largely incorrect. I'll put in a more coherent comment instead.

html/views/wb/tcdg2016wb.js
AdamSmasherDerby: It's not showing up this way in Eclipse, so I'm a little puzzled.

html/views/wb/tcdg2016wb.js
brian-brazil: Could you use your approach from https://github.com/rollerderby/scoreboard/blob/dev/html/views/standard/index.js#L87 ?

AdamSmasherDerby: That's kind of overkill.  The reason it's done that way in the SB is because it has to deal with the :before and :after pseudoclasses which can't be directly modified from the JS.  If I could have used the approach I've proposed here for the SB, I totally would have.

brian-brazil: Looking through the other dashboards, we appear to have a mix of techniques for this.

We should probably look at adding a utility function to manage the css classes for this, the XML side of the house has something vaguely in that area.

PR #138, opened 2018-08-21
Improve usability of the lineup tracker screen

frank-weinberg:
- Use roster numbers instead of skater names as main identifier
- Sort skaters by number by default
- Order positions as Jammer, Pivot, Blocker

### DIFF
--- a/html/controls/lineups.css
+++ b/html/controls/lineups.css
@@ -25,12 +25,12 @@ div#tableDiv table td.Position,div#tableDiv table td.Box {
 	padding-bottom: 3px;
 }
 
-div#tableDiv table th.Name { width: 15%; }
 div#tableDiv table th.Number { width: 10%; }
+div#tableDiv table th.Name { width: 15%; }
 div#tableDiv table th.Bench { width: 15%; }
+div#tableDiv table th.Jammer { width: 15%; }
 div#tableDiv table th.Pivot { width: 15%; }
 div#tableDiv table th.Blocker { width: 15%; }
-div#tableDiv table th.Jammer { width: 15%; }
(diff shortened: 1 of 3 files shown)


brian-brazil:
Do you have a screenshot?

frank-weinberg:
screenshot_20180822_001748

frank-weinberg:
Current version for comparison:
screenshot_20180822_002010

brian-brazil:
That's an improvement alright. Can we get rid of the ^v too?

frank-weinberg:
I wouldn't remove them altogether as they serve to indicate that the table can be sorted by these columns. But it may be possible to replace them by proper arrow glyphs.

brian-brazil:
Given it's now sorted by skater number, do we really need any other sorting options?

frank-weinberg:
They are automatically included by using `sortedtable.js`. Reimplementing part of that in order to remove functionality sounds like a bad use of time.

frank-weinberg:
Screenshot with modified headers:
screenshot_20180822_005129

frank-weinberg:
The CircleCI fail is due to the clock sync issue, the fix for which is not included in this branch.

brian-brazil:
I think longer term we should be looking at getting rid of them. They're UI clutter on a screen which needs to be used under time pressure, and a mis-press that could cause problems could easily happen.

frank-weinberg:
The screen will need to be modified anyway, when we start tracking Box entries and exits for statsbook-generation. (We need to distinguish a cancelled misclick from an actual entry/exit pair, a way to handle substitutes and probaly other edge case handling.)

PR #140, opened 2018-08-21
Allow for jam number being 0, which #130 introduced and #135 caught.

brian-brazil:


brian-brazil:
Looks like failures are as Java 7 doesn't like some of our code, and the CI is now checking for that.

brian-brazil:
Tests are now passing on Java 7.

frank-weinberg (approved):

PR #141, opened 2018-08-22
Store operator settings with login

frank-weinberg:
Also move them from individual buttons into a popup dialog, in order to
allow more settings to be added without cluttering the interface.

Add frontend elements for single button undo and add it as default
option for undo.

### DIFF
--- a/html/controls/operator.css
+++ b/html/controls/operator.css
@@ -52,6 +52,7 @@ table.RowTable { border-spacing: 0px; }
 	#TeamTime table.JamControl>tbody>tr>td { width: 100%; }
 	#TeamTime table.JamControl>tbody>tr:first-of-type>td { border-radius: 12px; }
 	#TeamTime .UndoControls:not(.ShowUndo) { display: none; }
+	#TeamTime .SingleUndoControl:not(.ShowUndo) { display: none; }
 
 	#TeamTime table.Team>tbody>tr>td { width: 50%; }
 	#TeamTime table.Team tr.Name td.Name div { max-height: 100px; max-width: 300px; overflow: hidden; font-size: 150%; }
--- a/html/controls/operator.css
+++ b/html/controls/operator.css
@@ -51,6 +51,7 @@ table.RowTable { border-spacing: 0px; }
 
 	#TeamTime table.JamControl>tbody>tr>td { width: 100%; }
 	#TeamTime table.JamControl>tbody>tr:first-of-type>td { border-radius: 12px; }
+	#TeamTime table.JamControl .Label { display: inline-block; min-width:50px; }
 	#TeamTime .UndoControls:not(.ShowUndo) { display: none; }
 	#TeamTime .SingleUndoControl:not(.ShowUndo) { display: none; }
 
--- a/html/controls/operator.js
+++ b/html/controls/operator.js
@@ -401,7 +401,7 @@ function createPeriodEndTimeoutDialog(td) {
 
 function createOvertimeDialog() {
 	var dialog = $("<div>");
-	$("<span>").text("Overtime can only be started at the end of Period ").appendTo(dialog);
+	$("<span>").text("Note: Overtime can only be started at the end of Period ").appendTo(dialog);
 	$sb("ScoreBoard.Clock(Period).MaximumNumber").$sbElement("<span>").appendTo(dialog);
 	$("<button>").addClass("StartOvertime").text("Start Overtime Lineup clock").appendTo(dialog)
 		.click(function() {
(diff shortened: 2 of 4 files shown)


brian-brazil:
I've given this a spin, and I'm not in favour of these changes. The solution to the UI having too much complexity is not to add more complexity on top of it, it's to make things simpler. I think we need to decide on a way to go, rather than humming and hawing and ending up with options for everything.

Speed score controls have been there for 3 years now, let's just make them always there.

Similarly with undo controls, they've been enabled by default since 3.2.2. Let's do away with the option.

With the start/stop button we've gone from a situation where it took a single click to now needing five clicks. This is unnecessary, and quite frustrating if you need to adjust clocks such as if a game is to start early. I think we should go back to always showing them like we used to. Maybe we can do away with them in future, but we're not there yet.

I'm not sure if the single undo is a good idea, as you don't know what you're undoing. I think we should get community input on this, and then implement exactly one outcome with no configurability.

I don't think autostart belongs as a per-operator settings. That's a tournament-level policy to be set by the THNSO as it affects gameplay, not a personal UI preference of the SBO like keybindings are.

frank-weinberg:
With a user base as large and as diverse as ours, having a one size fits all solution is unrealistic IMO. See the dumpster fires regarding which clock to show between timeouts and the next Jam, or yesterday's suggestion regarding the order of penalty codes, or the recurring discussions regarding the balance between allowing experienced users to do their job more efficiently and not overwhelming new users.

Regarding individual settings: 
I have suggested making speed score always there a few months ago in the dev group. it was shot down, as some people still run the scoreboard on small screens.

Single undo: I had dynamic labels for the jam control buttons implemented for #60 and plan to reintroduce them soon. But as there most likely will be people that want to turn that off, I wanted to be able to introduce a setting for it.

Autostart: If the scoreboard displays a Jam start in error, this does not have an effect on the game (the Jam is started by the whistle and hand signal from the JT), but the operator will have to undo/fix it. Thus I consider it an operator decision, if they want to try saving a key press by betting on the Jam starting on time.

brian-brazil:
> With a user base as large and as diverse as ours, having a one size fits all solution is unrealistic IMO

That doesn't mean that we should make no attempts to limit the scope of the software and how many different ways it can be used. Every additional way to do things is more user support, more confusion, more complexity, more gotchas, and more maintenance effort.

> it was shot down, as some people still run the scoreboard on small screens.

The operator screen is already longer than it is wide, the extra vertical pixels shouldn't make a difference to fitting everything on screen. And from a quick test it all fits on 800x600 (640x480 doesn't work at all). 

> But as there most likely will be people that want to turn that off, I wanted to be able to introduce a setting for it.

This is an attitude I'd like to avoid. Just because some users ask for something doesn't mean we must implement it exactly as requested, or to always preserve old ways of doing things. Rather I say we find which way is best overall, and implement it that single way.

Personally I suspect that single undo is a bad idea in practice, as it requires operators to have to do additional thinking to determine if it's safe in a case where they've already messed up. However let's get input on this from the community.

If there's no clear outcome from the community, then I say we leave it as-is.

> If the scoreboard displays a Jam start in error, this does not have an effect on the game

The rules require the correct jam clock to be shown. Thus this has game impact, and anything with game impact should come from the ruleset.

frank-weinberg:
> Personally I suspect that single undo is a bad idea in practice, as it requires operators to have to do additional thinking to determine if it's safe in a case where they've already messed up.

If there are 3 undo buttons, the operator will have to figure out, which of those is currently active.

> However let's get input on this from the community.

And I am pretty sure that we will get better input, if the community has the chance to try out both versions themselves, which gives another reason to have the option to switch between them.
We should keep in mind that this is a development version, which should be about testing new features. We can still discuss about removing options and choosing defaults when we are getting closer to release.

> The rules require the correct jam clock to be shown. Thus this has game impact, and anything with game impact should come from the ruleset.

Operator errors that cause the wrong clocks to be shown momentarily happen all the time. We cannot avoid that. I don't see why we should treat it different if the error comes from an action that the operator automated instead of performing it themselves.

We also have a precedent, where the socreboard switches clocks on it's own based on the assumption that the JT should be blowing their whistle: When the Jam clock has run down. Should we make that depend on a rule as well?

brian-brazil:
> If there are 3 undo buttons, the operator will have to figure out, which of those is currently active.

Do we have any reports of that being a problem? Might users be afraid to use a single undo if they don't know what it's going to undo? Let's ask the community.

> We should keep in mind that this is a development version, which should be about testing new features. 

"Development version" has little meaning for CRG, considering the last official release was well over a year ago and big tournaments have been using versions that aren't any form of release. As it currently stands we're basically running off master, with no notion of a development version. 

Development is not an excuse for half completed features to be included either, especially given that even here we can't agree the final fate of features added 3+ years ago. That's how cruft builds up.

I'm hoping to get an actual release out pretty soon, it's been way too long since we've had one and a lot of changes have built up.

> I don't see why we should treat it different if the error comes from an action that the operator automated instead of performing it themselves.

The difference is that the operator is looking at the jam timer, the scoreboard is not.

Allowing the scoreboard to make a mistake like that is something that should be specified by a THNSO for consistency, not something that should vary by SBO.

> We also have a precedent, where the socreboard switches clocks on it's own based on the assumption that the JT should be blowing their whistle

That's orthogonal, we're discussing if this is a tournament/rule-level policy or a per-operator preference. The argument you're making here is about whether we should have this feature at all, which isn't being debated.

> Should we make that depend on a rule as well?

It already does depend on a rule, the maximum time of the jam clock.

frank-weinberg:
I have added the dynamic labels for jam control buttons (without an option to disable for now).
However, in testing I noticed that the operator settings are not properly restored on reloading and the labels are not properly set after Start new game. In both cases it looks like either the event engine or the XML code swallows updates (possibly when a clear and a set follow closely after each other and the events are processed together).

I will not have time to look into this any further until after A Coruna playoffs.

> Allowing the scoreboard to make a mistake like that

The mistake is made by the JT, not the scoreboard. The scoreboard merely changes the effect of the mistake.

> It already does depend on a rule, the maximum time of the jam clock.

The duration of Lineup is a rule as well, so that does not distinguish the two. The only difference is that many years back someone decided to enable AutoStop and disable AutoStart by default.

PR #144, opened 2018-08-25
Pass skater not the team on removal.

brian-brazil:
This fixes an exception in the xml/json listener.


brian-brazil:
Now part of #146

PR #149, opened 2018-09-12
Events redo

brian-brazil:
This switches event delivery to be largely synchronous, and also in-order. Only one event is executed at a time.

Async is kept for the scoreboard itself, to avoid deadlocks, and when talking to systems like Twitter/XML/WS who may stall.

This is simpler, more predictable, and also happens to halve the length of the unittests.

@frank-weinberg


frank-weinberg:
Did you actually encounter deadlocks in DefaultScoreBoardModel or are you just being cautious?

brian-brazil:
I encountered deadlocks immediately in actual usage (but not in unittests).

Twitter/XML/WS is me being cautious. Twitter/XML I don't know enough about how they work, WS can run out of kernel TCP buffer space.

frank-weinberg:
Without knowing the specific cause of the deadlocks, I'm somewhat worried that DefaultTeamModel or DefaultClockModel could also be vulnerable to them. Both classes also have functions that are called from the frontend and trigger events which are in turn sent back to the frontend. (And having two threads do this with an unlucky ordering of locks sounds like a candidate for the source of the deadlocks.)

brian-brazil:
Neither DefaultClockModel nor DefaultTeamModel have a listener. DefaultTeamModel just passes events up to the DefaultScoreBoardModel, as it's the clearing house for events.

I believe the deadlocks are due to the various ad-hoc locking patterns in use, and running all the events async as I'm doing in DefaultScoreBoardModel should prevent any issues.

frank-weinberg (commented):
Aside from the concerns regarding deadlocks, I have no objections to the change. And while those concerns are sufficient for me to not merge the PR, I don't object, if you want to do it yourself.

brian-brazil:
Thanks! I don't think there's any more risk than what we have today, in fact I'd expect it to be a bit less. We'll see at British Champs this weekend anyway.

PR #152, opened 2018-09-16
Use a single lock for all core classes.

frank-weinberg:
In order to avoid potential for deadlocks, cover all the core logic in
Default*Model with a single lock. This could make reaction times a bit
slower, if multiple clients send state updates at the same time but as
long as all updates are handled quickly, this should be negligible. (We
should, however, not start to compute complex stats under the lock.)

Also make DefaultScoreBoardModel deliver events synchronously. 

The only remaining calls from code under the lock to code (potentially)
under different locks are to the baseLock in Ruleset.java, which in turn
does not reach other locks, and through the event interface. Since all
event listeners not covered by the coreLock are async and thus the
actual handling of the event will not be done while holding the
coreLock, we can conclude that this change and #149 together do not
introduce any potential deadlocks.


src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
```diff
@@ -242,45 +263,53 @@ public void setMinimumTime(long ms) {
 		}
 	}
 	public void changeMinimumTime(long change) {
-		synchronized (timeLock) {
+		synchronized (scoreBoardModel) {
```
brian-brazil: coreLock

src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
```diff
@@ -242,45 +263,53 @@ public void setMinimumTime(long ms) {
...
+		synchronized (coreLock) {
 			setMaximumTime(maximumTime + change);
 		}
 	}
 	public boolean isTimeAtStart(long t) {
-		if (isCountDirectionDown())
-			return t == getMaximumTime();
-		else
-			return t == getMinimumTime();
+		synchronized (scoreBoardModel) {
```
brian-brazil: coreLock

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -568,44 +581,24 @@ protected void createTeamModel(String id) {
 		scoreBoardChange(new ScoreBoardEvent(this, EVENT_ADD_TEAM, model, null));
 	}
 
-	// Have all events delivered by the ScoreBoard asynchronously.
```
brian-brazil, line 572: This should be mentioned in the change description.

brian-brazil (commented):
Is there any reason not to move ruleSet over to coreLock? It seems a bit odd that it's different.

frank-weinberg:
In my perception Ruleset is not part of the scoreboard core. The core classes (Default*Model) maintain the state of the scoreboard, Ruleset does not contain any state.

Everything else is addressed.

brian-brazil:
Well I'll be moving RuleSet into the core when I'm cleaning that up in a bit anyway.

PR #153, opened 2018-09-17
Wrap system clock for modifying in tests and undo

frank-weinberg:
instead of manipulating the internals of DefaultClockModel


src/com/carolinarollergirls/scoreboard/utils/ScoreBoardClock.java
```diff
@@ -0,0 +1,66 @@
+package com.carolinarollergirls.scoreboard.utils;
+
+import com.carolinarollergirls.scoreboard.defaults.DefaultScoreBoardEventProvider;
```
brian-brazil: I don't think we should have a circular dependency between packages like this, this class should go in defaults.

src/com/carolinarollergirls/scoreboard/utils/ScoreBoardClock.java
```diff
@@ -0,0 +1,66 @@
...
+	private ScoreBoardClock() {
+		offset = System.currentTimeMillis();
+	}
+	
+	public String getProviderName() { return "ScoreBoardClock"; }
+	public Class<ScoreBoardClock> getProviderClass() { return ScoreBoardClock.class; }
+	public String getProviderId() { return ""; }
+
+	public static ScoreBoardClock getInstance() {
+		if (instance == null) {
```
brian-brazil: If you're doing this you need locking.

A static initialiser might be simpler.

src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
```diff
@@ -386,15 +388,16 @@ private DefaultClockSnapshotModel(ClockModel clock) {
...
 		public static final String PROPERTY_INTERVAL_KEY = DefaultClockModel.class.getName() + ".interval";
 		private static long update_interval = DefaultClockModel.CLOCK_UPDATE_INTERVAL;
 
 		public UpdateClockTimerTask() {
 			try {
 				update_interval = Integer.parseInt(ScoreBoardManager.getProperty(PROPERTY_INTERVAL_KEY));
 			} catch ( Exception e ) { }
-			startSystemTime = System.currentTimeMillis();
+			startSystemTime = scoreBoardClock.getCurrentTime();
+			scoreBoardClock.addScoreBoardListener(this);
```
brian-brazil: Currently all listens are via the DefaultScoreBoardModel, so this should work the same way for consistency.

src/com/carolinarollergirls/scoreboard/utils/ScoreBoardClock.java
```diff
@@ -0,0 +1,66 @@
...
+		long last = currentTime;
+		currentTime += ms;
+		scoreBoardChange(new ScoreBoardEvent(this, EVENT_MANUAL_CHANGE, currentTime, last));
+	}
+	
+	public synchronized void stop() {
+		updateTime();
+		stopCounter++;
+	}
+	public synchronized void start(Boolean doCatchUp) {
```
brian-brazil: `boolean`

brian-brazil (commented):
Using events for this seems a bit weird, I think this logic would better live either in DefaultClockModel or have all this pass via ScoreBoardModel.

frank-weinberg:
Using events felt weird when writing the code. Your comments pushed me towards using a different approach.

src/com/carolinarollergirls/scoreboard/utils/ScoreBoardClock.java
```diff
@@ -0,0 +1,88 @@
...
+	private int stopCounter = 0;
+	private long lastRewind = 0;
+	
+	private static Timer timer = new Timer();
+
+	private static final ScoreBoardClock instance = new ScoreBoardClock();
+	
+	private List<ScoreBoardClockClient> clients = new ArrayList<ScoreBoardClockClient>();
+	
+	public static final long CLOCK_UPDATE_INTERVAL = 200; /* in ms */
```
brian-brazil, line 83: Why not set this to 50 instead of dividing?

There's also an override of this value in the android code that should be deleted.

src/com/carolinarollergirls/scoreboard/utils/ScoreBoardClock.java
```diff
@@ -0,0 +1,88 @@
...
+		stopCounter++;
+	}
+	public synchronized void start(boolean doCatchUp) {
+		if (!doCatchUp) {
+			offset = System.currentTimeMillis() - currentTime;
+		}
+		stopCounter--;
+	}
+	
+	public void registerClient(ScoreBoardClockClient client) {
```
brian-brazil: This should be synchronized

src/com/carolinarollergirls/scoreboard/utils/ScoreBoardClock.java
```diff
@@ -0,0 +1,88 @@
...
+	private static Timer timer = new Timer();
+
+	private static final ScoreBoardClock instance = new ScoreBoardClock();
+	
+	private List<ScoreBoardClockClient> clients = new ArrayList<ScoreBoardClockClient>();
+	
+	public static final long CLOCK_UPDATE_INTERVAL = 200; /* in ms */
+
+	public interface ScoreBoardClockClient {
+		void updateTime(long ms);
```
brian-brazil: A comment here mentioning this is passed the current time rather than an offset would be useful.

src/com/carolinarollergirls/scoreboard/utils/ScoreBoardClock.java
```diff
@@ -0,0 +1,88 @@
...
+			updateClients();
+		}
+	}
+	
+	private long offset;
+	private long currentTime;
+	private int stopCounter = 0;
+	private long lastRewind = 0;
+	
+	private static Timer timer = new Timer();
```
brian-brazil: Why is this static?

brian-brazil (commented):
This looks better.

src/com/carolinarollergirls/scoreboard/utils/ScoreBoardClock.java
frank-weinberg: UpdateClockTimerTask uses the unmodified value as tick interval.
(everything else should be addressed in the new commit)

PR #154, opened 2018-09-17
Slightly reorganize java packages

frank-weinberg:
move utility classes to new 'utils' package
move base classes of X XModel DefaultXModel triplets to new 'base' package


brian-brazil:
The base classes are all Views, in MVC terms so "View" would seem a better name.

I'd expect exceptions to live in the same package as where they're defined, not in a random utility class.

ScoreBoardMetricsCollector is not a utility, it belongs in Viewer.

frank-weinberg:
changed.

PR #155, opened 2018-09-19
Overhaul jam & timeout controls

frank-weinberg:
- Make jam control buttons display their current function
- Use a single undo button (and remove the option to hide it)
- Move setting timeout type to Official timeout to a new button instead of reusing the start timeout button
- Selecting a timeout type during a running timeout will change the type (and credit back the timout to the previous owner, if applicable)
- Pressing the "Timeout" button again will start a new timeout (e.g. if a TTO is followed by an OTO)
- Team Timeout buttons are labelled "Team TO" instead of "Timeout"

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
+++ b/src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
@@ -124,7 +124,8 @@ public class DefaultScoreBoardModel extends DefaultScoreBoardEventProvider imple
 			setLabel(BUTTON_START, ACTION_START_JAM);
 			setLabel(BUTTON_STOP, ACTION_LINEUP);
 			setLabel(BUTTON_TIMEOUT, ACTION_TIMEOUT);
-			setLabel(BUTTON_UNDO, ACTION_NONE);		}
+			setLabel(BUTTON_UNDO, ACTION_NONE);
+		}
 	}
 
 	public boolean isInPeriod() { return inPeriod; }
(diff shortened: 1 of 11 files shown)


src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -114,12 +114,17 @@ public void reset() {
...
 			settings.reset();
 			stats.reset();
 			// Custom settings are not reset, as broadcast overlays settings etc.
 			// shouldn't be lost just because the next game is starting.
-		}
+			
+			setLabel(BUTTON_START, ACTION_START_JAM);
+			setLabel(BUTTON_STOP, ACTION_LINEUP);
+			setLabel(BUTTON_TIMEOUT, ACTION_TIMEOUT);
+			setLabel(BUTTON_UNDO, ACTION_NONE);		}
```
brian-brazil: } is misplaced

brian-brazil (commented):
I've played around and noticed a few things:

- Undoing a timeout doesn't reset the OR/TO count
- When there's no action, the buttons look weird. I'd suggest having some text here.
- The Official TO button is oddly placed - can things be kept symmetrical? Maybe put it up top?

frank-weinberg:
I prefer inactive buttons to be recognizable as such at a glimpse. Having actual text on them would counter that. I've added "---" now, so they're not completely empty.

OTO button only makes sense in the row where all the TO type buttons are placed, IMO. I also don't see why it is unsymmetrical to have a button in the center.

brian-brazil:
That all looks fine now. Can you add unittests for the OR/TO count?

I also just noticed that on restart the undo button is the previous value, we should probably initialise it to NONE in that case given that it doesn't work.

frank-weinberg:
timout count on switchint type was already in. I added timeout count on undo now as well (plus that button init).

PR #156, opened 2018-09-20
Overhaul view settings

frank-weinberg:
- store them in FrontendSettings, so they are kept across games
- add option to display a whiteboard
- add option to en-/disable clock sync
- add option to choose the clock displayed after a timeout
- fix background to solid black
- remove option to disable per jam scores
- also separate settings for ddtv and interactive overlays

### DIFF
--- a/html/controls/operator.js
+++ b/html/controls/operator.js
@@ -952,7 +952,7 @@ function createScoreBoardViewContent(table) {
 	});
 	var applyPreviewButton = $("<button>Apply Preview</button>").button()
 		.click(function() {
-			$sb("ScoreBoard.Settings").find("Setting").each(function() {
+			$sb("ScoreBoard.FrontendSettings").find("Setting").each(function() {
 				oldPath = String($sb(this).$sbPath);
 				var path = oldPath.replace("(ScoreBoard.Preview_", "(ScoreBoard.View_");
 				if (path != oldPath) {
@@ -962,7 +962,7 @@ function createScoreBoardViewContent(table) {
 			});
 		});
 	// var viewOptions = $sb("Pages.Page(scoreboard.html)").children("PreviewOptions,ViewOptions");
-	_crgUtils.bindAndRun("ScoreBoard.Settings.Setting", "sbchange", function() {
+	_crgUtils.bindAndRun("ScoreBoard.FrontendSettings.Setting", "sbchange", function() {
 		var disableApplyButton = true;
 		$sb("Pages.Page(scoreboard.html).PreviewOptions").find("*").each(function() {
 			var viewPath = String($sb(this).$sbPath).replace("PreviewOptions","ViewOptions");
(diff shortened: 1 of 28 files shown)


html/controls/operator.js
```diff
@@ -1026,36 +1027,41 @@ function createScoreBoardViewPreviewRows(table, type) {
...
+	var intermissionControlButton = $("<button>Intermission Labels</button>").button().addClass("ui-button-small")
 		.click(function() { intermissionControlDialog.dialog("open"); });
-	var swapTeamsButton = $("<label/><input type='checkbox'/>");
-	$sb("ScoreBoard.Settings.Setting(ScoreBoard." + type + "_SwapTeams)").$sbControl(swapTeamsButton, { sbcontrol: {
+	var clockAfterTimeout = $sb("ScoreBoard.FrontendSettings.Setting(ScoreBoard.ClockAfterTimeout)").$sbControl("<label>Clock shown after Timeout: </label><select>", { sbelement: {
+		prependOptions: [
+			{ text: "Lineup", value: "Lineup" },
+			{ text: "Timeout", value: "Timeout" }
+		]}});
+	var syncClocksButton = $("<label/><input type='checkbox'/>");
```
brian-brazil, line 1051: Should we just hardcode this to on?

src/com/carolinarollergirls/scoreboard/view/FrontendSettings.java
```diff
@@ -16,5 +16,5 @@
 	public Map<String, String> getAll();
 	public String get(String k);
 
-	public static final String EVENT_SETTING = "FrontendSetting";
```
brian-brazil, line 19: This change will break the XML, it should be left as-is.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -351,7 +326,9 @@ private void _endTimeout() {
 		if (!tc.isRunning()) { return; }
 		
 		requestBatchStart();
-		tc.stop();
+		if (!frontendSettings.get(FRONTEND_SETTING_CLOCK_AFTER_TIMEOUT).equals(Clock.ID_TIMEOUT)) {
```
brian-brazil, line 349: Having frontend settings affect clocks like this doesn't feel right. Is there a way we could do this consistently in the backend, and have the fronted do the main logic?

Having it this way also means if it's set that timeout clock is running, then there's no feedback/encouragement for the operator to actually end the timeout - which won't get us towards consistent practices.

html/controls/operator.js
```diff
@@ -1064,28 +1070,21 @@ function createScoreBoardViewPreviewRows(table, type) {
 			{ text: "8%", value: "8" },
 			{ text: "10%", value: "10" }
 		]}});
-	var backgroundStyle = $sb("ScoreBoard.Settings.Setting(ScoreBoard." + type + "_BackgroundStyle)").$sbControl("<label>Background Style: </label><select>", { sbelement: {
```
brian-brazil, line 1081: The background is still showing as black to white.

html/views/scoreboard.css
```diff
@@ -525,17 +494,6 @@ div.WhiteBox {
 
 /* http://net.tutsplus.com/tutorials/html-css-techniques/quick-tip-understanding-css3-gradients/ */
```
brian-brazil: This can go

html/controls/operator.js
frank-weinberg: Where? I have solid black on both legacy and WS view.

brian-brazil: Weird, it's showing as all black now - even though I hard refreshed several times before.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: I understand the feeling. But the only other option I see would be to have both clocks running in the backend and madify all the frontends (including ones, we don't maintain) to choose the clock for display based on this setting. That's a lot of work just to please some loud people who don't understend "the end of a timeout is marked by a long rolling whistle".

And given that the requested functionality explicitly is to not change the displayed clock, I'm not sure how to give more feedback beyond unsetting timeout type.

html/controls/operator.js
frank-weinberg: I'm not opposed to further thinning out options. But for now this would just leave an ugly gap in the settings grid.

src/com/carolinarollergirls/scoreboard/view/FrontendSettings.java
frank-weinberg: It was easier to change the XML than all of the frontend at this point. And I think "FrontendSettings.FrontendSetting(...)" is a bit unwieldy.

PR #167, opened 2018-09-23
rework operator settings

frank-weinberg:
- Make speed score controls always visible
- store remaining option (start/stop buttons) per operator

### DIFF
--- a/html/controls/operator.js
+++ b/html/controls/operator.js
@@ -140,6 +140,7 @@ function createScoreTimeContent(table) {
 
 function setClockControls(value) {
 	$("#ShowClockControlsButton").checked = value;
+	$("label.ShowClockControlsButton").toggleClass("ui-state-active", value);
 	$("#TeamTime").find("tr.Control").toggleClass("Show", value);
 }
 
@@ -168,7 +169,7 @@ function createMetaControlTable() {
 	$("<a>").text("Key Control Edit mode enabled.	 Buttons do not operate in this mode.	 Move the mouse over a button, then press a normal key (not ESC, Enter, F1, etc.) to assign.")
 		.appendTo(helpTd);
 
-	$("<label>").text("Show Start/Stop Buttons").attr("for", "ShowClockControlsButton")
+	$("<label>").addClass("ShowClockControlsButton").text("Show Start/Stop Buttons").attr("for", "ShowClockControlsButton")
 		.appendTo(buttonsTd);
 	$("<input type='checkbox'>").attr("id", "ShowClockControlsButton")
 		.appendTo(buttonsTd)
--- a/html/controls/operator.js
+++ b/html/controls/operator.js
@@ -48,7 +48,7 @@ function setOperatorSettings(op) {
 		// the latter currently happens to be.
 		prefix = "ScoreBoard.FrontendSettings.Setting(ScoreBoard.Operator_Default.";
 	}
-	setClockControls(isTrue($sb(prefix+"StartStopButtons)").$sbGet()));
+	setClockControls($sb(prefix+"StartStopButtons)").$sbIsTrue());
 }
 
 // FIXME - this is done after the team/time panel is loaded,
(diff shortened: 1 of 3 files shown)


html/controls/operator.js
```diff
@@ -147,23 +166,22 @@ function createMetaControlTable() {
...
+	$sb("ScoreBoard.FrontendSettings.Setting(ScoreBoard.Operator.StartStopButtons)").$sbControl(showStartStopButton, { sbcontrol: {
+			button: true
+		}, sbelement: {
+			convert: function(value) {
+				showStartStopButton.filter("input:checkbox")
+				.button("option", "label", (isTrue(value)?"Start/Stop Buttons shown":"Start/Stop Buttons hidden"));
+				$("#TeamTime").find("tr.Control").toggleClass("Show", isTrue(value));
+				var operator = $("#operatorId").text();
+				if (operator) {
+					$sb("ScoreBoard.FrontendSettings.Setting(ScoreBoard.Operator__"+operator+".StartStopButtons)").$sbSet(value);
```
brian-brazil: I'd expect the operator to be in parens here

html/controls/operator.js
```diff
@@ -38,13 +38,30 @@ $sb(function() {
 	$("<button>").text("Logout").click(logout).button().css("float", "right").appendTo("#tabBar");
 });
 
+function setOperatorSettings(op) {
+	var operator;
+	if (op !== "") {
+		operator = "__"+ op + ".";
+	} else {
+		operator = "_Default.";
```
brian-brazil: The default is `default`

brian-brazil:
What happens if two people are logged in at once? I think we might want to make the current operator a persistent setting.

frank-weinberg:
Two operators logged in at once (on different clients) would overwrite each other's settings. (Didn't consider that possibility.)

Making the operator persistent would also have undesirable side effects: If you open a second operator screen and login with a different operator there, the settings and key bindings of the first screen would suddenly change.

So the only proper solutions would be to only allow a single instance of the operator panel to be open or to keep these settings localized in the frontend. The latter is way easier to implement and consistent with current behhaviour.

brian-brazil:
> The latter is way easier to implement and consistent with current behhaviour.

That should work, other than for autostart as it affects the backend.

html/controls/operator.js
frank-weinberg: I tried that. It caused the setting name to be nulled somewhere in the XML handling, leading to null pointer exceptions all over the place.

html/controls/operator.js
frank-weinberg: This is intentionally different from the setting of the default operator (note also the single vs. double underscore), as it should be a read only value, so new operators will get the default value and not whatever the last person used that didn't log in.

frank-weinberg:
So I guess this will settle our discussion where to put AutoStart in favor of the ruleset, simply out of technical necessity.

html/controls/operator.js
brian-brazil: Can you add a comment to that effect?

brian-brazil:
Yeah, if you don't have a persistent single login then it has to go in either FrontendSettings or Ruleset.

html/controls/operator.js
frank-weinberg: I had the same thought as I wrote the explanation. It's in there now.

frank-weinberg:
What would still cause interference is when two operator interfaces are open and set to the same operator. But that is probably unavoidable if you want to store the settings in the backend.

html/controls/operator.js
```diff
@@ -122,6 +138,11 @@ function createScoreTimeContent(table) {
 	initialLogin();
 }
 
+function setClockControls(value) {
+	$("#ShowClockControlsButton").checked = value;
```
brian-brazil: This isn't changing the visual display of the button.

frank-weinberg: For me it changes the text color to red (which is the same behaviour as current HEAD).

brian-brazil: Try setting it, and then reloading the page - the start/stop buttons will be there but this won't be red.

frank-weinberg: This should be working now.

brian-brazil: Something's still not right. It's activated now on reload, but it's taking an extra click to toggle it the first time.

frank-weinberg: Finally. (And while the majority of that 1 week delay is attributed to an ugly cold, getting that button to work still took way too much time.)

brian-brazil:
There's something weird going on when you duplicate a tab that's a new behaviour, but it's not due to this change.

PR #168, opened 2018-09-24
Use git tag as version, and include build info.

### LINKED ISSUE #120: Include build info in zips
We've been throwing around developer builds the last few months, which makes it a little bit tricky to determine who was running what as there's only a build timestamp included in the binary. I propose we dump all git state (branch, is it clean, last commit id etc.) plus anything else that might be useful somewhere in the zip file so we can distinguish builds.

brian-brazil:
This avoids having to remember to update a file on every release, and
when ad-hoc builds are made not at a tag they'll include the most recent
tag, how many commits since then, and the commit hash.

Also always include the build username, host, and time so we
can figure out where a build came from if needed.

Fixes #120

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -9,11 +9,8 @@
 	<property name="test.src.dir" value="tests" />
 	<property name="test.build.dir" value="test-bin" />
 	<property name="jdoc.dir" value="jdoc"/>
-	<property name="version.src.dir" value="${src.dir}/com/carolinarollergirls/scoreboard/version"/>
 	<property name="version.dest.dir" value="${dest.dir}/com/carolinarollergirls/scoreboard/version"/>
 	<property name="version.release.file" value="release.properties"/>
-	<property name="version.build.file" value="build.properties"/>
-	<property file="${version.src.dir}/release.properties" prefix="version"/>
 
 	<property name="compile.deprecation" value="true"/>
 	<property name="compile.debug" value="true"/>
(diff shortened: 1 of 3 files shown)


build.xml
```diff
@@ -239,6 +237,9 @@
 				<include name="config/penalties/**" />
 				<include name="config/default/**" />
 			</zipfileset>
+			<zipfileset dir="${version.dest.dir}" prefix="${zip.prefix}/html/" filemode="755" >
```
frank-weinberg, line 240: Why do you put the release file in the html dir?

build.xml
```diff
@@ -59,31 +56,32 @@
...
+			<isfalse value="${isRelease}"/>
+		</condition>
 		<mkdir dir="${version.dest.dir}"/>
-		<propertyfile file="${version.dest.dir}/${version.build.file}">
-			<entry key="build" value="${version.build}"/>
+		<propertyfile file="${version.dest.dir}/${version.release.file}">
+			<entry key="release" value="${version}"/>
+			<entry key="release.user" value="${user.name}"/>
+			<entry key="release.time" value="${version.time}"/>
+			<entry key="release.host" value="${host.NAME}"/>
```
frank-weinberg, line 81: On my machine this gives `localhost` despite a hostname being set.

build.xml
brian-brazil: Otherwise it's buried in a jar inside a jar, this way anyone can be told how to get it via http.

build.xml
brian-brazil: The logic is per https://ant.apache.org/manual/Tasks/hostinfo.html. It's something, the username will likely be more useful.

PR #175, opened 2018-10-03
Remove almost all the log noise at startup.

brian-brazil:
This is not relevant even to developers, and likely
confuses users.


frank-weinberg:
Can we remove the "Copied auto save file[...]"-messages as well?

brian-brazil:
Done

PR #176, opened 2018-10-05
Extend ruleset to support more game types

### LINKED ISSUE #111: Make Foul-Out Penalty Count a Ruleset item
(breaking this out from bullseye555's comment in #66 as it's a bit more of an undertaking than the other style tweaks)

>Can I add one to this - an option in Ruleset (or policies or config, whichever path it goes down) to be able to set # penalties in a game. Then when you're running a 20 min half game with reduced number of penalties, the PT and IWB screens will have their colour highlights based on that (eg: set 5 penalties instead of 7, and it will yellow at 3, orange at 4 and red at 5 etc)(can also open another ticket, if deemed too far from the original to be included in it) 

I think this is a good use of a Ruleset item, and we'll need to update the warning thresholds to be keyed off of this value as well, instead of just based on 7.  I don't think we should allow more than 9 penalties for display reasons, however (Do Rules have a concept of maximum possible value, outside of how Clock objects work?)

### LINKED ISSUE #170: Feature Requests -- RDCL
Adjustments to make the scoreboard fully compatible with [RDCL](http://therdcl.com/) games.

[RDCL rules](http://therdcl.com/wp-content/uploads/2017/12/RDCL-Rules.pdf)

- Allow for intermissions of different lengths. A full RDCL game is played with 4 quarters, with a minimum of 5 minutes between quarters 1&2, and quarters 3&4. (rule 2.4.2) Between quarters 2&3, a minimum of 15 minutes is required. (rule 2.2.2)
- Add an option to prevent the period (quarter) clock from expiring between jams. The rules require that if there is time on the period (quarter) clock after a jam ends, another one will be played. (rule 2.4.5) The most likely option would be to pause it at 0:01 if it reaches that during lineups.

Everything else (as far as I am aware) is easily done with the existing CRG scoreboard. I have not looked into the penalty tracker aspect of the scoreboard in relationship to RDCL. That would open up the option of adding additional support for rule 7.5, penalty jams beyond just using "Timeout before period end". (tl;dr: Penalties are served during the jam after they are earned, and for the full length of the jam. In the event that, during the last jam of the game, the team in the lead earns one or more penalties, an additional jam is played. This repeats until the winning team skates a clean jam.)

frank-weinberg:
- Make the number of penalties that lead to foul out configurable (up to 9)
- Add rules to select which types of timeout stop the period clock (if any)
- Add rule that stops period clock if timeouts last longer than a given time
- Support intermissions that are not all of the same length (and display the official score for more than 15 minutes on WFTDA sanctioned rules)
- Allow to disable the automatic ending of jams after 2 minutes
- Allow to enable automatic ending of team timeouts

Also remove rules for clock settings that are not sensible to change
(e.g. minimum times, which should be kept at 0) and rename the remaining
clock rules so they reflect the semantic of the rule (e.g. number of
periods) instead of the internal variable name (e.g.
Clock.Period.maximumNumber).

Under the hood:
- Refer to rules and settings via constants
- Add unit tests for non-default rule settings
- Do not display the common prefix "Rule." in the ruleset editor

closes #111 
closes #170

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/Ruleset.java
+++ b/src/com/carolinarollergirls/scoreboard/Ruleset.java
@@ -74,14 +74,9 @@ public class Ruleset {
 			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_OTO, "Stop the period clock on official timeouts?", false, "True", "False"));
 			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_TTO, "Stop the period clock on team timeouts?", false, "True", "False"));
 			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_OR, "Stop the period clock on official reviews?", false, "True", "False"));
-			newRule(   new TimeRule(false, ScoreBoard.SETTING_STOP_PC_AFTER, "Stop the period clock, if a timeout lasts longer than this time. Set to a high value to disable.", "60:00"));
+			newRule(   new TimeRule(false, ScoreBoard.SETTING_STOP_PC_AFTER_TO_DURATION, "Stop the period clock, if a timeout lasts longer than this time. Set to a high value to disable.", "60:00"));
 			
-			newRule( new StringRule(false, ScoreBoard.SETTING_INTERMISSION_TYPE_SEQUENCE, "List of the types of intermissions as they appear in the game, separated by commas.", "1,2"));
-			newRule(   new TimeRule(false, ScoreBoard.SETTING_INTERMISSION_DURATION + "1", "Duration of the first type of intermission", "15:00"));
-			newRule(   new TimeRule(false, ScoreBoard.SETTING_INTERMISSION_DURATION + "2", "Duration of the second type of intermission", "60:00"));
-			newRule(   new TimeRule(false, ScoreBoard.SETTING_INTERMISSION_DURATION + "3", "Duration of the third type of intermission", "5:00"));
-			newRule(   new TimeRule(false, ScoreBoard.SETTING_INTERMISSION_DURATION + "4", "Duration of the fourth type of intermission", "15:00"));
-			newRule(   new TimeRule(false, ScoreBoard.SETTING_INTERMISSION_DURATION + "5", "Duration of the fifth type of intermission", "15:00"));
+			newRule( new StringRule(false, ScoreBoard.SETTING_INTERMISSION_DURATIONS, "List of the duration of intermissions as they appear in the game, separated by commas.", "15:00,60:00"));
 			newRule(new BooleanRule(false, ScoreBoard.SETTING_INTERMISSION_DIRECTION, "Which way should the intermission clock count?", true, "Count Down", "Count Up"));
 			
 			newRule(new BooleanRule(false, ScoreBoard.SETTING_AUTO_START,   "Start a Jam or Timeout when the Linup time is over its maximum by BufferTime start a Jam or Timeout as defined below. Jam/Timeout/Period Clocks will be adjusted by the buffer time. This only works if the lineup clock is counting up.", false, "Enabled", "Disabled"));
--- a/src/com/carolinarollergirls/scoreboard/Ruleset.java
+++ b/src/com/carolinarollergirls/scoreboard/Ruleset.java
@@ -16,12 +16,15 @@ import org.json.JSONException;
 import org.json.JSONObject;
 import org.json.JSONTokener;
 
+import com.carolinarollergirls.scoreboard.penalties.PenaltyCodesManager;
 import com.carolinarollergirls.scoreboard.rules.BooleanRule;
 import com.carolinarollergirls.scoreboard.rules.IntegerRule;
 import com.carolinarollergirls.scoreboard.rules.Rule;
 import com.carolinarollergirls.scoreboard.rules.StringRule;
 import com.carolinarollergirls.scoreboard.rules.TimeRule;
-import com.carolinarollergirls.scoreboard.view.Clock;
+import com.carolinarollergirls.scoreboard.view.ScoreBoard;
+import com.carolinarollergirls.scoreboard.view.Skater.Penalty;
+import com.carolinarollergirls.scoreboard.view.Team;
 
 public class Ruleset {
 	public interface RulesetReceiver {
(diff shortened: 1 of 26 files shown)


src/com/carolinarollergirls/scoreboard/Ruleset.java
```diff
@@ -52,58 +55,48 @@ private static Ruleset initialize() {
...
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_TIMEOUT_DIRECTION, "Which way should the timeout clock count?", false, "Count Down", "Count Up"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_TO, "Stop the period clock on every timeout? If false, the options below control the behaviour per type of timeout.", true, "True", "False"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_OTO, "Stop the period clock on official timeouts?", false, "True", "False"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_TTO, "Stop the period clock on team timeouts?", false, "True", "False"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_OR, "Stop the period clock on official reviews?", false, "True", "False"));
+			newRule(   new TimeRule(false, ScoreBoard.SETTING_STOP_PC_AFTER, "Stop the period clock, if a timeout lasts longer than this time. Set to a high value to disable.", "60:00"));
+			
+			newRule( new StringRule(false, ScoreBoard.SETTING_INTERMISSION_TYPE_SEQUENCE, "List of the types of intermissions as they appear in the game, separated by commas.", "1,2"));
```
brian-brazil: This seems over complicated, why not accept `5:00,15:00,5:00`?

src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
```diff
@@ -384,8 +377,8 @@ protected boolean isMasterClock() {
 	public static final int DEFAULT_MINIMUM_NUMBER = 1;
 	public static final int DEFAULT_MAXIMUM_NUMBER = 999;
 	public static final long DEFAULT_MINIMUM_TIME = 0;
-	public static final long DEFAULT_MAXIMUM_TIME = 3600000;
-	public static final boolean DEFAULT_DIRECTION = false;
+	public static final long DEFAULT_MAXIMUM_TIME = 3600000; // 60 minutes
```
brian-brazil: While we're here, can we increase this? I sometimes have time to derby of more than an hour.

src/com/carolinarollergirls/scoreboard/Ruleset.java
```diff
@@ -52,58 +55,48 @@ private static Ruleset initialize() {
...
+			newRule(   new TimeRule(false, ScoreBoard.SETTING_JAM_DURATION, "Maximum duration of a jam", "2:00"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_JAM_DIRECTION, "Which way should the jam clock count?", true, "Count Down", "Count Up"));
+
+			newRule(   new TimeRule(false, ScoreBoard.SETTING_LINEUP_DURATION, "Duration of lineup", "0:30"));
+			newRule(   new TimeRule(false, ScoreBoard.SETTING_OVERTIME_LINEUP_DURATION, "Duration of lineup before an overtime jam", "1:00"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_LINEUP_DIRECTION, "Which way should the lineup clock count?", false, "Count Down", "Count Up"));
+			
+			newRule(   new TimeRule(false, ScoreBoard.SETTING_TTO_DURATION, "Duration of a team timeout", "1:00"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_TIMEOUT_DIRECTION, "Which way should the timeout clock count?", false, "Count Down", "Count Up"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_TO, "Stop the period clock on every timeout? If false, the options below control the behaviour per type of timeout.", true, "True", "False"));
```
brian-brazil, line 73: Do we really need 4 settings for this? I'd say either have one overall, or the 3 individual.

What happens when a timeout changes from one type to another?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -231,6 +265,31 @@ public void setTimeoutType(String owner, boolean review) {
 			}
 			setTimeoutOwner(owner);
 			setOfficialReview(review);
+			if (!settings.getBoolean(SETTING_STOP_PC_ON_TO)) {
+				boolean stopPc = false;
+				if (!owner.equals(TIMEOUT_OWNER_NONE)) {
```
brian-brazil, line 266: This doesn't handle when the timeout type is unknown, the clock won't be stopped in that case.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -617,11 +690,25 @@ public void scoreBoardChange(ScoreBoardEvent event) {
...
-			if (settings.getBoolean("ScoreBoard." + Clock.ID_LINEUP + ".AutoStart")) {
+			if (settings.getBoolean(SETTING_AUTO_START)) {
 				_possiblyAutostart();
 			}
 		}
 	};
+	protected ScoreBoardListener timeoutClockListener = new ScoreBoardListener() {
+		public void scoreBoardChange(ScoreBoardEvent event) {
+			if (settings.getBoolean(SETTING_AUTO_END_TTO) &&
+					!getTimeoutOwner().equals(TIMEOUT_OWNER_NONE) &&
```
brian-brazil, line 693: Why not explicitly check if it's team 1 or team 2?

src/com/carolinarollergirls/scoreboard/Ruleset.java
```diff
@@ -52,58 +55,48 @@ private static Ruleset initialize() {
...
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_OTO, "Stop the period clock on official timeouts?", false, "True", "False"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_TTO, "Stop the period clock on team timeouts?", false, "True", "False"));
+			newRule(new BooleanRule(false, ScoreBoard.SETTING_STOP_PC_ON_OR, "Stop the period clock on official reviews?", false, "True", "False"));
+			newRule(   new TimeRule(false, ScoreBoard.SETTING_STOP_PC_AFTER, "Stop the period clock, if a timeout lasts longer than this time. Set to a high value to disable.", "60:00"));
```
brian-brazil: STOP_PC_AFTER_TIMEOUT_DURATION would be a bit clearer

src/com/carolinarollergirls/scoreboard/Ruleset.java
frank-weinberg: Mostly because we don't have a rule type that covers parsing this.

src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
frank-weinberg: If you set your time to derby via "Start game" the maximum will be increased accordingly.

src/com/carolinarollergirls/scoreboard/Ruleset.java
brian-brazil: You could use String and then do the clock conversions yourself. The rule types aren't exactly doing anything fancy.

src/com/carolinarollergirls/scoreboard/defaults/DefaultClockModel.java
brian-brazil: I often set the intermission clock directly, such as if I've already started the new game and some roster changes come in - and then there's a schedule change

src/com/carolinarollergirls/scoreboard/Ruleset.java
frank-weinberg: Only one setting is less flexible. Having 3 individual settings leaves the question of what to do with timeouts for which no type is specified (yet or the whole time).

>What happens when a timeout changes from one type to another?

The clock is rolled back/forward as if the timeout had always been marked as that type.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: If "stop on all timeouts" is set, the clock will already be stopped in '_startTimeout()'.
Otherwise the clock intentionally doesn't stop in untyped timeouts.

src/com/carolinarollergirls/scoreboard/Ruleset.java
brian-brazil: Has anyone asked for that flexibility?

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: This way it's all &&, checking for teams would add an ||.

src/com/carolinarollergirls/scoreboard/Ruleset.java
frank-weinberg: I seem to remeber games that stopped the clock on TTO but not OTO or vice versa.

frank-weinberg:
updated

PR #178, opened 2018-10-06
Add option to replace instead of undo

frank-weinberg:
Useful e.g. if what the operator thought was a jam starting whistle
turns out to be a timeout instead.

### DIFF
--- a/html/controls/operator.js
+++ b/html/controls/operator.js
@@ -49,6 +49,7 @@ function setOperatorSettings(op) {
 		prefix = "ScoreBoard.FrontendSettings.Setting(ScoreBoard.Operator_Default.";
 	}
 	setClockControls($sb(prefix+"StartStopButtons)").$sbIsTrue());
+	setReplaceButton($sb(prefix+"ReplaceButton)").$sbIsTrue());
 }
 
 // FIXME - this is done after the team/time panel is loaded,
@@ -144,6 +145,13 @@ function setClockControls(value) {
 	$("#TeamTime").find("tr.Control").toggleClass("Show", value);
 }
 
+function setReplaceButton(value) {
+	$("#EnableReplaceButton").prop("checked", value);
+	$("label.EnableReplaceButton").toggleClass("ui-state-active", value);
+	$("#ClockUndo").toggleClass("Hidden KeyInactive", value);
+	$("#ClockReplace").toggleClass("Hidden KeyInactive", !value);
+}
+
 function createMetaControlTable() {
 	var table = $("<table><tr><td/></tr><tr><td/></tr><tr><td/></tr></table>")
 		.addClass("MetaControl");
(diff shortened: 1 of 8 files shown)


html/controls/operator.js
```diff
@@ -144,6 +145,13 @@ function setClockControls(value) {
 	$("#TeamTime").find("tr.Control").toggleClass("Show", value);
 }
 
+function setReplaceButton(value) {
+	$("#EnableReplaceButton").prop("checked", value);
+	$("label.EnableReplaceButton").toggleClass("ui-state-active", value);
+	$("#ClockUndo").toggleClass("Hidden KeyInactive", value);
```
brian-brazil, line 151: This isn't working, I'm always seeing both buttons.

I'd suggest always showing both.

html/controls/operator.js
```diff
@@ -433,8 +455,15 @@ function createOvertimeDialog() {
 
 function createJamControlTable() {
 	var table = $("<table><tr><td/></tr></table>").addClass("JamControl");
+	var replaceInfoTr = createRowTable(1).addClass("ReplaceInfo Hidden").appendTo(table.find("td"));
 	var controlsTr = createRowTable(4,1).appendTo(table.find("td")).find("tr:eq(0)").addClass("Controls");
 
+	var replaceInfoText = $("<span>").html("Replace &quot;<span id=\"replacedLabel\"></span>&quot; with").appendTo(replaceInfoTr);
```
brian-brazil, line 468: This text is always showing.

brian-brazil (commented):
Frontend doesn't seem right here.

Just thinking there - should we log operator clock-related actions to the js console? That'd help debug if someone presses the wrong thing

brian-brazil:
Ignore that, a 2nd forced refresh got it working.

I'd still suggest always showing both the replace and undo buttons - in a hurry you don't want to have to do the extra click to turn it on

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -189,14 +198,17 @@ public void stopJamTO() {
 				createSnapshot(ACTION_STOP_JAM);
 				setLabels(ACTION_START_JAM, ACTION_NONE, ACTION_TIMEOUT);
 				_endJam(false);
+				finishReplace();
```
brian-brazil, line 230: We should probably have unittests covering each of these new finishReplace calls

brian-brazil:
I've got a better feel for how this works now, I'm wondering if we should just hardcode this as the only way.

frank-weinberg:
The idea is that the operator chooses one mode of operation and uses that all the time. (In replace mode clicking "Undo" and then replace with "no action" has the same effect as straight undo, so there's no need for a straight undo button. (And we are striving to reduce the number of buttons, not increase it.)

frank-weinberg:
And the reason I've not hardcoded it is that there is a good chance operators would discover it the first time they press "undo" on a newer version of the scoreboard. And many operators won't be open to learn about new functionality just as they press undo. I hope we can do it that way once the new mode has become known amongst operators.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: I've extended the unittest that was already there instead of adding new ones.

PR #180, opened 2018-10-06
Fix button labels when a jam auto ends after 2 min

frank-weinberg:


brian-brazil:
Do we have unittests for these?

frank-weinberg:
No. Could probably add that to the existing tests.

frank-weinberg:
Added unit tests (and can't reproduce the crash from CircleCI)

brian-brazil:
We should probably figure out what that is, try adding `showoutput="yes"` to the junit task in build.xml.

frank-weinberg:
That didn't really help. But it's a different testcase now.

brian-brazil:
Looks like we're running out of RAM, possibly as we've two JVMs running given the forked unittests. Add maxmemory="128m" to the JVM line.

frank-weinberg:
Should we leave the showoutput flag?

brian-brazil:
Can't hurt.

frank-weinberg:
done

Issue #173, opened 2018-10-01
Official Reviews not shown in SB Display


AlpenYeti:
On Official Reviews, the dot does not appear to be impacted. It does not blink while the OR is being stated, and does not go off if OR has been lost.

brian-brazil:
Thanks for reporting, which screen is this exactly?

AlpenYeti:
I report the OR on the Operator Control Pannel, where the counter updates correctly, but the dot does not reflect the state of the OR on the "Main Scoreboard" display pannel.

AdamSmasherDerby:
In addition, it appears that background colors are not being set for the OR dots from the "scoreboard_dots" color.

AdamSmasherDerby:
I don't have time to fix this weekend (tournment this weekend) but it looks like the color problem is the same as referenced in pull #121 - the divs for the official review include pseudoclasses, (:before and :after) which can't have the colors set directly the way normal classes are.

JeneralPain:
I can confirm in the 3.9.5 test build I have done, this issue is now resolved. When we close #183 completely then this issue will be closed.

JeneralPain:
This has now been fixed thanks to #183 

When the next release comes out, you'll find this bug goes away. :)

PR #183, opened 2018-10-13
Fixed non-blinking official review dots. Resolves issue #173

AdamSmasherDerby:
Turns out the problem fortunately didn't involve mucking with pseudoclasses.  Rather, it appears the code which made a WS query to determine which dot to blink for team timeouts was cut-n-pasted for official reviews.  Since there is only ever one official review dot, that did nae work so well.  Fixed.

In addition, I tidied up the gradient for the dots, and commented out an unused line in the wb file that was throwing errors into the console.  If you really, really want me to put those in separate PRs I will.

### DIFF
--- a/html/views/wb/tcdg2016wb.js
+++ b/html/views/wb/tcdg2016wb.js
@@ -101,7 +101,7 @@ function displayPenalty(t, s, p) {
 	var limit = WS.state['ScoreBoard.Setting(Rule.Penalties.NumberToFoulout)'];
 	var fo_exp = ($($('.Team' + t + ' .Skater.Penalty[id=' + s + '] .BoxFO_EXP')[0]).data("id") != null);
 	$('.Team' + t + ' .Skater.Penalty[id=' + s + '] .Box').each(function(idx, elem) { cnt += ($(elem).data("id") != null ? 1 : 0); });
-	totalBox.text(cnt);
+	//totalBox.text(cnt);
 	$('.Team' + t + ' .Skater[id=' + s + ']').toggleClass("Warn1", cnt == limit-2 && !fo_exp);
 	$('.Team' + t + ' .Skater[id=' + s + ']').toggleClass("Warn2", cnt == limit-1 && !fo_exp);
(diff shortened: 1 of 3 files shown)


JeneralPain:
Unless there is an objection, I am going to MERGE and CLOSE this pull?

frank-weinberg:
We are currently running a policy that each PR is reviewed by at least one other developer before being merged. I don't see a reason to divert from this policy here.

JeneralPain:
That's fine, I'm just trying to keep things moving as best I can.

html/views/wb/tcdg2016wb.js
```diff
@@ -101,7 +101,7 @@ function displayPenalty(t, s, p) {
 	var limit = WS.state['ScoreBoard.Setting(Rule.Penalties.NumberToFoulout)'];
 	var fo_exp = ($($('.Team' + t + ' .Skater.Penalty[id=' + s + '] .BoxFO_EXP')[0]).data("id") != null);
 	$('.Team' + t + ' .Skater.Penalty[id=' + s + '] .Box').each(function(idx, elem) { cnt += ($(elem).data("id") != null ? 1 : 0); });
-	totalBox.text(cnt);
+	//totalBox.text(cnt);
```
brian-brazil: Delete rather than commenting out.

brian-brazil (commented):
The retained doesn't look quite right, there's a noticable black line on the right and bottom plus shading behind it.

AdamSmasherDerby:
Line from WB deleted, and since I can't reproduce the issue Brian is seeing, I've reverted the gradient to its previous setting.

html/views/wb/tcdg2016wb.js
AdamSmasherDerby: Done

brian-brazil:
Still seeing it when it's blinking: http://i.imgur.com/ta5ID2H.png

I'm on Chrome.

AdamSmasherDerby:
Then that's not related to the PR, because I literally didn't touch anything else.

Also, looking at the image you posted, I don't see any gradient at all, which is odd.

PR #190, opened 2018-10-19
 Whiteboard Team colors same as Penalty tracker

riffreff:
This would make the whiteboard headers coloured and pull those colours from the operator colours. 
Congruent to the PT screen

(first commit seems already merged, but still shows up here)

### DIFF
--- a/html/views/wb/tcdg2016wb.css
+++ b/html/views/wb/tcdg2016wb.css
@@ -284,7 +284,7 @@ a {
 */
 .Team .Name .Number, .Team .Box, .Team .FO_EXP, .Team .Total { width: 5%; }
 .Number { left: 0%; width: 10%; text-align: right; }
-.Name { left: 12%; width: 44%; height: 100%; text-align: left; overflow: hidden; }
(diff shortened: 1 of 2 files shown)


frank-weinberg:
Whiteboard is public facing, using operator color is counterintuitive here.

We should discuss the handling of colors in general: The current setup of having n pairs of colors defined for m > n screens is bound to always cause difficulties. Let's either have a separate color for each screen or (better IMO) define one pair of colors and a set of binary options that define which screens use colors at all.

brian-brazil:
Yeah, this whole area needs cleanup. At the least we'd want one operator and one public colourset.

riffreff:
At the moment there is NO colour indication on the whiteboard. Making it the same as the Penalty Tracker is very intuitive. 

This was meant to be a quick resolution to user requests. Which could simply be implemented and a "bigger" solution could come later

riffreff:
This was meant to be a quick resolution to user requests. Whcih could simply be omplemented and a "bigger" solution could come later

PR #192, opened 2018-10-30
Change rulesets to use normal XML/WS way of working.

brian-brazil:
This is also shorter code wise, even with the new unittests. This goes back to Settings for everything - but now with a simpler implementation. Rules are accessed directly as needed.

I didn't implement an API for changing the active ruleset, but it should be easy to build on top of what's here.


src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
```diff
@@ -0,0 +1,339 @@
...
+/**
+ * Copyright (C) 2008-2012 Mr Temper <MrTemper@CarolinaRollergirls.com>
+ *
+ * This file is part of the Carolina Rollergirls (CRG) ScoreBoard.
+ * The CRG ScoreBoard is licensed under either the GNU General Public
+ * License version 3 (or later), or the Apache License 2.0, at your option.
+ * See the file COPYING for details.
+ */
+
+import java.util.ArrayList;
```
frank-weinberg:
```suggestion
```
unused

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
```diff
@@ -0,0 +1,339 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
```
frank-weinberg:
```suggestion
```
unused

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
```diff
@@ -0,0 +1,339 @@
...
+            }
+
+            rulesets.remove(id);
+            r.removeScoreBoardListener(this);
+            scoreBoardChange(new ScoreBoardEvent(this, EVENT_REMOVE_RULESET, r, ""));
+            requestBatchEnd();
+        }
+    }
+
+    private ScoreBoardModel sbm = null;
```
frank-weinberg:
```suggestion
```
unused

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
```diff
@@ -29,7 +28,7 @@
 import com.carolinarollergirls.scoreboard.utils.ClockConversion;
 import com.carolinarollergirls.scoreboard.utils.ScoreBoardClock;
 import com.carolinarollergirls.scoreboard.view.Clock;
-import com.carolinarollergirls.scoreboard.view.FrontendSettings;
+import com.carolinarollergirls.scoreboard.view.Rulesets;
 import com.carolinarollergirls.scoreboard.view.ScoreBoard;
 import com.carolinarollergirls.scoreboard.view.Settings;
 import com.carolinarollergirls.scoreboard.view.Skater.Penalty;
```
frank-weinberg:
```suggestion
```
unused (as is PenaltyCodesManager)

src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
```diff
@@ -283,6 +280,30 @@ private void processSettings(String path, Settings s) {
         }
```
frank-weinberg: processSettings is unused

src/com/carolinarollergirls/scoreboard/rules/BooleanRule.java
```diff
@@ -4,46 +4,35 @@
 import org.json.JSONObject;
```
frank-weinberg: Both imports are unused

src/com/carolinarollergirls/scoreboard/rules/Rule.java
```diff
@@ -4,11 +4,10 @@
 import org.json.JSONObject;
```
frank-weinberg: Both imports are unused

src/com/carolinarollergirls/scoreboard/view/ScoreBoard.java
```diff
@@ -60,18 +59,15 @@
...
+    public Rulesets getRulesets();
 
     public Stats getStats();
 
     public XmlScoreBoard getXmlScoreBoard();
 
-    public static final String FRONTEND_SETTING_CLOCK_AFTER_TIMEOUT = "ScoreBoard.ClockAfterTimeout";
+    public static final String SETTING_CLOCK_AFTER_TIMEOUT = "ScoreBoard.ClockAfterTimeout";
 
     public static final String SETTING_NUMBER_PERIODS = "Rule." + Clock.ID_PERIOD + ".Number";
```
frank-weinberg: Since the values are accessed through the Ruleset Object now and in order to seperate them from the previous FRONTEND_SETTINGs, the constants' prefix should be changed to RULE_.
Also as they are no longer mixed with other settings, the prefix "Rule." in the String has become superfluous.

src/com/carolinarollergirls/scoreboard/xml/ScoreBoardXmlListener.java
```diff
@@ -8,6 +8,9 @@
  * See the file COPYING for details.
  */
 
+
+import java.util.Set;
+
 import org.jdom.Document;
 import org.jdom.Element;
```
frank-weinberg, line 16: ScoreBoardManager is an unused import

tests/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModelTests.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import com.carolinarollergirls.scoreboard.ScoreBoardManager;
+import com.carolinarollergirls.scoreboard.jetty.JettyServletScoreBoardController;
+import com.carolinarollergirls.scoreboard.model.ScoreBoardModel;
+import com.carolinarollergirls.scoreboard.model.RulesetsModel;
+import com.carolinarollergirls.scoreboard.model.RulesetsModel.RulesetModel;
+import com.carolinarollergirls.scoreboard.view.ScoreBoard;
+import com.carolinarollergirls.scoreboard.view.Rulesets;
```
frank-weinberg:
```suggestion
```
unused

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
```diff
@@ -0,0 +1,339 @@
...
+            return Long.parseLong(get(k));
+        }
+    }
+    public void set(String k, String v) {
+        synchronized (coreLock) {
+            Rule r = rules.get(k);
+            if (r == null || !r.isValueValid(v)) {
+                return;
+            }
+            current.put(k, v);
```
frank-weinberg, line 177: When doing this there should be some kind of indication in `name` and possibly `id` that the ruleset has been modified. (Maybe add a pre- or suffix "(modified)")

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
```diff
@@ -0,0 +1,339 @@
...
+            if (id.equals(rootId)) {
+                return;
+            }
+            Ruleset r = rulesets.get(id);
+            if (r == null) {
+                return;
+            }
+            requestBatchStart();
+            // Point any rulesets with the deleted one as their parent
+            // to their grandparent.
```
frank-weinberg, line 228: This should also copy any non-overwritten settings to the child, so the child isn't modified.

tests/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModelTests.java
```diff
@@ -0,0 +1,89 @@
...
+        Map<String,String> s = new HashMap<String,String>();
+        s.put(ScoreBoard.SETTING_PERIOD_DURATION, "1:00");
+        child.setAll(s);
+        rm.setCurrentRuleset(id1);
+        assertEquals(60000, rm.getLong(ScoreBoard.SETTING_PERIOD_DURATION));
+    }
+
+    @Test
+    public void testReparentingToGrandparent() {
+        RulesetModel child = rm.addRuleset("child", rootId, id1);
```
frank-weinberg:
```suggestion
        rm.addRuleset("child", rootId, id1);
```
`child` is unused

frank-weinberg (changes requested):

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
brian-brazil: They're both used for constants for rule names.

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
brian-brazil: This is used by the XML loader, so that'd be unwise. As for the user's ability to change things, the user has no way to do that currently so I was leaving that all up to your planned changes in this area.

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
brian-brazil: Rulesets only contain their overrides, the resolution is done in the JS. There's no need for this here.

src/com/carolinarollergirls/scoreboard/view/ScoreBoard.java
brian-brazil: > Since the values are accessed through the Ruleset Object now and in order to seperate them from the previous FRONTEND_SETTINGs, the constants' prefix should be changed to RULE_.

Done.

> Also as they are no longer mixed with other settings, the prefix "Rule." in the String has become superfluous.

I was leaving that to a future cleanup.

brian-brazil:
I believe I've addressed all of your comments.

src/com/carolinarollergirls/scoreboard/defaults/DefaultScoreBoardModel.java
frank-weinberg: Those rules are not referenced in DefaultScoreBoardModel.

brian-brazil: Ah, I was looking at the wrong file. Fixed.

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
frank-weinberg: At the very least, there should be a flag that indicates that the ruleset has been changed. Even if the frontend doesn't use the functionality yet, it's never a good idea to have such unwritten requirements for future changes.

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
frank-weinberg: If there is a setting that is overridden in the deleted ruleset but not overwritten again in the reparented child, this setting will change for the child. I would consider this unexpected behaviour.

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
brian-brazil: There is no functionality for the frontend to use, this code is only called when loading the save.

src/com/carolinarollergirls/scoreboard/defaults/DefaultRulesetsModel.java
brian-brazil: That's all handled in the UI code. The rules are only fully rendered in the JS, the WS only contains the overrides that a given ruleset is making.

brian-brazil: I think understand what you mean now, let me take a look at that.

frank-weinberg: The UI doesn't do that either. Pressing the "Delete" button calls:
`WS.Command("RemoveRuleset", o);`
And the WS command only calls this function.
So overrides that are set in the deleted ruleset are not copied over to its children, changing them unless they happen to reoverride the rule.

brian-brazil: That's done.

tests/com/carolinarollergirls/scoreboard/defaults/DefaultSettingsModelTests.java
```diff
@@ -8,54 +8,23 @@
 import org.junit.Test;
```
frank-weinberg, line 6: AssertEquals and AssertNull are now unused

src/com/carolinarollergirls/scoreboard/xml/ScoreBoardXmlListener.java
```diff
@@ -284,10 +301,7 @@ protected Element getScoreBoardElement() {
     }
 
     private Element getSettingsElement(SettingsModel settings) {
```
frank-weinberg: This function is unused

frank-weinberg (commented):
There's three more "unused" items that I missed yesterday. Sorry for the extra loop.

brian-brazil:
Updated.

frank-weinberg (approved):

PR #207, opened 2018-11-25
Move core into single package

frank-weinberg:
In the process simplfy structure from Interfaces view.X, model.XModel
and implementation DefaultXModel to single Interface core.X and
implementation core.implementation.XImpl


src/com/carolinarollergirls/scoreboard/core/ScoreBoard.java
```diff
@@ -14,12 +14,16 @@
 import com.carolinarollergirls.scoreboard.xml.XmlScoreBoard;
 
 public interface ScoreBoard extends ScoreBoardEventProvider {
+    /** Reset the entire ScoreBoard. */
```
brian-brazil: Remove "entire", as we don't reset Settings anymore.

src/com/carolinarollergirls/scoreboard/core/implementation/ClockImpl.java
```diff
@@ -1,4 +1,4 @@
-package com.carolinarollergirls.scoreboard.defaults;
+package com.carolinarollergirls.scoreboard.core.implementation;
```
brian-brazil: I think just impl would be enough, and it's faster to type.

src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
```diff
@@ -29,14 +29,13 @@
...
+import com.carolinarollergirls.scoreboard.core.ScoreBoard;
 import com.carolinarollergirls.scoreboard.json.JSONStateManager;
-import com.carolinarollergirls.scoreboard.model.ScoreBoardModel;
-// import org.eclipse.jetty.util.resource.Resource;
 
 
 public class JettyServletScoreBoardController {
-    public JettyServletScoreBoardController(ScoreBoardModel model, JSONStateManager jsm) {
-        scoreBoardModel = model;
+    public JettyServletScoreBoardController(ScoreBoard model, JSONStateManager jsm) {
```
brian-brazil: We should probably change args like this to `sb`

brian-brazil (commented):
model is still used in names here and there.

src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
frank-weinberg: I've eliminated all occurences of the string 'model' from the java code now.

PR #209, opened 2018-12-01
Caps lock

brian-brazil:


brian-brazil:
This updates state any time a key is pressed (including capslock). That's the most that seems possible with JS.

frank-weinberg:
When pressing Caps Lock itself the display status is set the wrong way. On all other key presses (including Shift or Control) it works as expected.

brian-brazil:
It's working correctly for me. What OS/Browser are you on?

frank-weinberg:
This was with Firefox and IE on Win7. Can try with Firefox and Chrome under Linux later today.

frank-weinberg:
On Linux it works as expected in both browsers. So it's a Windows problem. (And I won't be at my Windows computer for the next two weeks.)

brian-brazil:
Okay, presumably Windows changes when the Caps Lock is activated at a different point in the key press than Linux.

brian-brazil:
I've tweaked this so it should work better across OSes.

frank-weinberg:
This appears to work correctly now on all combinations of Win7 or Linux with Firefox or Chrome.

PR #210, opened 2018-12-21
Turn most sets of string constants into enums

frank-weinberg:
- Rules
- FloorPositions and Roles
- TimeoutOwners that are not Teams
- Buttons

Actions (Button labels) are kept as string constants 

Events will be enumized in a separate PR


src/com/carolinarollergirls/scoreboard/rules/Rule.java
```diff
@@ -1,35 +1,48 @@
...
+    NUMBER_REVIEWS(new IntegerRule("Team.OfficialReviews", "How many official reviews each team is granted per game or period", 1)),
+    REVIEWS_PER_PERIOD(new BooleanRule("Team.OfficialReviewsPer", "Are official reviews granted per period or per game?", true, "Period", "Game")),
+
+    PENALTIES_FILE(new StringRule("Penalties.DefinitionFile", "File that contains the penalty code definitions to be used", "/config/penalties/wftda2018.json")),
+    FO_LIMIT(new IntegerRule("Penalties.NumberToFoulout", "After how many penalties a skater has fouled out of the game. Note that the software currently does not support more than 9 penalties per skater.", 7));
+
+    private Rule(AbstractRule r) { rule = r; }
+    
+    public AbstractRule getRule() { return rule; }
+    public String toString() {return rule.getFullName(); }
```
brian-brazil, line 45: I'd expect toString to return all the information about a rule, not just one field. I think getFullName should be used directly where this is needed.

We could also make it getName now.

src/com/carolinarollergirls/scoreboard/core/FloorPosition.java
```diff
@@ -0,0 +1,37 @@
+package com.carolinarollergirls.scoreboard.core;
+
+public enum FloorPosition {
+    JAMMER(Role.JAMMER, "Jammer"),
+    PIVOT(Role.PIVOT, "Pivot"),
```
brian-brazil: A comment that this could be a 4th blocker would be useful.

src/com/carolinarollergirls/scoreboard/core/PositionNotFoundException.java
```diff
@@ -9,12 +9,12 @@
  */
 
 public class PositionNotFoundException extends RuntimeException {
```
brian-brazil: Can this class be eliminated with these changes?

src/com/carolinarollergirls/scoreboard/rules/Rule.java
frank-weinberg: This is toString of the enum, not the rule itself. Javadoc for `Enum.toString()` says:

> Returns the name of this enum constant, as contained in the declaration. 
 This method may be overridden, though it typically isn't necessary or desirable. An enum type 
 should override this method when a more "programmer-friendly" string form exists.

In this case the more programmer-friendly form is the string we use to identify the rule in JSON and XML.

frank-weinberg:
comments are addressed.

PR #212, opened 2019-01-07
Align structure of Backend objects, JSON and XML paths

frank-weinberg:
User visible changes: none

Changes to the frontend-backend interface:
- WS/JSON interface now supports setting any element under scoreboard that can be set via the XML interface. Not yet implemented: Support for Change/Reset Flags. Previous proprietar commands for Penalties and Rulesets have been removed.
- Paths have been aligned between the two interfaces and with the backend structure. This affected JSON paths for Rulesets, Media, Team.AlternateName, Team.Color, Settings, and PenaltyCodes and XML paths for Rulesets, Media, and PenaltyCodes (the latter being an addition).

Internal Changes in the Backend:
- DefaultScoreBoardEventProvider has been extended to provide infrastructure for managing child elements, values, and commands, that are exposed to the frontend. Functionality specific to a certain value (validity checks, followup actions, etc) can be added through overrides in the *Impl classes.
- By using the aforementioned infrastructure XML and JSON structure are directly derived from the backend structure. The multiple unwieldy if-elseif-chains have been dropped.

Side effect: Migrating frontend pages from XML to JSON has become easier as the paths of the elements are guaranteed to be identical in both interfaces. At the same time the urge for these migrations is lessened as maintaining both of them through changes in the backend structure adds negligible overhead.


src/com/carolinarollergirls/scoreboard/event/ScoreBoardEventProvider.java
```diff
@@ -18,32 +18,85 @@
...
+     *  (Typically a UUID.)
+     */
     public String getProviderId();
+    /**
+     * The parent element in the XML and JSON representation
+     */
     public ScoreBoardEventProvider getParent();
+    /**
+     * This should return all the enums that contain values, children, or commands
+     * that can be accessed from the frintend
```
brian-brazil, line 44: frontend

src/com/carolinarollergirls/scoreboard/core/Media.java
```diff
@@ -23,16 +24,16 @@
...
         public String getType();
         public String getId();
         public String getName();
         public void setName(String s);
         public String getSrc();
-        public static final String EVENT_FILE = "File";
     }
 
+    public enum Child implements MultiProperty {
+	FILE;
```
brian-brazil: Shouldn't there be more properties here? I'd expect Name at least

src/com/carolinarollergirls/scoreboard/event/ScoreBoardEvent.java
```diff
@@ -64,10 +64,11 @@ public String toString() {
...
     protected ScoreBoardEventProvider provider;
-    protected String property;
+    protected Property property;
     protected Object value;
     protected Object previousValue;
 
-    public static final String BATCH_START = "Batch Start";
-    public static final String BATCH_END = "Batch End";
+    public interface Property {}
+    public interface SingleProperty extends Property {}
```
brian-brazil: A comment explaining what these are would be useful

src/com/carolinarollergirls/scoreboard/xml/ScoreBoardXmlConverter.java
```diff
@@ -370,23 +369,32 @@ public void processScoreBoard(ScoreBoard scoreBoard, Element element) {
...
                         scoreBoard.clockUndo(true);
-                    } else if (name.equals("StartOvertime")) {
+			break;
+		    case START_OVERTIME:
                         scoreBoard.startOvertime();
-                    } else if (name.equals("OfficialTimeout")) {
+			break;
+		    case OFFICIAL_TIMEOUT:
                         scoreBoard.setTimeoutType(TimeoutOwners.OTO, false);
+			break;
```
brian-brazil: We should probably have a default here with a log line to catch bugs

src/com/carolinarollergirls/scoreboard/utils/PropertyConversion.java
```diff
@@ -0,0 +1,31 @@
...
+	    {
+	        sb.append( oneString.substring(0,1) );
+	        sb.append( oneString.substring(1).toLowerCase() );
+	    }
+	    return sb.toString();
+	}
+	return prop.toString();
+    }
+    
+    public static Property fromFrontend(String name, List<Class<? extends Property>> types) {
```
brian-brazil, line 21: You only ever pass one type in, so this can be simplified

src/com/carolinarollergirls/scoreboard/json/ScoreBoardJSONListener.java
```diff
@@ -448,14 +452,26 @@ private void initialize(ScoreBoard sb) {
...
         updateState();
     }
+    
+    String getPath(ScoreBoardEventProvider p) {
+	String path = "";
+	if (p.getParent() != null) {
+	    path = getPath(p.getParent()) + ".";
+	}
+	path = path + p.getProviderName();
+	if (!"".equals(p.getProviderId()) && p.getProviderId() != null) {
```
brian-brazil: This is a bit odd.

p.getProviderId() != null && !p.getProviderId().isEmpty()

I think there may also be cases where an empty value is valid

src/com/carolinarollergirls/scoreboard/event/ScoreBoardEvent.java
```diff
@@ -16,8 +16,9 @@ public ScoreBoardEvent(ScoreBoardEventProvider sbeP, PermanentProperty p, Object
         this(sbeP, (Property)p, v, prev);
     }
 
-    public ScoreBoardEvent(ScoreBoardEventProvider sbeP, AddRemoveProperty p, ValueWithId v, ValueWithId prev) {
-        this(sbeP, (Property)p, v, prev);
+    public ScoreBoardEvent(ScoreBoardEventProvider sbeP, AddRemoveProperty p, ValueWithId v, boolean r) {
```
brian-brazil, line 19: The boolean here is a bit magic, could it be clearer from the usage?

src/com/carolinarollergirls/scoreboard/event/DefaultScoreBoardEventProvider.java
```diff
@@ -50,20 +55,64 @@ public void removeScoreBoardListener(ScoreBoardListener listener) {
...
+    public boolean set(PermanentProperty prop, Object value, Flag flag) { return set(prop, value, flag, null, null, 0); }
+    public boolean set(PermanentProperty prop, Object value, Flag flag, Number min, Number max, long tolerance) {
+	synchronized (coreLock) {
+	    Object last = values.get(prop);
+	    if (flag == Flag.CHANGE) {
+		if (last instanceof Integer) {
+		    value = (Integer)last + (Integer)value;
+		} else if (last instanceof Long) {
+		    value = (Long)last + (Long)value;
+		}
```
brian-brazil, line 94: We should log or something here, to catch bugs

tests/com/carolinarollergirls/scoreboard/core/impl/ClockImplTests.java
```diff
@@ -587,20 +586,21 @@ public void testChangeTime() {
...
 
-        clock.changeTime(-201);
+        clock.timerTick(-201);
         assertEquals(999, clock.getTime());
         assertEquals(1, collectedEvents.size());
 
         clock.setCountDirectionDown(true);
         collectedEvents.clear();
-        clock.changeTime(1);
+        clock.timerTick(-1);
```
brian-brazil, line 602: Why is this changing to negative?

src/com/carolinarollergirls/scoreboard/core/Skater.java
```diff
@@ -47,15 +47,15 @@
...
-	PENALTY_FOEXP;
+	FLAGS;
     }
     public enum Child implements AddRemoveProperty {
 	PENALTY;
     }
 
     public static interface Penalty extends ScoreBoardEventProvider {
         public String getId();
+        public String getUuid();
```
brian-brazil, line 56: We seem to usually have UUID

src/com/carolinarollergirls/scoreboard/core/Rulesets.java
```diff
@@ -43,14 +44,13 @@
...
-	ID,
-	NAME,
-	RULE_DEFINITIONS,
-	RULE,
-	RULESET;
+	CURRENT_RULESET_ID,
+	CURRENT_RULESET_NAME;
     }
     public enum Child implements AddRemoveProperty {
-	KNOWN_RULESETS;
```
brian-brazil: One of the reasons I went with a Known prefix is so that you won't pull in all the rulesets if you just care about the current settings. I think that's still the case here.

src/com/carolinarollergirls/scoreboard/core/impl/SkaterImpl.java
```diff
@@ -63,16 +63,20 @@ public boolean set(PermanentProperty prop, Object value, Flag flag) {
 		requestBatchStart();
 		Object last = get(prop);
 		result = super.set(prop, value, flag);
-		if(result && prop == Value.PENALTY_BOX && getPosition() != null) {
-		    if ((Boolean)value && getPosition().getFloorPosition() == FloorPosition.JAMMER
-			    && team.getLeadJammer().equals(Team.LEAD_LEAD)) {
-			team.setLeadJammer(Team.LEAD_LOST_LEAD);
+		if(result) {
```
brian-brazil, line 79: Syntax is off here

src/com/carolinarollergirls/scoreboard/core/Skater.java
```diff
@@ -35,10 +33,9 @@
     public void setPenaltyBox(boolean box);
     public String getFlags();
     public void setFlags(String flags);
-    public List<Penalty> getPenalties();
-    public Penalty getFOEXPPenalty();
     // A null code removes the penalty.
     public void penalty(String id, String number, int period, int jam, String code);
+    public void sortPenalties();
```
brian-brazil, line 37: This doesn't feel right, I'd expect this to be managed internal to the penalty handling.

src/com/carolinarollergirls/scoreboard/core/impl/RulesetsImpl.java
```diff
@@ -36,29 +32,72 @@
...
+	children.get(Child.RULESET).put(ROOT_ID, root);
         for (Rule r : Rule.values()) {
             r.getRuleDefinition().setParent(this);
             r.getRuleDefinition().setIndex(r.ordinal());
-            rules.put(r.toString(), r);
-            rootSettings.put(r, r.getRuleDefinition().getDefaultValue());
+            children.get(Child.RULE_DEFINITION).put(r.toString(), r.getRuleDefinition());
+            root.add(Ruleset.Child.RULE, new ValWithId(r.toString(), r.getRuleDefinition().getDefaultValue()));
         }
-        // The root ruleset is always created from the above list,
```
brian-brazil, line 47: This comment should stay

src/com/carolinarollergirls/scoreboard/jetty/WS.java
```diff
@@ -127,9 +135,12 @@ public synchronized void onMessage(String message_data) {
                         v = value.toString();
                     }
                     ScoreBoardManager.printMessage("Setting " + key + " to " + v);
-
-                    if (key.startsWith("ScoreBoard.Settings.")) {
-                        sb.getSettings().set(key.substring(20), v);
+                    Matcher m = pathElementPattern.matcher(key);
+                    if (m.matches() && m.group("name").equals("ScoreBoard") &&
```
brian-brazil, line 122: This is duplicating the check inside set

html/controls/rulesets.js
```diff
@@ -235,50 +235,45 @@ function displayRuleset(id) {
...
-	};
 	// Note what rulesets we know about so we can determine
 	// the id of the new one when we get an update.
 	knownRulesets = {};
 	$.each(rulesets, function(id) {
 		knownRulesets[id] = true;
 	});
+	var uuid;
+	do {
+		uuid = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {var r = Math.random()*16|0,v=c=='x'?r:r&0x3|0x8;return v.toString(16);}).toUpperCase();
```
brian-brazil, line 246: Apparently Javascript's RNG isn't the best, we probably want to add in the current time here

html/controls/pt/index.js
```diff
@@ -309,12 +308,13 @@
 	}
 
 	function submitPenalty() {
-		var p = $('.PenaltyEditor .Period').val();
-		var j = $('.PenaltyEditor .Jam').val();
-		var c = $('.PenaltyEditor .Codes .Active').attr('code');
-		var fo_exp = penaltyType === 'FO_EXP';
-
-		WS.Command("Penalty", { teamId: teamId, skaterId: skaterId, penaltyId: penaltyId, period: p, jam: j, code: c, fo_exp: fo_exp });
+		var sortKey = 'ScoreBoard.Team(' + teamId + ').Skater(' + skaterId + ').SortPenalties';
```
brian-brazil, line 311: This feels messy, and also could be racey if there's two PTs. I think we need a way to take this all in at once

tests/com/carolinarollergirls/scoreboard/core/impl/SkaterImplTests.java
```diff
@@ -45,7 +54,13 @@ public void add_penalty() {
 
     @Test
     public void add_penalty_with_id() {
-        skater.penalty("f03d5e2e-e581-4fcb-99c7-7fbd49101a36", "0", 1, 3, "C");
+	skater.set(Value.SORT_PENALTIES, false);
+	Penalty p = (Penalty)skater.get(Child.PENALTY, "0", true);
+	p.set(Penalty.Value.ID, "f03d5e2e-e581-4fcb-99c7-7fbd49101a36", Flag.FORCE);
```
brian-brazil, line 59: Why does this need to be forced?

src/com/carolinarollergirls/scoreboard/event/ScoreBoardEventProvider.java
```diff
@@ -18,32 +18,85 @@
...
     public ValueWithId childFromString(AddRemoveProperty prop, String id, String sValue);
+    /*
+     * Will return null if no such child is found
+     */
     public ValueWithId get(AddRemoveProperty prop, String id);
+    /*
+     * If no such child is found and add is true, call create(prop, id) to create one.
+     * If creation is not implemented for this type of child or add is false, null is returned.
+     */
     public ValueWithId get(AddRemoveProperty prop, String id, boolean add);
```
brian-brazil, line 81: getOrAdd would be clearer than a bool

brian-brazil (commented):
I've still to give it a run, but here's a first pass.

frank-weinberg:
This is superseded by #221

PR #217, opened 2019-01-28
Release 3.9.6

JeneralPain:
Fixes quite a bit from the looks of it?


frank-weinberg:
This should not be a 3.9.x release. We have agreed that 3.9.6 should be 3.9.5 + bug fixes but no interface changes (and thus no or at least almost no new features).

JeneralPain:
Cancelling PR, need to re-review how to add just what needs to be added.

PR #220, opened 2019-03-24
Add banners to WS scoreboard.

### LINKED ISSUE #75: Web Socket view in 3.9.4-a and 3.9.3 does not display Ads
We have confirmed in 3.9.3 and the test build of 3.9.4a when utilising the web socket view for the scoreboard, the ad banners loaded do not display.

They do however work in the legacy display.

brian-brazil:
Fixes #75

Fix legacy scoreboard banners in passing.

This is the last feature that the WS scoreboard is missing, so we can delete the legacy scoreboard next.

### DIFF
--- a/html/views/scoreboard.js
+++ b/html/views/scoreboard.js
@@ -417,7 +417,7 @@ function showClocks(clock) {
 function setupSponsorBanners() {
 	var div = $("#SponsorBox");
 	var setNextSrc = function() {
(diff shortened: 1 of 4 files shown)


JeneralPain:
@brian-brazil or @frank-weinberg 
Was this merged into dev or has it been reversed out because it looks like you deleted it @brian-brazil ?

PR #221, opened 2019-04-01
Support complete WFTDA statsbook

brian-brazil:
To make review easier...


config/penalties/wftda2018.json
```diff
@@ -1,5 +1,6 @@
 {
 	"penalties": [
+		{ "code": "?", "verbalCues": ["Unknown"]},
```
brian-brazil, line 3: What about the other file?

html/controls/operator.js
```diff
@@ -765,22 +764,62 @@ function createTeamTable() {
...
+		injButton.appendTo(leadJammerTd);
+		var ni = sbTeam.$sb("NoInitial");
+		var niButton = ni.$sbControl("<button>").text("NI").val("true")
+			.attr("id", "Team"+team+"NI").addClass("KeyControl").button();
+		ni.$sbBindAndRun("sbchange", function(event, value) {
+			niButton.val(!isTrue(value));
+			niButton.toggleClass("Active", isTrue(value));
+		});
+		niButton.appendTo(leadJammerTd);
 		/* some strange bug, css direction is unset for leadJammerTd
```
brian-brazil: Is this still the case?

html/controls/operator.js
```diff
@@ -892,6 +928,163 @@ function createTimeTable() {
...
+				if (WS.state[prefix + '.WalltimeEnd'] == 0 && WS.state[prefix + '.WalltimeStart'] > 0) {
+					row.find("td.Duration").text("running");
+				} else {
+					row.find("td.Duration").text(_timeConversions.msToMinSec(v));
+				}
+			}
+		}
+	});
+	
+	WS.Register(['ScoreBoard.CurrentPeriodNumber'], function(k, v) {
```
brian-brazil, line 1071: This is a bit unsafe, if we want exact matches we should add a function for that.

html/controls/plt/plt.js
```diff
@@ -0,0 +1,477 @@
...
+			WS.Set("ScoreBoard.Team(" + teamId + ").Skater(" + skaterId + ").Penalty(" + penaltyNumber + ").Remove", true);
+			penaltyEditor.dialog('close');
+		}
+	}
+
+	var codeIdRegex = /Code\(([^\)]+)\)/;
+	function penaltyCode(k, v) {
+		var match = (k || "").match(codeIdRegex);
+		if (match == null || match.length == 0)
+			return;
```
brian-brazil: Please use explicit braces here

html/json/WS.js
```diff
@@ -93,11 +93,12 @@ var WS = {
 		WS.send(JSON.stringify(req));
 	},
 
-	Set: function(key, value) {
+	Set: function(key, value, flag = '') {
```
brian-brazil: Hmm, what browsers do we support? Default values aren't that old.

src/com/carolinarollergirls/scoreboard/core/BoxTrip.java
```diff
@@ -0,0 +1,62 @@
...
+        ID(String.class, ""),
+        IS_CURRENT(Boolean.class, false),
+        CURRENT_FIELDING(Fielding.class, null),
+        START_FIELDING(Fielding.class, null),
+        START_JAM_NUMBER(Integer.class, 0),
+        START_BETWEEN_JAMS(Boolean.class, false),
+        START_AFTER_S_P(Boolean.class, false),
+        END_FIELDING(Fielding.class, null),
+        END_JAM_NUMBER(Integer.class, 0),
+        END_BETWEEN_JAMS(Boolean.class, false),
```
brian-brazil, line 33: What's this for?

src/com/carolinarollergirls/scoreboard/core/Clock.java
```diff
@@ -20,12 +22,10 @@
...
     public void start();
     public void stop();
-    public void startNext();
-
-    public String getId();
+    public void restart();
 
     public String getName();
-    public void setName(String name);
+    //    public void setName(String name);
```
brian-brazil: Can it be removed?

src/com/carolinarollergirls/scoreboard/core/Jam.java
```diff
@@ -0,0 +1,53 @@
...
+    public long getWalltimeEnd();
+
+    public TeamJam getTeamJam(String id);
+
+    public void start();
+    public void stop();
+
+    public enum Value implements PermanentProperty {
+        PERIOD_NUMBER(Integer.class, 0),
+        STAR_PASS(Boolean.class, false),
```
brian-brazil: This could be named better, it's not obvious how this is modeled.

src/com/carolinarollergirls/scoreboard/core/ScoreBoard.java
```diff
@@ -47,6 +55,14 @@
...
     public boolean isInPeriod();
     public void setInPeriod(boolean inPeriod);
+    public Period getOrCreatePeriod(int p);
+    public Period getCurrentPeriod();
+    public int getCurrentPeriodNumber();
+
+    public boolean isInJam();
+    public Jam getUpcomingJam();
+    
+    public void updateTeamJams();
```
brian-brazil, line 66: A comment would be useful here

src/com/carolinarollergirls/scoreboard/core/ScoreBoard.java
```diff
@@ -84,26 +96,63 @@
...
-    public static final String EVENT_OFFICIAL_REVIEW = "OfficialReview";
-    public static final String EVENT_ADD_CLOCK = "AddClock";
-    public static final String EVENT_REMOVE_CLOCK = "RemoveClock";
-    public static final String EVENT_ADD_TEAM = "AddTeam";
-    public static final String EVENT_REMOVE_TEAM = "RemoveTeam";
-    public static final String EVENT_SETTING = "Setting";
+    public enum Value implements PermanentProperty {
+        CURRENT_PERIOD_NUMBER(Integer.class, 0),
+        CURRENT_PERIOD(Period.class, null),
+        UPCOMING_JAM(Jam.class, null),
```
brian-brazil, line 109: Why upcoming and not current?

src/com/carolinarollergirls/scoreboard/event/ScoreBoardEventProviderImpl.java
```diff
@@ -0,0 +1,446 @@
...
+        InverseReferenceUpdateListener l = new InverseReferenceUpdateListener(this, localProperty, remoteProperty);
+        addScoreBoardListener(l);
+        return l;
+    }
+
+    public Object valueFromString(PermanentProperty prop, String sValue, Flag flag) {
+        synchronized (coreLock) {
+            @SuppressWarnings("rawtypes")
+            Class type = prop.getType();
+            if (type == TimeoutOwner.class) { return scoreBoard.getTimeoutOwner(sValue); }
```
brian-brazil, line 239: This is a layering violation, this should come from the object rather than this class

src/com/carolinarollergirls/scoreboard/event/ScoreBoardEventProviderImpl.java
```diff
@@ -0,0 +1,446 @@
...
+        if (readonly) {
+            addWriteProtectionOverride(targetProperty, Flag.COPY);
+        } else {
+            reverseCopyListeners.put(targetProperty, 
+                    new CopyScoreBoardListener(sourceElement, sourceProperty, false));
+        }
+        set(targetProperty, sourceElement.get(sourceProperty), Flag.COPY);
+        return l;
+    }
+    protected ScoreBoardListener setCopy(final PermanentProperty targetProperty, ScoreBoardEventProvider indirectionElement,
```
brian-brazil, line 195: Comments would be useful to understand what indirection, inverse, copy, recalcuate etc. mean.

src/com/carolinarollergirls/scoreboard/utils/Comparators.java
```diff
@@ -0,0 +1,71 @@
+package com.carolinarollergirls.scoreboard.utils;
+
+import java.util.Comparator;
+
+import com.carolinarollergirls.scoreboard.core.BoxTrip;
```
brian-brazil: Layering violation, utils shouldn't be using core.

src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
```diff
@@ -189,62 +193,14 @@ public long getTimeRemaining() {
...
-    protected long checkNewTime(long ms) {
-        if (ms < minimumTime && minimumTime - ms > 500) {
-            return minimumTime;
-        } else if (ms > maximumTime && ms - maximumTime > 500) {
-            return maximumTime;
-        } else {
-            return ms;
-        }
-    }
+    public boolean resetTime() { return set(Value.TIME, getTime(), Flag.RESET); }
```
brian-brazil: YYou don't use this return value

brian-brazil (commented):
This does make things simpler, and I see what you mean about Java 8

html/controls/operator.js
frank-weinberg: That sounds worthy of a separate PR.

html/json/WS.js
frank-weinberg: Officially we supprt Firefox and Chrome. I don't think we give a minimum version anywhere.

src/com/carolinarollergirls/scoreboard/core/BoxTrip.java
frank-weinberg: Box Trips ending between Jams, e.g. because the Penalty is removed.

src/com/carolinarollergirls/scoreboard/core/Jam.java
frank-weinberg: Suggestion?

src/com/carolinarollergirls/scoreboard/core/ScoreBoard.java
frank-weinberg: Because this will hold a Jam until it becomes current, at which point the jam is moved to the current period and this will hold the next jam.

src/com/carolinarollergirls/scoreboard/event/ScoreBoardEventProviderImpl.java
frank-weinberg: I don't understand this comment.

frank-weinberg:
Unittests are in need of a general overhaul anyways. Some of the existing ones have become redundant, some should be in different files now. And most of the new code is not covered at all yet. That's on my list of things to do before a stable 4.x release. (I don't think we need to have them for a beta.)

brian-brazil:
For a specific case where there was a bug, a unittest should really be included.

html/controls/operator.js
brian-brazil: Sure

src/com/carolinarollergirls/scoreboard/event/ScoreBoardEventProviderImpl.java
brian-brazil: Check if the fromString method exists via an interface, rather than putting a list of classes here. We should only need to special case the POJOs.

html/json/WS.js
brian-brazil: Default values only started being a thing in 2015. I don't think it's safe to assume everyone has that.

src/com/carolinarollergirls/scoreboard/core/Jam.java
brian-brazil: Actually a comment that this is if either team had one should do it.

src/com/carolinarollergirls/scoreboard/event/ScoreBoardEventProviderImpl.java
brian-brazil: It's still not clear to me what an indirectionProperty is, the comment is mostly repeating the type signature.

src/com/carolinarollergirls/scoreboard/core/Comparators.java
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+    }
+
+    public static Comparator<Period> PeriodComparator = new SbepComparator<Period>();
+    public static Comparator<Jam> JamComparator = new SbepComparator<Jam>();
+    public static Comparator<TeamJam> TeamJamComparator = new SbepComparator<TeamJam>();
+    public static Comparator<ScoringTrip> ScoringTripComparator = new SbepComparator<ScoringTrip>();
+    public static Comparator<Fielding> FieldingComparator = new SbepComparator<Fielding>();
+    public static Comparator<BoxTrip> BoxTripComparator = new Comparator<BoxTrip>() {
```
brian-brazil: Digging a bit more, this would better live in the BoxTrip class implementing the Comparable interface: https://docs.oracle.com/javase/7/docs/api/java/lang/Comparable.html

List.sort will use that automatically then.

src/com/carolinarollergirls/scoreboard/event/ScoreBoardEventProviderImpl.java
frank-weinberg: Specifically for TimeoutOwner this does not really work. As we don't have an Object yet, the fromString() method needs to be static. But the list of possible `TimeoutOwner`s includes `Team`s, which are only created at runtime in a specific `ScoreBoardImpl` object. So we would need to
a) inject the current `ScoreBoard` instance into a static variable in `TimeoutOwner` so `fromString()` can reference it.
b) use reflection to call the static `fromString()` method (this would also be required for the other classes)
I don't think that is an improvement over what we have now.

brian-brazil: You could pass the scoreBoard object as a parameter to valueFromString. You could also move this all over to the ScoreBoard object.

> I don't think that is an improvement over what we have now.

Two packages mutually depending on each other is to be avoided. The event provider code shouldn't know what types of object it's working with.

frank-weinberg: If we want to hand the `scoreboard` object as a method parameter, it has to come from the `json` resp. `xml` package, two packages that also are referenced from `core` - we would again just shift the circular dependency around (and if we don't remove the other uses of the object in SBEPI, we don't even remove a dependency).

I argue that the `scoreboard` object is at the heart of this program and trying to isolate it in the `core` package is going to cause a huge amount of overhead and will significantly decrease the maintainability of the code, just for the sake of following a principle, as we can be rather certain that this class isn't going away in a refactoring. So this is a case, where a deviation from the principle is justified.

brian-brazil: The only thing even using this method is WS/xml, which can get it from the scoreboard object they already have directly. There's no reason for this code to be here.

brian-brazil:
:+1: 

I don't 100% agree with everything here and there's known bugs, however the right thing to do is merge and iterate rather than trying to work on a 12k line PR.

PR #223, opened 2019-04-12
When deactivating Retained, remove the OR.

brian-brazil:
Make retained button green rather than red, it's
not a "problem" to be retained.


frank-weinberg:
Haven't tried, but from reading the code, it seems that this will remove the OR when (accidentally) turning retained on and off before an OR was taken.

brian-brazil:
That's true, I guess I should wait until you've added the timeout stats and then fix this?

frank-weinberg:
The way I plan to do them, this should be fixed as a side effect of adding timeout stats.

PR #226, opened 2019-04-12
Recalculate score after sp when jam trips are removed.

brian-brazil:
Also make it simpler to recalculate based on an AddRemoveProperty
having something added/removed.


frank-weinberg:
Shouldn't we allow arbitrary watched Properties on indirect sources as well?

brian-brazil:
I still don't fully understand that bit of code, so didn't want to go further out of scope.

Also added my unittests.

PR #228, opened 2019-04-12
Fix various issues on the LT sheets:

frank-weinberg:
- update box trips symbols on every possibly relevant change
- use correct comparator when sorting box trips
- display correct cells on copy to statsbook screen

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
@@ -83,6 +83,14 @@ public class BoxTripImpl extends ScoreBoardEventProviderImpl implements BoxTrip 
                 ((Fielding)f).set(Fielding.Value.PENALTY_BOX, false);
             }
         }
+        if ((prop == Value.END_FIELDING || prop == Value.END_AFTER_S_P || prop == Value.END_BETWEEN_JAMS)
+                && getEndFielding() != null) {
+            getEndFielding().updateBoxTripSymbols();
+        }
+        if ((prop == Value.START_FIELDING || prop == Value.START_AFTER_S_P || prop == Value.START_BETWEEN_JAMS)
+                && getStartFielding() != null) {
+            getStartFielding().updateBoxTripSymbols();
+        }
     }
     
     protected void itemAdded(AddRemoveProperty prop, ValueWithId item) {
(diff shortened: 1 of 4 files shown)


src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
```diff
@@ -83,6 +83,14 @@ protected void valueChanged(PermanentProperty prop, Object value, Object last, F
                 ((Fielding)f).set(Fielding.Value.PENALTY_BOX, false);
             }
         }
+        if ((prop == Value.END_FIELDING || prop == Value.END_AFTER_S_P || prop == Value.END_BETWEEN_JAMS)
```
brian-brazil, line 86: Should we use recalculated here rather than hand-coding this?

src/com/carolinarollergirls/scoreboard/core/impl/FieldingImpl.java
```diff
@@ -114,7 +121,12 @@ public void updateBoxTripSymbols() {
         for (ValueWithId v : getAll(Child.BOX_TRIP)) {
             trips.add((BoxTrip) v);
         }
-        Collections.sort(trips);
+        Collections.sort(trips, new Comparator<BoxTrip>() {
+            public int compare(BoxTrip b1, BoxTrip b2) {
+                if (b1 == b2) { return 0; }
```
brian-brazil, line 126: This is redundant

brian-brazil (commented):
A few unittests to cover the fixed behaviour would be good.

src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
frank-weinberg: This is not a good fit for recalculate:
1. The fielding does not know if it's the start or end of any BoxTrip, so we would need nontrivial logic to maintain the listeners.
2. updateBoxTripSymbols() updates 3 values at once. That doesn't fit well into computeValue. And going from computeValue to a function that will in turn set the value requires fiddling around with flags.

So overall this solution is a lot cleaner.

src/com/carolinarollergirls/scoreboard/core/impl/FieldingImpl.java
frank-weinberg: It catches the case where b1 == b2 == null.

brian-brazil: That's already covered by the next two lines

frank-weinberg: No. The next line would return 1 instead of 0 in that case.

brian-brazil:
This still needs tests.

frank-weinberg:
Added a unittest for the sorting issue. The frontend issues can't be covered.

PR #229, opened 2019-04-12
Remove old scoreboard.

brian-brazil:
The WS scoreboard is now feature-equivilent.

Remove minimal animations option, it does
nothing for WS.


frank-weinberg:
I object to removing side padding:
1. I have seen problems where browser or OS have cut off the edges. No projector will fix that.
2. Even when it can be fixed in the projector you need to have someone at hand who knows how or has time to figure it out. This is not a given for a significant portion of derby events. 

I also do not see the point of removing other box style options. Not every venue uses a projector to display the scoreboard. So other options may be of use to only a small percentage of users but keeping them is essentially no effort.

Third, those two commits probably should not be a single PR.

brian-brazil:
> I have seen problems where browser or OS have cut off the edges. No projector will fix that.

If it's a aspect ratio issue (which seems most likely), 10% isn't enough to fix it. At that stage you can either resize the browser or fix the display settings.

> Not every venue uses a projector to display the scoreboard. So other options may be of use to only a small percentage of users but keeping them is essentially no effort.

It's still something to maintain, and the output should be as bright as possible in any case.

> Third, those two commits probably should not be a single PR.

I was being lazy.

frank-weinberg:
> > I have seen problems where browser or OS have cut off the edges. No projector will fix that.
> 
> If it's a aspect ratio issue (which seems most likely), 10% isn't enough to fix it. At that stage you can either resize the browser or fix the display settings.

Resizing the Browser is not an option in full screen. And fixing the display settings can be difficult if you don't know which exact setting is the culprit, the OS is in a foreign language and/or you don't have the admin password.
I've personally had at least a dozen instances in the last few years where using this option came in handy.

> > Not every venue uses a projector to display the scoreboard. So other options may be of use to only a small percentage of users but keeping them is essentially no effort.
> 
> It's still something to maintain, and the output should be as bright as possible in any case.

We've already spent more effort on this discussion than maintaining this option took in the last 5+ years.

brian-brazil:
Updated.

PR #232, opened 2019-04-13
Make WS updates more efficient.

brian-brazil:
Use a trie to find relevant callbacks, this cuts about 90% of
WS-related CPU.

We don't need to go through all the paths when we've already found one.

### DIFF
--- a/html/json/WS.js
+++ b/html/json/WS.js
@@ -4,6 +4,7 @@ var WS = {
 	connectCallback: null,
 	connectTimeout: null,
 	callbacks: new Array(),
+	callbackTrie: {},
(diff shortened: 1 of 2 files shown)


frank-weinberg:
Maybe we should split the path into properties and IDs (i.e. split at each '.' and '(') and use those components to index the trie instead of every single letter. This would get rid of unwanted matches when one property is a substring of another.

brian-brazil:
If we do that we should do it back in the backend first, and it is an API change. We'd also have to think about escaping.

I'd suggest getting this in for now, as there's big improvements.

PR #238, opened 2019-04-16
Let PLTs manually advance the Jam between jams.

frank-weinberg:
Any box trip starts/endings entered before advancing are treated as
having occured during the jam that just ended. Ones entered after
advancing are treated as having occured between jams.

If the jam is not manually advanced, it will be automatically advanced
at the start of the next jam.

### DIFF
--- a/html/controls/plt/plt.css
+++ b/html/controls/plt/plt.css
@@ -15,17 +15,16 @@ table.PLT, .PLT table, .PLT tr, .PLT td, #PenaltyEditor table, #PenaltyEditor tr
 }
 .PLT.Team thead td { font-size: 1em; font-weight: bold; background-color: #222; color: #FFF; }
 .PLT.Team td { text-align: center; vertical-align: middle; }
-.PLT.Team .Number, .Team .Box, .PLT.Team .Total { width: 4.76%; }
-.PLT.Team .Role, .Team .Sitting, .PLT.Team .Trips { width: 7.14%; }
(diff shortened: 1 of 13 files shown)


html/controls/plt/plt.css
```diff
@@ -44,6 +43,9 @@ table.PLT, .PLT table, .PLT tr, .PLT td, #PenaltyEditor table, #PenaltyEditor tr
 .PLT.Team .Serving { background-color: #70ff70; }
 .PLT.Team .OnTrack { background-color: #70ff70; }
 .PLT.Team .inBox { background-color: #ff4040; }
+.PLT.Team .Advance { border: none; }
+.PLT.Team .Advance.Active { background-color: #70ff70; }
+.PLT.Team tr:nth-child(6n+3) .Advance.Active::after { content: "next Jam"; }
```
brian-brazil: Next Jam

Green usually indicates a button/state, so this being used as a general background is a bit unusual.

frank-weinberg: It is a button and it indicates a state.

brian-brazil: I see now, that wasn't obvious.  I'd also expect green to mean that we're already in the next Jam state. Text along the lines of "Start entering upcoming jam" would be clearer.

I've also found that if you add a skater on the LT part of team 1, it affects the toggles for team 2 (but not the other way around).

frank-weinberg: I have now made it as clear as I could with the space constraints.

Also allow for editing lineups, not just box trips, as people will want to do that if for some reason they didn't finish entering them during the Jam.

PR #244, opened 2019-04-26
Allow PLTs to mark penalties as served

frank-weinberg:
This has been a consistent feature request of all users of the PLT screen so far.


brian-brazil:
When you create a new penalty, the checkbox has whatever it's last state was rather than cleared.

frank-weinberg:
That was an easy fix.

PR #248, opened 2019-05-01
Replace PT/LT/WB screens by screens based on PLT

frank-weinberg:
Using the same code for all of these screens will make maintenance a lot
easier.


html/controls/plt/index.html
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
 	<head>
+		<meta charset="utf-8"/>
```
brian-brazil, line 4: Are we not setting this in the Java?

html/controls/lt/index.js
```diff
@@ -0,0 +1,17 @@
+(function () {
+	$('head').append('<link rel="stylesheet" href="/controls/plt/plt.css" type="text/css" />');
+	$('head').append('<link rel="stylesheet" href="lt.css" type="text/css" />');
+
+	WS.Connect();
+	WS.AutoRegister();
+
+	var teamId = _windowFunctions.getParam("team");
+
+	preparePltTable($('#plt'), teamId, 'lt');
```
brian-brazil: These seem the wrong way around, a comment at least would be useful.

Are we duplicating any code here?

html/index.html
```diff
@@ -102,18 +101,20 @@ <h2><span class="fa fa-2x fa-desktop"></span> SCOREBOARD DISPLAY</h2>
...
+            <li><a href="controls/lt/?team=1">Lineup Tracking (Team 1)</a></li>
+            <li><a href="controls/lt/?team=2">Lineup Tracking (Team 2)</a></li>
           </ul>
           <ul>
             <li><a href="controls/plt/?team=1">Penalty + Lineup Tracking (Team 1)</a></li>
             <li><a href="controls/plt/?team=2">Penalty + Lineup Tracking (Team 2)</a></li>
           </ul>
           <ul>
-            <li><a href='views/wb/whiteboard.html'>Penalty Whiteboard Display<span class="pill pill-beta">BETA</span></a></li>
-            <li><a href='views/wb/tcdg2016wb.html'>TCDG16 Penalty Graphics Display</a></li>
```
brian-brazil, line 109: The views/images directory can go now too.

html/controls/plt/index.html
frank-weinberg: Apparently not. Firefox has always been giving an error message in the console and without this line the sum character doesn't display properly.

html/controls/lt/index.js
frank-weinberg: I've changed file and function names to make it more obvious, what these parts are.

PR #249, opened 2019-05-01
Use utf-8 charset for html and other text responses.

brian-brazil:
The version of jetty we're using really doesn't
make setting this correctly easy. Hack in the charset in
the mime entries, overriding what Jetty has by default.

### DIFF
--- a/src/org/eclipse/jetty/http/mime.properties
+++ b/src/org/eclipse/jetty/http/mime.properties
@@ -1,0 +1,181 @@
+ai	= application/postscript
+aif	= audio/x-aiff
+aifc	= audio/x-aiff
+aiff	= audio/x-aiff
+apk = application/vnd.android.package-archive
+asc	= text/plain; charset=utf-8
+asf     = video/x.ms.asf
+asx     = video/x.ms.asx
+au	= audio/basic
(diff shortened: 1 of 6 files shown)


src/org/eclipse/jetty/http/mime.properties
```diff
@@ -0,0 +1,181 @@
...
+wml	= text/vnd.wap.wml; charset=utf-8
+wmlc	= application/vnd.wap.wmlc
+wmls	= text/vnd.wap.wmlscript; charset=utf-8
+wmlsc	= application/vnd.wap.wmlscriptc
+wrl	= model/vrml
+wtls-ca-certificate	= application/vnd.wap.wtls-ca-certificate
+xbm	= image/x-xbitmap
+xht	= application/xhtml+xml
+xhtml	= application/xhtml+xml
+xls	= application/vnd.ms-excel
```
frank-weinberg, line 171: Shouldn't we be including .xlsx here as well?

brian-brazil: This is only a copy of what upstream has, and I doubt we'll be serving static xlsx files.

PR #251, opened 2019-05-04
Make registering multiple WS names work as you'd expect.

brian-brazil:
This now provides the first non-null value, so you don't
have to add that logic by hand.

Simplify how scoreboard team names are handled.

This incidentally makes it so that sbDisplay will
now only pick up exact key matches.

### DIFF
--- a/html/json/WS.js
+++ b/html/json/WS.js
@@ -270,8 +270,26 @@ var WS = {
 		$.each($("[sbDisplay]"), function(idx, elem) {
 			elem = $(elem);
 			var paths = WS.getPaths(elem, "sbDisplay");
-			if (paths.length > 0)
-				WS.Register(paths, { element: elem, modifyFunc: window[elem.attr("sbModify")] });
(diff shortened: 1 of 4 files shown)


html/json/WS.js
```diff
@@ -270,8 +270,26 @@ var WS = {
 		$.each($("[sbDisplay]"), function(idx, elem) {
 			elem = $(elem);
 			var paths = WS.getPaths(elem, "sbDisplay");
-			if (paths.length > 0)
-				WS.Register(paths, { element: elem, modifyFunc: window[elem.attr("sbModify")] });
+			if (paths.length > 0) {
+				// When there's multiple names, use the
+				// first non-empty one.
+				var mf = function(k, v) {
+					var v = null;
```
frank-weinberg, line 277: The local variable shouldn't have the same name as the function parameter.

brian-brazil: Done

PR #252, opened 2019-05-04
Fix team colours and copy.html

brian-brazil:
Use the same default background for everywhere
we use the team colours on an operator page, so
that it won't be visible for the SBO and not others.

### DIFF
--- a/html/controls/plt/plt-input.css
+++ b/html/controls/plt/plt-input.css
@@ -14,7 +14,10 @@ table.PLT, .PLT table, .PLT tr, .PLT td, #PenaltyEditor table, #PenaltyEditor tr
 	top: 0%;
 }
 /* Colors are the same as the operator controls, so per-team colours look the same on both. */
-.PLT.Team thead td { font-size: 200%; font-weight: bold; background-color: #cccccc; color: #333333; }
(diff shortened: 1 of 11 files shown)


brian-brazil:
I've added a few more improvements for this one. The IWB still needs work colour wise, but at least the team colours will be sanely displayed.

html/controls/plt/plt-input.css
```diff
@@ -13,7 +13,12 @@ table.PLT, .PLT table, .PLT tr, .PLT td, #PenaltyEditor table, #PenaltyEditor tr
 	width: 99%;
 	top: 0%;
 }
-.PLT.Team thead td { font-size: 1em; font-weight: bold; background-color: #222; color: #FFF; }
+/* Colors are the same as the operator controls, so per-team colours look the same on both. */
+.PLT.Team thead td { font-size: 200%; font-weight: bold; background-color: #cccccc; color: #333333; }
```
frank-weinberg: This font-size breaks the screen on typical tablet resolutions.

brian-brazil: Bad copy on my part, I prefer the smaller size. Fixed.

PR #256, opened 2019-05-09
Select team after creating it.

brian-brazil:
This is a common confusion. This is hacky, but we can
improve once this is on WS.

### DIFF
--- a/html/controls/operator.js
+++ b/html/controls/operator.js
@@ -1435,6 +1435,8 @@ function createTeamsContent() {
 		var teamid = _crgUtils.checkSbId(teamname);
 		$sb("Teams.Team("+teamid+").Name").$sbSet(teamname);
 		newTeamName.val("").keyup().focus();
+		// Presume we'll have gotten the XML update within 100ms.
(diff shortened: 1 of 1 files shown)


html/controls/operator.js
```diff
@@ -1435,6 +1435,8 @@ function createTeamsContent() {
 		var teamid = _crgUtils.checkSbId(teamname);
 		$sb("Teams.Team("+teamid+").Name").$sbSet(teamname);
 		newTeamName.val("").keyup().focus();
+		// Presume we'll have gotten the XML update within 100ms.
+		setTimeout(function(){selectTeam.val(teamname);}, 100);
```
frank-weinberg, line 1439: How certain are you that 100ms is sufficient, especially when the operator screen is not on the same machine as the backend?

brian-brazil: It'll all be in the same sports hall, and thus within 1-2ms. I'd be more worried about the backend being slow.

frank-weinberg: This only selects the team in the dropdown but doesn't change the bottom part.

PR #257, opened 2019-05-10
Link to the Copy endpoint, add instructions.

brian-brazil:


frank-weinberg:
I have intentionally not linked this from the main page. IMO, this screen should not be readily available to users without significantly more documentation including caveats, so we don't get overwhelmed with support questions because people break their statsbook trying to use it. (I was thinking of putting this on a wiki page and documenting the URL there.)

The default way to get data from CRG to a statsbook should be Smasher's tool.

brian-brazil:
I think this is very useful, and worth linking to.

brian-brazil:
We could put a general caveat here, and link to Smasher's tool too.

frank-weinberg:
Something along the lines of "the statsbook may break on improper pasting, don't come to us for support" and a link to smasher's tool indicating it is the preferred way would be the minimum I'd want to see.
Once there is a more detailed wiki page, we should probably include a link to that as well.

brian-brazil:
Done

PR #258, opened 2019-05-10
Fix various issues around undo and deletion.

brian-brazil:
Fix NPE on start jam -> undo -> start jam.
Ensure we can start a jam after deleting current period.
Fix missing requestBatchEnd when deleting jams.
Fix deleting jams

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/impl/JamImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/JamImpl.java
@@ -82,6 +82,7 @@ public class JamImpl extends NumberedScoreBoardEventProviderImpl<Jam> implements
                 scoreBoard.updateTeamJams();
                 ((ScoreBoardEventProviderImpl) parent).scoreBoardChange(
                         new ScoreBoardEvent(this, BatchEvent.END, Boolean.TRUE, Boolean.TRUE));
+                requestBatchEnd();
                 break;
             case INSERT_BEFORE:
                 requestBatchStart();
--- a/src/com/carolinarollergirls/scoreboard/core/impl/PeriodImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/PeriodImpl.java
@@ -90,7 +90,10 @@ public class PeriodImpl extends NumberedScoreBoardEventProviderImpl<Period> impl
     public void restoreSnapshot(PeriodSnapshot s) {
         synchronized (coreLock) {
             if (s.getId() != getId()) {	return; }
-            set(Value.CURRENT_JAM, s.getCurrentJam());
+            if (getCurrentJam() != s.getCurrentJam()) {
+              remove(NChild.JAM, getCurrentJam());
+              set(Value.CURRENT_JAM, s.getCurrentJam());
+            }
         }
     }
 
--- a/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
@@ -83,11 +83,19 @@ public class ScoreBoardImpl extends ScoreBoardEventProviderImpl implements Score
                 lc.setMaximumTime(getRulesets().getLong(Rule.LINEUP_DURATION));
             }
         } else if (prop == Value.UPCOMING_JAM) {
-            removeAll(Period.NChild.JAM);
             if (value instanceof Jam) {
                 add(Period.NChild.JAM, (Jam)value);
             }
         }
+    }
+
+    @Override
+    protected void itemRemoved(AddRemoveProperty prop, ValueWithId item) {
+      if (prop == NChild.PERIOD) {
+        // If the current period was removed, assign
+        // the upcoming jam to the now current period.
+        getCurrentPeriod().getCurrentJam().setNext(getUpcomingJam());
+      }
     }
 
     @Override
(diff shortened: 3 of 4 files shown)


src/com/carolinarollergirls/scoreboard/core/impl/JamImpl.java
```diff
@@ -82,6 +82,7 @@ public void execute(CommandProperty prop) {
                 scoreBoard.updateTeamJams();
                 ((ScoreBoardEventProviderImpl) parent).scoreBoardChange(
                         new ScoreBoardEvent(this, BatchEvent.END, Boolean.TRUE, Boolean.TRUE));
+                requestBatchEnd();
```
frank-weinberg, line 85: requestBatchEnd() doesn't work here, as parent is no longer listening to events from this. The instruction above sends the Batch end explicitly to parent instead.

src/com/carolinarollergirls/scoreboard/core/impl/PeriodImpl.java
```diff
@@ -90,7 +90,10 @@ public PeriodSnapshot snapshot() {
     public void restoreSnapshot(PeriodSnapshot s) {
         synchronized (coreLock) {
             if (s.getId() != getId()) {	return; }
-            set(Value.CURRENT_JAM, s.getCurrentJam());
+            if (getCurrentJam() != s.getCurrentJam()) {
+              set(Value.CURRENT_JAM, s.getCurrentJam());
+              getLast(NChild.JAM).unlink();
```
frank-weinberg: The jam should be moved, not deleted. Deleting will remove all the lineups etc that have been entered for the jam.

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
```diff
@@ -83,13 +83,21 @@ protected void valueChanged(PermanentProperty prop, Object value, Object last, F
...
             }
         }
     }
 
+    @Override
+    protected void itemRemoved(AddRemoveProperty prop, ValueWithId item) {
+      if (prop == NChild.PERIOD) {
+        // If the current period was removed, assign
+        // the upcoming jam to the now current period.
+        getCurrentPeriod().getCurrentJam().setNext(getUpcomingJam());
```
frank-weinberg, line 97: This should not be needed. Next-links should be updated under the hood if a jam is removed. If they aren't, that's a bug elsewhere.

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
```diff
@@ -83,13 +83,21 @@ protected void valueChanged(PermanentProperty prop, Object value, Object last, F
                 lc.setMaximumTime(getRulesets().getLong(Rule.LINEUP_DURATION));
             }
         } else if (prop == Value.UPCOMING_JAM) {
-            removeAll(Period.NChild.JAM);
```
frank-weinberg, line 86: This line was required because on restore from autosave there could be multiple jams, leading to problems. I don't see any code addressing that elsewhere in the PR.

frank-weinberg (changes requested):

src/com/carolinarollergirls/scoreboard/core/impl/JamImpl.java
brian-brazil: It's causing a test failure without it as the batch count doesn't line up.

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
brian-brazil: How can that happen? This is deleting jams incorrectly currently.

src/com/carolinarollergirls/scoreboard/core/impl/PeriodImpl.java
brian-brazil: That's fixed.

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
brian-brazil: This is a whole period being removed, and the jam->jam link crosses period boundaries so needs special handling.

frank-weinberg: That case should be covered by the under the hood code.

tests/com/carolinarollergirls/scoreboard/core/impl/ScoreboardImplTests.java
```diff
@@ -1530,15 +1555,15 @@ public void testPeriodJamInsertRemove() {
         assertEquals(4, p1.getAll(Period.NChild.JAM).size());
         assertEquals(1, p1.getMinNumber(Period.NChild.JAM) + 0);
         assertEquals(4, p1.getMaxNumber(Period.NChild.JAM) + 0);
-        assertEquals(1, sb.getAll(Period.NChild.JAM).size());
-        assertEquals(5, sb.getMinNumber(Period.NChild.JAM) + 0);
+        assertEquals(5, sb.getAll(Period.NChild.JAM).size());
```
frank-weinberg, line 1558: Another red flag: This should not contain anything but the upcoming jam.

brian-brazil: Why would we only have 1 jam in a period when we're on jam 5 of that period?

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
brian-brazil: Where should that be happening exactly?

tests/com/carolinarollergirls/scoreboard/core/impl/ScoreboardImplTests.java
frank-weinberg: sb is not a period but the scoreboard object.

frank-weinberg:
I've added #265 that should supersede this one.
And I've taken note that I should document all the assumptions underlying the new code sooner rather than later.

PR #261, opened 2019-05-12
If a player is un-fouledout they're eligible again.

brian-brazil:


frank-weinberg:
Oh, this also explains the report I got that a skater was not marked eligible again after they had sat out their 3 jams.

PR #266, opened 2019-05-14
Switch operator jammer/pivot to WS, adjust logic.

brian-brazil:
With AdvanceFielding, the existing code no longer really
works. This is now better, however weird stuff can happen
e.g. you set the pivot and jammer to the same skater
while the fielding hasn't been advanced.

In practice unlikely to be an issue, enough to unbreak
non-ePLT setups and will also work in ePLT setups where the
ePLT is the person entering skaters.

The proper way to fix this involves backend changes, but this should do for now. The operator always affects the running or upcoming jam.

### DIFF
--- a/html/controls/operator.js
+++ b/html/controls/operator.js
@@ -824,20 +824,63 @@ function createTeamTable() {
 
 		starPassTd.buttonset();
 
-		var makeSkaterDropdown = function(pos, elem, sort) {
-			var sortFunc = _windowFunctions.alphaCompareByProp;
-			if (sort == "Num") sortFunc = _windowFunctions.numCompareByProp;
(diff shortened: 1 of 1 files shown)


frank-weinberg (changes requested):
We should migrate the setupSelect function from crgUtils (and probably others) to WS and use that (also in existing WS screens) instead of writing (almost) the same code a dozen times.

The InJam/not in jam distinction shouldn't be necessary as Team.Position and Team.NoPivot are hardwired to CurrentOrUpcomingTeamJam.

A cleaner fix via the backend is in #270.

brian-brazil:
> We should migrate the setupSelect function from crgUtils (and probably others) to WS and use that (also in existing WS screens) instead of writing (almost) the same code a dozen times.

Agreed, this was intended as the first set in doing so.

PR #268, opened 2019-05-15
Bug fixes

brian-brazil:
Various small things I found when creating the videos.

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
@@ -55,7 +55,7 @@ public class ClockImpl extends ScoreBoardEventProviderImpl implements Clock {
                     value = (((Long)value / 1000L) * 1000L) + (Long)last % 1000L;
                 }
             }
-            if ((flag == Flag.RESET && isCountDirectionDown()) || (Long)value > getMaximumTime() + 500) {
+            if ((flag == Flag.RESET && isCountDirectionDown()) || (Long)value > getMaximumTime() + 500 && (!isCountDirectionDown() || flag != Flag.RECALCULATE)) {
                 return getMaximumTime();
             }
             if ((flag == Flag.RESET && !isCountDirectionDown()) || (Long)value < getMinimumTime() - 500) {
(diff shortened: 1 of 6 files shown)


src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
```diff
@@ -150,6 +150,7 @@ public void scoreBoardChange(ScoreBoardEvent event) {
             } else {
                 setMaximumTime(DEFAULT_MAXIMUM_TIME);
             }
+            resetTime();
```
frank-weinberg: This will cause breakage if the ruleset is changed when a clock is running.

And what unexpected time are you seeing? Changing maximum time is supposed to keep timeElapsed constant, which has just been reset to 0 in "Start new game".

brian-brazil: Rulesets can only be changed at Start New Game.

If you have a 20m period time from a ruleset, the period clock will show as 10m after the State New Game.

frank-weinberg: > Rulesets can only be changed at Start New Game.

That's true right now, but we have already thought about changing that.

>If you have a 20m period time from a ruleset, the period clock will show as 10m after the State New Game.

That's probably a race between capping the time at maximum time and reducing the time on a countdown clock when maximum time is reduced. It seems the former is applied first, while it should be the latter. This should be fixed in that code (especially as it affects more cases).

brian-brazil: Okay, found that. PTAL.

src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
```diff
@@ -55,7 +55,7 @@ else if (flag != Flag.INTERNAL) {
                     value = (((Long)value / 1000L) * 1000L) + (Long)last % 1000L;
                 }
             }
-            if ((flag == Flag.RESET && isCountDirectionDown()) || (Long)value > getMaximumTime() + 500) {
+            if ((flag == Flag.RESET && isCountDirectionDown()) || (Long)value > getMaximumTime() + 500 && !isCountDirectionDown()) {
```
frank-weinberg: This would allow the user to set a time above max for countdown clocks.
Using `(!isCountDirectionDown() || flag != Flag.RECALCULATE)` should work. (Recalculate is only triggered when min or max are changed. Changing min can't trigger this condition, changing max is supposed to be excluded.)

brian-brazil: Done

PR #270, opened 2019-05-17
Position related improvements & bugfixes

frank-weinberg:
* When skaters are marked as "Not Skating" or bench staff, don't show them in selection dropdowns or the PT/LT/PLT input
* When the operator enters a pivot, mark them as pivot (supersedes #266)
* Only mark "No Pivot", if the position is explicitly marked as "No Skater fielded" or 4 Blockers have been entered. Add "No Pivot" Marker to Operator screen.
* Don't allow a star pass if there is no pivot.
* Fix the bug where having multiple operator screens would add additional 0 point scoring trips.
* Remove minimum and maximum number, they are no longer used

### DIFF
--- a/html/controls/lt/lt-sheet.js
+++ b/html/controls/lt/lt-sheet.js
@@ -104,7 +104,7 @@ function prepareLtSheetTable(element, teamId, mode) {
 			var jamRow = $('<tr>').addClass('Jam').attr('nr', nr);
 			if (mode != 'copyToStatsbook') {
 				$('<td>').addClass('JamNumber Darker').text(nr).appendTo(jamRow);
-				$('<td>').addClass('NP Darker').click(function() { WS.Set(prefix+'NoPivot', $(this).text() == ""); }).appendTo(jamRow);
+				$('<td>').addClass('NP Darker').click(function() { WS.Set(prefix+'NoNamedPivot', $(this).text() == ""); }).appendTo(jamRow);
 			}
 			$.each( [ "Jammer", "Pivot", "Blocker1", "Blocker2", "Blocker3" ], function() {
 				var pos = String(this);
@@ -246,27 +246,34 @@ function prepareFieldingEditor(teamId) {
 	}		
 	
 	function processSkater(k, v) {
-		var match = (k || "").match(/Skater\(([^\)]+)\)\.Number/);
+		var match = (k || "").match(/Skater\(([^\)]+)\)\.([^\.]+)/);
 		if (match == null || match.length == 0)
 			return;
 
 		var id = match[1];
-		var option = $(".FieldingEditor #skater option[value='"+id+"']")
-		var inserted = false;
-		if (v != null && option.length == 0) {
-			var option = $('<option>').attr('value', id).text(v);
+		var field = match[2];
+		if (field != 'Role' && field != 'Number') { return; }
+		var role = WS.state['ScoreBoard.Team('+teamId+').Skater('+id+').Role'];
+		var number = WS.state['ScoreBoard.Team('+teamId+').Skater('+id+').Number'];
+		
+		var playing = (role != null && role != 'NotInGame'); 
+		
+		var option = $("#FieldingEditor #skater option[value='"+id+"']");
+		if (playing && option.length == 0) {
+			var option = $('<option>').attr('value', id).text(number);
+			var inserted = false;
 			$('#FieldingEditor #skater').children().each(function (idx, s) {
-				if (s.text > String(v) && idx > 0) {
+				if (s.text > number && idx > 0) {
 					$(s).before(option);
 					inserted = true;
 					return false;
 				}
 			});
 			if (!inserted) option.appendTo($('#FieldingEditor #skater'));
-		} else if (v == null) {
+		} else if (!playing) {
 			option.remove();
 		} else {
-			option.text(v);
+			option.text(number);
 		}
 	}
 	
--- a/html/javascript/utils.js
+++ b/html/javascript/utils.js
@@ -252,9 +252,9 @@ _crgUtils = {
 	 *		 Optional name of each child subelement value to use
 	 *		 for the option value.
 	 *		 If not set, the child node's $sbId will be used.
-	 *   optionFilterElement: string
-	 *   	Optional name of a child subelement that will trigger a
-	 *   	reevaluation of the filter when changed
+	 *	 optionFilterElement: string
+	 *		 Optional name of a child subelement that will trigger a
+	 *		 reevaluation of the filter when changed
 	 */
 	setupSelect: function(s, params) {
 		s = $(s);
--- a/html/javascript/utils.js
+++ b/html/javascript/utils.js
@@ -252,6 +252,9 @@ _crgUtils = {
 	 *		 Optional name of each child subelement value to use
 	 *		 for the option value.
 	 *		 If not set, the child node's $sbId will be used.
+	 *   optionFilterElement: string
+	 *   	Optional name of a child subelement that will trigger a
+	 *   	reevaluation of the filter when changed
 	 */
 	setupSelect: function(s, params) {
 		s = $(s);
@@ -287,6 +290,7 @@ _crgUtils = {
 		var optionChildFilter = (params.optionChildFilter || function() { return true; });
 		var optionNameElement = params.optionNameElement;
 		var optionValueElement = params.optionValueElement;
+		var optionFilterElement = params.optionFilterElement;
 
 //FIXME - need to add code to unbind if/when the option/select is removed from the DOM!
 		var setOptionName = function(option, node) {
@@ -337,8 +341,20 @@ _crgUtils = {
 
 		if (optionParent && optionChildName) {
 			_crgUtils.bindAddRemoveEach(optionParent, optionChildName, function(event, node) {
-				if (optionChildFilter(node))
+				if (optionChildFilter(node)) {
 					addOption(createOption(node));
+				}
+				if (optionFilterElement) {
+					node.$sb(optionFilterElement).$sbBindAndRun("sbchange", function(event, value) {
+						if (optionChildFilter(node)) {
+							if (!s.find("option[data-optionid='"+node.$sbPath+"']").size()) {
+								addOption(createOption(node));
+							}
+						} else {
+							removeOption(node);
+						}
+					});
+				}
 			}, function(event, node) {
 				removeOption(node);
 			});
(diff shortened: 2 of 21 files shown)


src/com/carolinarollergirls/scoreboard/core/TeamJam.java
```diff
@@ -60,7 +60,8 @@
         DISPLAY_LEAD(Boolean.class, false),
         STAR_PASS(Boolean.class, false),
         STAR_PASS_TRIP(ScoringTrip.class, null),
-        NO_PIVOT(Boolean.class, true);
+        NO_NAMED_PIVOT(Boolean.class, true),
```
brian-brazil, line 68: A comment to distinguish these would be useful.

html/controls/lt/lt-sheet.js
```diff
@@ -246,27 +246,34 @@ function prepareFieldingEditor(teamId) {
 	}		
 	
 	function processSkater(k, v) {
-		var match = (k || "").match(/Skater\(([^\)]+)\)\.Number/);
+		var match = (k || "").match(/Skater\(([^\)]+)\)\.([^\.]+)/);
```
brian-brazil, line 249: Don't forget you can use the new k.Skater etc.

html/javascript/utils.js
```diff
@@ -252,6 +252,9 @@ _crgUtils = {
 	 *		 Optional name of each child subelement value to use
 	 *		 for the option value.
 	 *		 If not set, the child node's $sbId will be used.
+	 *   optionFilterElement: string
+	 *   	Optional name of a child subelement that will trigger a
```
brian-brazil: space vs tabs here

brian-brazil (commented):
There should be unittests for the pivot behaviour. I found issues with it:

- select 4 blockers -> pivot number shows on operator
- stop jam, don't hit yellow button. Change pivot from operator => affects pivot from previous jam

src/com/carolinarollergirls/scoreboard/core/impl/TeamImpl.java
```diff
@@ -61,6 +64,17 @@ protected Object computeValue(PermanentProperty prop, Object value, Object last,
         if (prop == Value.OFFICIAL_REVIEWS && (Integer)value > scoreBoard.getRulesets().getInt(Rule.NUMBER_REVIEWS)) {
             return scoreBoard.getRulesets().getInt(Rule.NUMBER_REVIEWS);
         }
+        if (prop == Value.TRIP_SCORE && flag != Flag.COPY && scoreBoard.isInJam() && (Integer)value > 0) {
+            tripScoreTimerTask.cancel();
```
brian-brazil, line 74: Should we also clear it when a trip is added by hand?

frank-weinberg:
> select 4 blockers -> pivot number shows on operator

This would only be avoidable with significant effort. The 4th blocker is internally put in the pivot Position (following the data model of the WFTDA LT sheet) and the Dropdown is tied to the Skater value from that Position. If we were to override the value in the JS when NoPivot is set, it would send the override value to the backend as new value for the position, unsetting the 4th blocker.
Having the "No Pivot" indicator light up as the 4th Blocker is entered is IMO sufficient as a workaround. Most people will have seen an LT sheet and be able to figure out what's happening - if they pay attention to the dropdown at all in ePLT mode (and this can only happen there).

> stop jam, don't hit yellow button. Change pivot from operator => affects pivot from previous jam

Fixed. (As well as the other remarks.)

src/com/carolinarollergirls/scoreboard/core/impl/TeamImpl.java
frank-weinberg: Good idea. I also added it when a trip is removed by hand.

html/controls/lt/lt-sheet.js
frank-weinberg: I left that for when the file is rewritten to follow your approach.

brian-brazil:
> Having the "No Pivot" indicator light up as the 4th Blocker is entered is IMO sufficient as a workaround. Most people will have seen an LT sheet and be able to figure out what's happening - if they pay attention to the dropdown at all in ePLT mode (and this can only happen there).

That's insufficient, as the SO can't easily set the pivot to that skater then - plus showing No Pivot plus a pivot is confusing. 

I'm also wondering if we could avoid a new button, by having an Unknown Pivot option.

> If we were to override the value in the JS when NoPivot is set, it would send the override value to the backend as new value for the position, unsetting the 4th blocker.

It wouldn't, see my PR which was doing similar.


Here's a fun bug for you: start jam, 4 points, before the trip is advanced Score -1: Result: trip is now 7. Continues with `2x-1` pattern.

frank-weinberg:
> That's insufficient, as the SO can't easily set the pivot to that skater then

Which they wouldn't normally need to as we are in a (P)LT scenario. And in the rare case they want to, they can unset and set, which isn't exactly hard to do. (If there is really desire for this to be single click, we could make the "No Pivot" button set/unset the flag. But my guess is that there would be more instances of people accidentally clicking the button than people actually wanting to use it, which is why I didn't add that.)

> plus showing No Pivot plus a pivot is confusing.

That number isn't labelled as a pivot anywhere. And again, having "No Pivot" and the number of a blocker in the pivot box is standard for LT paperwork since forever and I can't remember people being confused by that.

> It wouldn't, see my PR which was doing similar.

Your PR was not using the crgUtils dropdown. Reimplementing all of that functionality for aesthetics in an uncommon edge case and saving a click in an extremely rare edge case counts as significant effort in my book.

And unless you think having a 4th Blocker displayed in the pivot dropdown is worse than the SBO not being able to properly enter a pivot at all, this PR still improves the situation even before looking at the other fixes and enhancements. And it doesn't keep anyone from putting in effort to add further improvements afterwards.

brian-brazil:
> That number isn't labelled as a pivot anywhere. 

The UI makes it pretty clear that it's the pivot, as otherwise it'd show No Pivot there. It's also what users would expect since star passes were added 3-4 years ago.

> And again, having "No Pivot" and the number of a blocker in the pivot box is standard for LT paperwork since forever and I can't remember people being confused by that.

SBO is not LT.

> Your PR was not using the crgUtils dropdown.

No, crgUtils wasn't powerful enough and we'll want it using WS at some point anyway.

frank-weinberg:
You are aware that you are not merging this PR because of something that it does not change at all? The 4th Blocker shows up in the dropdown since full LT paperwork is supported. (And the fact that it hasn't been brought up yet is an indicator that it isn't that big of a problem in practice.)

brian-brazil:
To supersede #266 you need to fix the bug that it's fixing, and we can't know that this PR is covering everything it needs to without that fix.

frank-weinberg:
The bug you claimed #266 is fixing is that the operator marking a Skater as Pivot actually marked it as Blocker. That is fixed here.
And #266 introduces regressions in terms of code quality (duplication of code, unnecessary ifs, ...) and functionality (If the (P)LT has marked 4 Blockers the operator setting a Pivot will overwrite one of them but the operator will not have any indication that they are doing so.)

brian-brazil:
> The bug you claimed #266 is fixing is that the operator marking a Skater as Pivot actually marked it as Blocker. That is fixed here.

And that a Blocker is not showing up in the Pivot field.

frank-weinberg:
> And that a Blocker is not showing up in the Pivot field.

That was not mentioned. And either way, your proposed fix introduced an inherent regression that is IMO worse than what you are trying to fix. So this PR does not block a solution to the issue.

And if we make "there is another issue that this PR doesn't address" a reason for not merging, we're not going to merge a lot of PRs going forward.

brian-brazil:
This proposal makes the UI more complicated, and uninutitve, adding new
unexpected semantics and a UI element that I don't believe is necessary and
is not beside what it relates to.

This was the first bit of feedback I gave on this PR, blockers showing up
in the pivot drop-down.

frank-weinberg:
Which "new unexpected semantics" are you talking about?

> a UI element that I don't believe is necessary and is not beside what it relates to

One function of the element is to indicate that at this point the "Star Pass" button is inactive because the software has been told there is no pivot. Thus (and because of available space) it is placed next to the SP button.

> This was the first bit of feedback I gave on this PR, blockers showing up in the pivot drop-down.

And it still is not something that this PR introduced. Plus, having an indicator telling that this skater is a 4th blocker and not a pivot is an improvement over current HEAD, where you can't distinguish between a Pivot and a 4th Blocker at all in the UI. Plus, there is no suggestion of how to fix this without introducing other (IMO worse) drawbacks or completely changing how 4th Blockers are handled in all of CRG.

brian-brazil:
> Which "new unexpected semantics" are you talking about?

That the Pivot dropdown no longer sets/is the pivot in all cases. Plus there's now two places with the text "No Pivot", which mean different things.

> is an improvement over current HEAD, where you can't distinguish between a Pivot and a 4th Blocker at all in the UI. Plus, 

3.9.5 actually handled this all correctly, so this is a regression.

frank-weinberg:
> That the Pivot dropdown no longer sets/is the pivot in all cases.

How many times do I have to reiterate that this behaviour is not introduced by this PR?

> 3.9.5 actually handled this all correctly, so this is a regression.

Having been in HEAD for weeks, this regression is present in all of the open PRs. Are you arguing that none of them should be merged until this has been addressed? Then please start by proposing a solution that actually improves the situation. (What you proposed so far - displaying "No PIvot" when there are 4 Blockers introduces problems that are worse than the current situation:
- If you let the operator set a Pivot, then that will overwrite the 4th Blocker, without the operator even being able to notice it. And the operator has no way of undoing this input.
- If you don't let the operator set a Pivot, the the dropdown will not work in this situation without any indication for the operator as to why this is the case.)

brian-brazil:
One thing I was thinking was to have separate No Pivot and Unknown Pivot entries in the dropdown, and we might also highlight the No Pivot as it's unusual (outside of after a star pass).

> If you let the operator set a Pivot, then that will overwrite the 4th Blocker, without the operator even being able to notice it. And the operator has no way of undoing this input.

3.9.5 would have had 4 blockers and a pivot in this situation, which isn't exactly great either.

frank-weinberg:
The two labels are not completely intuitive either: How would the interface behave in the "No Pivot" case?

I have now explicitly labeled the dropdown as Pivot or 4th Blocker - this should remove the idea that whatever number shows up there is always the Pivot. And I've set the "No Pivot" indicator to be the neighbouring button for both teams to strengthen that it is also related to this. (I tried to place it in the middle between SP and dropdown but a 3 column layout caused a line break in the dropdown cell.)

So can we get the fix for multiple operator screens in before this weekends' tournaments?

brian-brazil:
> neighbouring button for both teams to strengthen that it is also related to this.

The Star Pass should be on the same side in both cases, as we're not mirroring the UI in this part of the screen.

I can iterate from here.

frank-weinberg:
I don't see how this is better, but here you go.

brian-brazil:
It's consistency with the Lead&etc buttons above, which aren't mirrored. Maybe they should be, but let's not get into that here.

PR #273, opened 2019-05-21
Put stripes on roster page.

brian-brazil:
With a 16:9 display and many skaters, it can be tricky
to see which button goes with which skater number.


frank-weinberg (commented):
If I hadn't known the stripes were there, I probably wouldn't have seen them. They should be higher contrast to the background.

brian-brazil:
Done, I was trying to avoid it being too distracting.

PR #276, opened 2019-05-21
Make mobile screens non-zoomable, and non-selectable.

brian-brazil:
This avoids accidental zooming and selecting on touchscreens.

### DIFF
--- a/html/components/plt-input.js
+++ b/html/components/plt-input.js
@@ -734,32 +734,59 @@ function openOptionsDialog() {
 	optionsDialog.dialog('open');
 }
 
-function prepareOptionsDialog(teamId) {
+function prepareOptionsDialog(teamId, onlySettings) {
 	var table = $('<table>').appendTo($('#OptionsDialog'));
-	$('<tr>').append($('<th>').text('Select Team')).appendTo(table);
-	$.each( [ '1', '2' ], function() {
-		var tId = String(this);
-		var row = $('<tr>').addClass('selectTeam'+tId).appendTo(table);
-		$('<td>').append($('<button>').addClass('name').toggleClass('selected', tId === teamId).button().click(function() {
-			if (tId !== teamId) {
-				window.location.href = window.location.href.replace( /[\?#].*|$/, '?team='+tId );
-			} else {
-				optionsDialog.dialog('close');
-			}
-		})).appendTo(row);
-	});
 
-	WS.Register(['ScoreBoard.Team(*).Name', 'ScoreBoard.Team(*).AlternateName(operator)'], function(k, v) {
-		var displayName = 'Team ' + k.Team + ': ' + WS.state['ScoreBoard.Team('+k.Team+').Name'];
-		var altName = WS.state['ScoreBoard.Team('+k.Team+').AlternateName(operator)'];
-		if (altName != null) { displayName = displayName + ' / ' + altName; }
-		$('.selectTeam'+k.Team+' .name span').text(displayName);
+	var zoomable = $("<label/><input type='checkbox'/>").addClass("ui-button-small");
+	var id = newUUID();
+	zoomable.first().attr("for", id);
+	var zoomInput = zoomable.last().attr("id", id).button();
+	zoomInput.prop("checked", _windowFunctions.checkParam("zoomable", 1));
+	zoomable.button("option", "label", "Pinch Zoom " + (zoomInput.prop("checked")?"Enabled":"Disabled"));
+	zoomInput.change(function(e) {
+		zoomable.button("option", "label", "Pinch Zoom " + (zoomInput.prop("checked")?"Enabled":"Disabled"));
 	});
+	zoomInput.change();
+
+	if (!onlySettings) {
+		teamId = (teamId == "2" ? "2" : "1");	 // Ensure we start with a sane value.
+		$('<tr>').append($('<th>').text('Select Team')).appendTo(table);
+		$.each( [ '1', '2' ], function() {
+			var tId = String(this);
+			var row = $('<tr>').addClass('selectTeam'+tId).appendTo(table);
+			$('<td>').append($('<button>').attr('team', tId).addClass('name').toggleClass('selected', tId === teamId).button().click(function() {
+				teamId = tId;
+				table.find('[team]').removeClass("selected");
+				table.find('[team="'+tId+'"]').addClass("selected");
+			})).appendTo(row);
+		});
+		$('<tr>').append($('<th>').text('Options')).appendTo(table);
+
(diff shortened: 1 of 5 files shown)


frank-weinberg:
I can see people preferring to zoom and scroll on small devices. There should be a way to enable that. (I'm fine with no zooming being the default, though.)

brian-brazil:
This only limits scrolling horizontally, it can still scroll vertically. On my phone at least that's all this usable if you turn it the right way.

frank-weinberg:
On my phone the cells in the LT sheet are so small that zoom-tap-unzoom is preferable to trying to hit the right one multiple times.

brian-brazil:
Did you try rotating your phone horizontally? In general I'd hope that bigger devices would be used, if you have to zoom and unzoom that's going to be distracting.

frank-weinberg:
Having the screen horizontal all the time means I have to scroll all the time. Turning when I need to access the sheet means roate, wait for screen refresh, scroll to the correct line, make input, rotate and wait again. That's a lot slower than zoom, tap, unzoom, make input.
And the operators probably also hope to have bigger devices, but in practice you sometimes have no option.

brian-brazil:
How would you suggest making it enableable?

frank-weinberg:
Full on approach:
The option is set as a url parameter like the team already is. To set it there are two options: From the index screen or from the screen itself:
- Having all combinations of team and options on the index screen doesn't scale. I'd put a single entry for each screen that opens a popup in which you can set everything and then click a "Load" button to get to the actual screen.
- On the screen we can also have a popup. We could open that on clicking the team name, which will feel natural, if we include changing teams in it.

brian-brazil:
> We could open that on clicking the team name, which will feel natural, if we include changing teams in it.

That could work. Note that we can change these settings in javascript.

brian-brazil:
I've made it so this comes from a URL param.


This is about it for changes for me. I'm considering rearranging the settings page to be better organised, do you think we need to keep the swap option?

frank-weinberg:
Toggling the option does not have any effect. Even after reloading the page. Only manually adding/removing the paramter works.

frank-weinberg:
> do you think we need to keep the swap option?

I have never used it but it might be worth having a look in the FB group. I think someone asked if this was used a longer while ago.

brian-brazil:
> Toggling the option does not have any effect. Even after reloading the page. Only manually adding/removing the paramter works.

You need to select the team so that the page loads with the setting.

> I have never used it but it might be worth having a look in the FB group.

I'm not seeing anything relevant in the past 2 years. I'd need to remove another setting to get it to a row of 3 for scoreboard view settings tough, and I don't think the others are going.

frank-weinberg:
>You need to select the team so that the page loads with the setting.

That's not really intuitive. Maybe add a "Save" button?

Other issues:
- There's no way to set the option on the PT or IWB page.
- The difference between buttons that change text and text color when active and buttons that change background color when active is especially notable here. We should probably look into always changing the background color.

brian-brazil:
> That's not really intuitive. Maybe add a "Save" button?

That means cookies, and would be confusing when passed between people.

> There's no way to set the option on the PT or IWB page.

PT there's no options screen, so you'd have to go via URL. IWB isn't an input screen, so it's safe to always have zoom.

frank-weinberg:
>That means cookies, and would be confusing when passed between people.

Why? The button would just do what reselecting the previously selected team does already.

> PT there's no options screen

That can be added.

> IWB isn't an input screen, so it's safe to always have zoom.

It's currently defaulting to off.

brian-brazil:
> Why? The button would just do what reselecting the previously selected team does already.

We'd need to store state somewhere. Right now there's no state, it all comes from the URL parameters.

> It's currently defaulting to off.

I didn't put the code on the IWB page, so it should zoom as normal.

frank-weinberg:
> We'd need to store state somewhere. Right now there's no state, it all comes from the URL parameters.

And why should a new button be unable to use the info from the URL parameters?

brian-brazil:
I'm missing something here. There are no URL parameters when you open the page at first.

frank-weinberg:
>There are no URL parameters when you open the page at first.

But there are when you open the options dialog later on. And in those cases it's completely unintuitive to be forced to reselect a team.
(On first load forcing a team selection by hiding the save button should be fine, though it might be preferable to make the save button the only way out and the teams buttons only preset a selection, with a default set on first load.)

brian-brazil:
Ah, I hadn't seen that bit of the UI. I'll update.

brian-brazil:
I've gone with consistency with penalty editing, so selecting a Team is what does it for PLT where as the zoom is the only setting for PT so that makes the action. If there's more settings for PT in the future we can adjust as wanted.

frank-weinberg:
> I've gone with consistency with penalty editing

...the one piece of the PT interface that regularly people stumble over because it's not intuitive.

> If there's more settings for PT in the future we can adjust as wanted.

Why not make it scalable right away? This is knowingly introducing technical debt. 
On a related note the variable `input` should be renamed to indicate which input it's refering to.

And you didn't address the remark about inconsistent ways of highlighting enabled options.

brian-brazil:
> Why not make it scalable right away? This is knowingly introducing technical debt.

As you indicated above it's only needed on small devices like mobile phone, whereas larger tablets are going to be the vast majority of devices used for input.

> On a related note the variable input should be renamed to indicate which input it's refering to.

Done.

> And you didn't address the remark about inconsistent ways of highlighting enabled options.

Oh sorry, I forgot. I completely agree.

frank-weinberg:
>>Why not make it scalable right away? This is knowingly introducing technical debt.

>As you indicated above it's only needed on small devices like mobile phone, whereas larger tablets are going to be the vast majority of devices used for input.

I was talking about making the dialog scalable as in not requiring changes to what is there when adding further options down the road.

> Oh sorry, I forgot. I completely agree.

And your non-action is to imply that you consider this beyond the scope of this PR?

brian-brazil:
> I was talking about making the dialog scalable as in not requiring changes to what is there when adding further options down the road.

Ah, I had read scalable as zoomable. Future proof is the more common term.

Updated.

> And your non-action is to imply that you consider this beyond the scope of this PR?

I took your "We should probably look into" to mean that.

I would like to get this in, so I can get videos done, use a build at Skate Around this weekend, and we can release soon after.

frank-weinberg:
> I took your "We should probably look into" to mean that.

I was primarily seeing the other screens as out of scope here. But it's ok to change them all in one go, I guess.

brian-brazil:
> But it's ok to change them all in one go, I guess.

Yeah, looking at it briefly again this is going to be one of those changes that's pretty far reaching so it'll take a bit just to figure out how to approach this.

There's some small things it'd be nice to get in, but they can wait. Barring any bug fixes and possibly #377 this will end up as 4.0.0 all going well.

PR #277, opened 2019-05-21
Speedup loading of LT sheet.

brian-brazil:
This is about 5x faster.

### DIFF
--- a/html/controls/lt/lt-sheet.js
+++ b/html/controls/lt/lt-sheet.js
@@ -4,6 +4,10 @@ function prepareLtSheetTable(element, teamId, mode) {
 	$(initialize);
 
 	var teamName = "";
+	// Looking these up via the DOM is slow, so cache them.
+	var periodElements = {};
+	var jamElements = {};
+
 	function initialize() {
 
 		if (mode != 'plt') {
@@ -15,10 +19,30 @@ function prepareLtSheetTable(element, teamId, mode) {
 				element.find('#head').css('color', WS.state['ScoreBoard.Team(' + teamId + ').Color(operator_fg)']);
 			});
 		}
-		
-		WS.Register(['ScoreBoard.CurrentPeriodNumber'], function(k, v) { createPeriod(v); });
+
+		WS.Register(['ScoreBoard.Jam']); // for fielding editor
+		WS.Register(['ScoreBoard.Period'], handleUpdate);
+
+		if (mode == 'plt') {
+			WS.Register(['ScoreBoard.UpcomingJamNumber'], function (k, v) { element.find("#upcoming .JamNumber").text(v); });
+			WS.Register(['ScoreBoard.Team('+teamId+').NoPivot'], function(k, v) { element.find("#upcoming .NP").text(isTrue(v)?'X':''); });
+
+			WS.Register(['ScoreBoard.InJam', 'ScoreBoard.CurrentPeriodNumber'], function() {
+				for (var p in periodElements) {
+					periodElements[p].find("#upcoming").toggleClass('Hide', isTrue(WS.state['ScoreBoard.InJam']) || WS.state['ScoreBoard.CurrentPeriodNumber'] != p);
+				}
+			});
+
+			WS.Register(['ScoreBoard.Team('+teamId+').Position('], function(k, v) {
+				if (k.parts[3] == "Number") {
+					element.find("#upcoming .Skater."+k.Position).text(v);
+				} else if (k.parts[3] == "CurrentBoxSymbols") {
+					element.find("#upcoming .Box.Box"+k.Position).text(v);
+				}
+			});
+		}
 	}
-	
+
 	function teamNameUpdate() {
 		teamName = WS.state['ScoreBoard.Team(' + teamId + ').Name'];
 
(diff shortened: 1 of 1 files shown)


frank-weinberg:
This does multiple calls to WS.Register, so according to your criteria from #283 is unacceptable. Please fix.

ETA: And edit the other screens accordingly.

brian-brazil:
They're all top level, so it's fine. Having a registration in response to a WS callback is the issue.

frank-weinberg:
A registration is not more expensive if it's done in a callback handler, so that rule is arbitrary.

brian-brazil:
That's not so, at least for the very first set of registrations. If multiple things register ScoreBoard.Period before we connect to WS then each callback only has to handle that initial data once when all of that is registered at once with the backend. If that's done after the connection, then each callback has to process it again resulting in n^2 in the number of relevant registrations. We have 4 ScoreBoard.Periods for example in the operator screen, so that was the main factor in slowing initial page load.

frank-weinberg:
That sounds like it should be fixed in the WS code then instead of introducing undocumented constraints on other code that force other inefficiencies down the road.

frank-weinberg:
And looking at it, the path trie already gives you all the pieces you need: You can easily check if the requested path or a prefix of it has been registered and if so there's a function that gives you all they keys that need to be fed to the handler. I'd say the fix should be under 20 lines (but don't have access to my dev environment atm).

brian-brazil:
> That sounds like it should be fixed in the WS code then instead of introducing undocumented constraints on other code that force other inefficiencies down the road.

I tried that, it was quite a bit slower - and didn't resolve the leak issues. You still want everything to happen in a single registration at initial connect in any case, all those roundtrips take a while.

brian-brazil:
> that force other inefficiencies down the road.

If you want to argue that this approach is causing inefficiencies, please do so with benchmarks/profiles. I've already dug into and optimised this extensively, and this is about as fast as it gets - which also happens to be fast enough for our needs. (I'm still not happy that the copy page takes 1.5s to load, but I can live with that).

frank-weinberg:
> I tried that, it was quite a bit slower

Serving the updates locally was slower than doing a round trip to the backend and reiterating all the other handlers for the paths affected? Please show that benchmark in full, I'd wager a bet that there are other effects at play.

> If you want to argue that this approach is causing inefficiencies, please do so with benchmarks/profiles

I have provided as many benchmarks here as you have provided for #283. And I don't need a benchmark to know that having a subset of calls and each call having to process less filter code will be faster.

brian-brazil:
> Serving the updates locally was slower than doing a round trip to the backend and reiterating all the other handlers for the paths affected?

Yes, finding all the keys one you know you already have the data takes a while. It was several times slower than the original code.

> Please show that benchmark in full, I'd wager a bet that there are other effects at play.

I tried a few approaches in that vein, that code is all gone now as it never worked out.

> I have provided as many benchmarks here as you have provided for #283.

My stance is based on all the profiling I did in my other PRs, which tells me that dynamic registrations are a bad idea. Your point might save us a ms of CPU (doing some quick math), but it likely costs a lot more in terms of user latency. Plus the risk that others may think this is an appropriate approach.

Per my profiles, this PR makes things 5x faster following the approaches I've taken elsewhere. If you wish to argue that the approach I'm advocating (which has sped up all the other screens substantially already) has missed this as a significant potential optimisation - in direct opposition to the evidence I've seen - that requires data to back it up.

frank-weinberg:
> My stance is based on all the profiling I did in my other PRs, which tells me that dynamic registrations are a bad idea.

But those used a large number of dynamic registers, #283 uses 6 in a typical scenario. So your data isn't applicable to that PR.

brian-brazil:
My data expands beyond the overall results, as part of optimising I did many many profiles of intermediate code which gives me a good sense of what matters.

Profiling your PR quickly, my expectation is correct: Doing things this way adds ~15ms when there's no timeouts and ~50ms when there's a few. That represents a 12% slowdown for WS handling on my dummy data.

frank-weinberg:
So the overhead for reprocessing all the previously registered handlers for `Period(p).Timeout` with some timeouts present is 38ms. And looking the key set up in JS would have been several times slower? Sounds like that lookup has some serious potential for optimization.

brian-brazil:
That's more an indication that your handler could be faster, the overhead for previous handlers is negligible. There's also dead time and overheads from handling the additional WS message before even getting to the handler.

frank-weinberg:
Huh? What did you compare to to generate those numbers? Are you saying that populating the popup is slower than not doing that?

brian-brazil:
> Are you saying that populating the popup is slower than not doing that?

Basically yes, it should be faster. But on the key point, what you're doing is slowing things down unnecessarily by adding back in round trips that I spent quite some time removing.

frank-weinberg:
So you're arguing, I should do it differently, because if I don't do it at all that's faster? You're not exactly generating trust in your profiling that way.

brian-brazil:
No, I'm arguing that what you're doing is unnecessarily slow by following conventions that have already been found to cause performance problems.

frank-weinberg:
And back to my earlier point: If additional round trips are the main part of the 38ms and looking up existing paths in JS is "several times slower" then that lookup is in dire need of optimisation.

brian-brazil:
> looking up existing paths in JS is "several times slower" then that lookup is in dire need of optimisation.

We don't do that lookup, and I don't see a reason to as even with it we'd still want the top-level only pattern to avoid leaks and round trips.

Put another way, we could optimise it but it's better not to do it in the first place for various reasons.

frank-weinberg:
> to avoid leaks and round trips.

The lookups would eliminate the round trips. That's the point of doing them. And with a proper handling of Registers, you could avoid the leaks by unregistering.

And the gain of doing it this way would be having multiple small functions instead of a single big handler of everything, which improves the maintainability of the code.

>  we could optimise it but it's better not to do it in the first place for various reasons.

Of which you presented none. Leaky code can be rejected because it's leaky. We don't need to make delayed registering artificially slow in general for that.

brian-brazil:
> The lookups would eliminate the round trips. 

Only if the required data had already been requested, which is not guaranteed.

> And with a proper handling of Registers, you could avoid the leaks by unregistering.

Manual management of something like that is asking for trouble. We've enough fun keeping the right parts of the DOM in place, without adding a new form of memory management.

> And the gain of doing it this way would be having multiple small functions instead of a single big handler of everything, which improves the maintainability of the code.

I disagree here, with proper structuring it's easy to follow. Plus avoids the execution darting around various functions.

> Of which you presented none.

As indicated above it's leaky, inefficient in multiple ways (roundtrips, additional processing on initial load, dead time), and adding complexity in several places merely for the sake of allowing an additional way of doing things that's only safe and fast if you're really really careful.

> We don't need to make delayed registering artificially slow in general for that.

I'm not proposing to make delayed registering artificially slow, it's already actually slow. Making it fast wouldn't solve all the other problems it has, and would likely make it worse.

frank-weinberg:
Please read up on why long methods, especially long methods that do multiple things are bad practice. That debate has long been settled.

> Only if the required data had already been requested, which is not guaranteed.

If you do register many subelements of say a period, you could do a blank register of the period upfront and then register the small handlers as you need them.

> Manual management of something like that is asking for trouble.

If you have small handlers that only affect a single element of the DOM, you can remove them as the element is removed. Just capsule the adding and removing of elements in functions and that will be easy to maintain. If we migrate $sb or create something similar, this can pobably be mostly automated.

> Making it fast [...] would likely make it worse.

... says the guy who is blocking a PR because it's causing 50ms of extra page load time.

brian-brazil:
> says the guy who is blocking a PR because it's causing 50ms of extra page load time

I'm blocking it because it's using non top-level registrations. The 50ms I can improve whenever I'm next doing profiling in that area.

frank-weinberg:
But you couldn't give a single reason beyond those 50ms why the non top-level registrations are bad in that case. (You only listed points where they can lead to bad code if used improperly, and to top it off are refusing to discuss if/how these could be mitigated.)
So you are blocking based on an arbitrary principle. And even worse, adhering to that principle requires writing long multi purpose functions, which are generally accepted to be a bad practice.

And you are blocking #270 equally arbitrarily because it doesn't fix a behaviour that it didn't introduce and for which nobody has yet come up with a proper way of fixing it.

If that's how you want to do stuff, I'll introduce an arbitrary rule of my own: I'l not review or merge any of your PRs until those two are in. And I'll also stop wasting time debating these principles.

brian-brazil:
> But you couldn't give a single reason beyond those 50ms why the non top-level registrations are bad in that case.

I've given several clear reasons. So far this is the first case where they're used close to properly, and it's still inefficient. I'm also a bit suspicious of the amount of management code for the first/last jam, and wonder if using null/present on some attribute would be simpler as is done elsewhere.

> And even worse, adhering to that principle requires writing long multi purpose functions, which are generally accepted to be a bad practice.

That principle is far more nuanced, and not relevant here.

> So you are blocking based on an arbitrary principle

I'm blocking based on spending many many hours optimising the screens that used this pattern, which was causing significant slowdowns. After going through that I see need for this pattern (and I did try some things to get it workable). You say yourself that you'd have to do a pile of work just to solve some of the issues I've raised with it.

If we want a fast user experience and maintainable code, this pattern does not help us. The existing pattern works well though.

> And you are blocking #270 equally arbitrarily because it doesn't fix a behaviour that it didn't introduce and for which nobody has yet come up with a proper way of fixing it.

At last reply, I asked only for a minor UI tweak. I also don't think it's unreasonable to ask to fix a regression that a PR is touching major aspects of, particularly when superseding a PR that fixed the same regression.

frank-weinberg:
>particularly when superseding a PR that fixed the same regression.

As we have seen in that discussion, your PR didn't fix the issue, it merely replaced one problem with another (worse) problem.

html/controls/lt/lt-sheet.js
```diff
@@ -15,10 +19,30 @@ function prepareLtSheetTable(element, teamId, mode) {
...
 			});
 		}
-		
-		WS.Register(['ScoreBoard.CurrentPeriodNumber'], function(k, v) { createPeriod(v); });
+
+		WS.Register(['ScoreBoard.Jam']); // for fielding editor
+		WS.Register(['ScoreBoard.Period'], handleUpdate);
+
+		if (mode == 'plt') {
+			WS.Register(['ScoreBoard.UpcomingJamNumber'], function (k, v) { element.find("#upcoming .JamNumber").text(v); });
```
frank-weinberg, line 27: You did introduce quite a few DOM lookups in this part. And most of them for elements that change frequently.

brian-brazil: They're not per-skater or per-penalty, so they don't affect load times notably and everything else I profiled was fast enough not to worry.

frank-weinberg:
What did you change with that force push?

brian-brazil:
That was me resolving the conflict from a rebase.

PR #278, opened 2019-05-21
Consider all running clocks when doing clock sync.

### LINKED ISSUE #191: Clocks not synced if manually started/stopped
Maimed reported at the European Cup that the clocks got out of sync. Current theory is that this was operator error, as it corrected itself at the next jam. Thus it seems that manually starting/stopping clocks may get things out of sync. Not the end of the world (or something operators should have to do), but probably worth looking into.

brian-brazil:
This is simpler, and avoids going out of sync.

Fix precision issue in isDisplayChange.

Fixes #191

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
@@ -230,9 +230,9 @@ public class ClockImpl extends ScoreBoardEventProviderImpl implements Clock {
         //the frontend rounds values that are not full seconds to the earlier second
         //i.e. 3600ms will be displayed as 3s on a count up clock and as 4s on a count down clock.
         if (isCountDirectionDown()) {
-            return Math.floor(((float)current-1)/1000) != Math.floor(((float)last-1)/1000);
+            return Math.floor(((double)current-1)/1000) != Math.floor(((double)last-1)/1000);
         } else {
-            return Math.floor((float)current/1000) != Math.floor((float)last/1000);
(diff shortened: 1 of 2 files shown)


src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
```diff
@@ -363,18 +359,13 @@ public UpdateClockTimerTask() {
...
-                    // with respect to the master clock
-                    long nowMs = currentTime % 1000;
-                    if (masterClock != null) {
-                        nowMs = masterClock.getTime() % 1000;
-                        if (masterClock.isCountDirectionDown()) {
-                            nowMs = (1000 - nowMs) % 1000;
-                        }
+                    // with respect to the running clocks.
+                    ClockImpl masterClock = clocks.get(0);
+                    long nowMs = masterClock.getTime() % 1000;
```
frank-weinberg, line 366: This is hiding a class variable.

frank-weinberg (commented):
This seems to sporadically fail the clockSync test. (Seen it once  in 5-10 runs.)

src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
brian-brazil: Ah, forgot to remove that.

Let me dig into that test. Things were test-order dependent previously, but this change should have fixed that in passing....

brian-brazil: I've run all the tests for ~2.5 hours, with no failures.

PR #280, opened 2019-05-22
Change LT dialog checkboxes to buttons.

brian-brazil:
This is easier to hit on a tablet.
Red background, as these are unusual situations.


frank-weinberg:
Red feels wrong. The only buttons we have left in red are box trips. And they aren't exactly unusual.

brian-brazil:
Changed.

PR #281, opened 2019-05-22
Init colours for overlay teams when there's none set.

brian-brazil:


frank-weinberg:
I can't see these colors being applied anywhere. (Independent of this change.)

brian-brazil:
Look at the colour selectors in the bottom left of the overlay admin. They
were in the wrong state if there were no overlay colours set.

frank-weinberg:
I found that. But I didn't see those colors get applied to anything.

PR #283, opened 2019-05-22
Add timeout stats

frank-weinberg:


brian-brazil (commented):
If you have a timeout before Jam 1 starts, it's not reported as Jam 0 and has a duration of nan:nan.

The retained button was meant to be changed to green.

Clicking the review/timeout counts should also open the dialog box.

We should show the period clock at each timeout in the dialog, it's required for OR paperwork.

html/controls/operator.js
```diff
@@ -1089,6 +1088,212 @@ function createJamDialog() {
...
 
+function createTimeoutDialog() {
+	var lastPeriodRegistered = 0;
+	var lastPeriodListed = 0;
+	var firstJamListed = [0];
+	var lastJamListed = [0];
+	var periodDropdownTemplate = $('<select>').attr('id', 'PeriodDropdown')
+		.append($('<option>').attr('value', 0).text('P0'));
+	var jamDropdownTemplate = [$('<select>').attr('id', 'JamDropdown').attr('period', 0)
+		.append($('<option>').attr('value', 0).text('J0'))];
```
brian-brazil: This doesn't seem to be working.

html/controls/operator.js
```diff
@@ -1089,6 +1088,212 @@ function createJamDialog() {
...
+	}).appendTo(footer.find('td:eq(3)'));
+	
+	WS.Register(['ScoreBoard.CurrentPeriodNumber'], function(k, v) {
+		var p = Number(v);
+		var i;
+		for (i = lastPeriodRegistered + 1; i <= p; i++) {
+			firstJamListed[p] = 0;
+			lastJamListed[p] = 0;
+			jamDropdownTemplate[i] = $('<select>').attr('id', 'JamDropdown').attr('period', i);
+			WS.Register(['ScoreBoard.Period('+i+').FirstJamNumber'], processFirstJamNumber);
```
brian-brazil: Use one register for all of this, it's slow and leaks otherwise.

html/controls/operator.js
```diff
@@ -1089,6 +1088,212 @@ function createJamDialog() {
...
+		}
+		footer.find('#PeriodDropdown').val(p);
+	});
+
+	function processFirstJamNumber(k, v) {
+		var p = Number(k.Period);
+		var j = Number(v);
+		var i;
+		if (j == 0 && p > 0) {
+			jamDropdownTemplate[p].find('option').remove();
```
brian-brazil: Here's what's removing your J0

html/controls/operator.js
```diff
@@ -1089,6 +1088,212 @@ function createJamDialog() {
...
+			firstJamListed[p] = 0;
+			lastJamListed[p] = 0;
+			return;
+		}
+		if (firstJamListed[p] == 0 && p > 0) {
+			var last = Number(WS.state['ScoreBoard.Period('+p+').CurrentJamNumber']);
+			for (i = j; i <= last; i++) {
+				var option = $('<option>').attr('value', i).text('J'+i);
+				jamDropdownTemplate[p].append(option.clone());
+				table.find('#JamDropdown[period='+p+']').append(option);
```
brian-brazil: You're ending up with duplicates in this list somehow.

P2 is also not showing, and switching period isn't changing the listed jams.

html/controls/operator.js
```diff
@@ -1089,6 +1088,212 @@ function createJamDialog() {
...
+		}
+		for (i = lastJamListed[p]; i > j; i--) {
+			jamDropdownTemplate[p].find('option[value='+i+']').remove();
+			table.find('#JamDropdown[period='+p+'] option[value='+i+']').remove();
+			lastPeriodListed = i-1;
+		}
+	}
+
+	function processTimeout(k, v) {
+		console.log(k.field);
```
brian-brazil: Debug code

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
```diff
@@ -379,7 +369,7 @@ private void _preparePeriod() {
         jc.resetTime();
         restartPcAfterTimeout = false;
         for (ValueWithId t : getAll(Child.TEAM)) {
-            ((Team)t).resetTimeouts(false);
+            ((Team)t).recountTimeouts();
```
brian-brazil: Hmm, should this happen automatically via a changedValue?

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
```diff
@@ -152,14 +157,15 @@ public void reset() {
             removeAll(Period.NChild.JAM);
             removeAll(NChild.PERIOD);
             set(Value.CURRENT_PERIOD, getOrCreate(NChild.PERIOD, "0"));
+            noTimeoutDummy = new TimeoutImpl(getCurrentPeriod(), "noTimeout");
+            getCurrentPeriod().add(Period.Child.TIMEOUT, noTimeoutDummy);
+            set(Value.CURRENT_TIMEOUT, noTimeoutDummy);
```
brian-brazil, line 166: I'd expect this to be null, not a non-singleton placeholder.

src/com/carolinarollergirls/scoreboard/core/impl/TeamImpl.java
```diff
@@ -165,6 +184,7 @@ public void reset() {
             removeAll(Child.ALTERNATE_NAME);
             removeAll(Child.COLOR);
             removeAll(Child.SKATER);
+            removeAll(Child.TIME_OUT);
```
brian-brazil, line 199: This doesn't seem to be working, a Start New Game isn't clearing all the timeouts

html/controls/operator.js
frank-weinberg: Using one Register would require registering `ScoreBoard.Period` I'm pretty certain that's slower just because of the huge amount of events that would feed into the method.

And it's not very leaky as I'm only registering once for each period number and it's safe to assume we're not going to see a lot of different period numbers over the lifetime of the screen.

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
frank-weinberg: 'null' would require special case handling for every acess to CURRENT_TIMEOUT.
And it's as singleton as Period 0 and Jam 0, which are used analoguously.

frank-weinberg:
Everything I didnt' reply to should be fixed.

html/controls/operator.js
```diff
@@ -1089,6 +1089,214 @@ function createJamDialog() {
...
+	var headers = $("<tr><td/><td/><td/><td/><td/><td/><td/></tr>").appendTo(table);
+	$("<a>").text("Period").addClass("Title")
+		.appendTo(headers.find("td:eq(0)").addClass("Title"));
+	$("<a>").text("After Jam").addClass("Title")
+		.appendTo(headers.find("td:eq(1)").addClass("Title"));
+	$("<a>").text("Duration").addClass("Title")
+		.appendTo(headers.find("td:eq(2)").addClass("Title"));
+	$("<a>").text("Period Clock").addClass("Title")
+	.appendTo(headers.find("td:eq(3)").addClass("Title"));
+	$("<a>").text("Type").addClass("Title")
```
brian-brazil, line 1129: The retained column could do with a heading, it's a pixel wide until there's an OR currently.

html/controls/operator.js
brian-brazil: We already register all of ScoreBoard.Period, and over a tournament it could potentially be 50+ periods.

html/controls/operator.js
brian-brazil: I'm still seeing dupes when I flip between P1 and P2

html/controls/operator.js
frank-weinberg: We don't call the methods used here for all of ScoreBoard.Period.
And amongst those 50+ Periods there will be none with a number greater than 2. So we register our 3 methods once each for ScoreBoard.Period(0).X, ScoreBoard.Period(1).X, and ScoreBoard.Period(2).X That's 9 registrations. I'm certain that they will not be more costly than tens of thousands of method invocations.

brian-brazil: That's 9 registrations, per game. There can be more than one game, and each of those will then leak.

I've put quite a bit of effort into optimising all the new stats pages, and removing registrations such as this are a big part of it. Please keep to solely top-level registrations.

frank-weinberg: I DO NOT reregister for each game. I register each period NUMBER once over the lifetime of the screen. That's why I distinguish between `lastPeriodRegistered` and `lastPeriodListed`.

THe other items are in now.

brian-brazil: Do you have benchmarks indicating this makes a notable difference? I'd prefer to keep a consistent message about how to do things efficiently.

html/controls/operator.js
```diff
@@ -1148,13 +1150,14 @@ function createTimeoutDialog() {
...
 			table.find('#PeriodDropdown option[value='+i+']').remove();
-			clearPeriod(p);
+			clearPeriod(i);
 			lastPeriodListed = i-1;
 		}
 		footer.find('#PeriodDropdown').val(p);
 	});
 	
 	function addJam(p, j, append) {
+		console.log('add P'+p+'J'+j);
```
brian-brazil: You've various debug code here still

html/controls/operator.js
frank-weinberg: I don't have benchmarks. I also don't see an inconsistent message. The rule is "Only use IDs in your calls to WS.Register, if the number of different IDs encountered over the lifetime of the screen remains small." (There are tons of calls with `Team(id)` in them, including in your optimised modules. Which is fine, as we'll only ever see two team ids.)

brian-brazil: It's more nuanced than that, it matters also how many times you call it - calling it multiple times with different context is just as bad. I'd rather others not copy&paste and cause the problems I just finished fixing.

frank-weinberg: That's what we review code for (amongst other reasons).

brian-brazil: Yes, and I'm pointing out the performance issues here. Anything not top-level will slow down page load, as it will trigger a registration to the backend, and then every registered function that matches it - so that's multiple RTTs at startup and an extra few DOM paintings. Please change this, this pattern needs to be eliminated from the codebase.

frank-weinberg:
One last peace offer. This should address your performance worries for standard use.

brian-brazil:
You've a conflict.

frank-weinberg:
not anymore

brian-brazil:
It'll do for now for flat track. Among other things we've code for jam/period numbers in a few places, so should consolidate at some point.

Issue #218, opened 2019-02-24
Make it clear when <30 seconds lineup time at end of period


Twixxi691:
If there are less than 30 seconds left in the period and there will not be another jam, make this clear on the scoreboard - this could be by turning the clocks red, hiding the lineup time, or some other indicator.

AdamSmasherDerby:
I emphatically support this. (And I believe I actually suggested it a while back. :) )

PR #284, opened 2019-05-23
Indicate if there'll be another jam at the end of a period.

brian-brazil:


frank-weinberg:
When you reload the screen during lineup, the code will wrongly assume that the last Jam ended at the time of reloading. It also doesn't take into account when a Team Timeout or OR has been taken or when an overtime jam should be held. Nor does it account for different lengths of jam or lineup or the rule that there will always be another jam when the period clock runs down between jams that we added for RDCL.

Plus #218 requests a "no more jams" indicator for the main scoreboard. Instead of computing this at multiple places in the frontend, it should probably be computed once in the backend, especially given that the frontend can't reliably determine the time of the last calloff.

brian-brazil:
Speedy has other plans here.

PR #285, opened 2019-05-23
Add SP indicator+toggle to (P)LT screen

frank-weinberg:


html/controls/plt/plt-input.js
```diff
@@ -24,16 +24,18 @@ function preparePltInputTable(element, teamId, mode, statsbookPeriod, alternateN
...
 			} else {
-				$('<td>').appendTo(thead);
+				$('<td>').text('Bench').appendTo(thead);
+				$('<td>').text('Jammer').appendTo(thead);
+				$('<td>').text('Pivot').appendTo(thead);
+				$('<td>').text('Blocker').appendTo(thead);
+				$('<td>').text('Box').appendTo(thead);
 			}
+			$('<td>').attr('id', 'StarPass').text('SP').click(function() {
+				WS.Set('ScoreBoard.Team('+teamId+').StarPass', !$(this).hasClass('Active'));
```
brian-brazil, line 37: Can it be made more visually obvious that this is a button?

frank-weinberg: Do you have a suggestion as to how (especially given that none of the other buttons on this screen is specifically marked as such)?

brian-brazil: I was thinking either a different background, or putting round corners inside the td.

frank-weinberg: I've gone for a blue background as no other button on the screen has round corners.

PR #287, opened 2019-05-24
Add "No more jams" indication

### LINKED ISSUE #218: Make it clear when <30 seconds lineup time at end of period

frank-weinberg:
When there won't be another regular jam in this period without
intervention (TTO/OR), mark the headers of running clocks in red on the
operator screen and display the lineup/timeout clock with a red
background on the main view.

Closes #218 

Includes #283 as I needed those stats to reliably detect if there has been a timeout that gives another jam.

### DIFF
--- a/html/controls/operator.js
+++ b/html/controls/operator.js
@@ -942,21 +942,18 @@ function createPeriodDialog() {
 	var table = $("<table>").appendTo(dialog);
 	var headers = $("<tr><td/><td/><td/><td/><td/></tr>").appendTo(table);
 	$("<a>").text("Nr").addClass("Title")
-		.appendTo(headers.find("td:eq(0)").addClass("Title"));
+		.appendTo(headers.children("td:eq(0)").addClass("Title"));
 	$("<a>").text("Jams").addClass("Title")
-		.appendTo(headers.find("td:eq(1)").addClass("Title"));
+		.appendTo(headers.children("td:eq(1)").addClass("Title"));
 	$("<a>").text("Duration").addClass("Title")
-		.appendTo(headers.find("td:eq(2)").addClass("Title"));
+		.appendTo(headers.children("td:eq(2)").addClass("Title"));
 
-	var periodRegex = /Period\(([^\)]+)\)\.([^\.]+)/;
 	WS.Register(['ScoreBoard.Period'], function (k, v) {
-		var match = (k || "").match(periodRegex);
-		if (match == null || match.length == 0) { return; }
-		var nr = match[1];
-		if (nr == 0) { return; }
+		var nr = k.Period;
+		if (nr == null || nr == 0) { return; }
 		var prefix = "ScoreBoard.Period(" + nr + ")";
-		var key = match[2];
-		if (k != prefix + "." + key) { return; }
+		var key = k.field;
+		if (k.parts.length > 3) { return; }
 		if (!(["CurrentJamNumber", "Duration", "Number", "Running"].includes(key))) { return; }
 
 		var row = table.find("tr.Period[nr="+nr+"]");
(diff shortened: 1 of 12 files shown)


src/com/carolinarollergirls/scoreboard/rules/Rule.java
```diff
@@ -36,6 +37,8 @@
     NUMBER_REVIEWS(new IntegerRule("Team.OfficialReviews", "How many official reviews each team is granted per game or period", 1)),
     REVIEWS_PER_PERIOD(new BooleanRule("Team.OfficialReviewsPer", "Are official reviews granted per period or per game?", true, "Period", "Game")),
     NUMBER_RETAINS(new IntegerRule("Team.MaxRetains", "How many times per game or period a team can retain an official review", 1)),
+    
+    MANDATORY_OVERTIME(new BooleanRule("Overtime.Mandatory", "Will a tied game always go to overtime?", true, "True", "False")),
```
brian-brazil: This isn't used

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
```diff
@@ -82,6 +85,24 @@ protected void setupScoreBoard() {
     protected Object computeValue(PermanentProperty prop, Object value, Object last, Flag flag) {
         if (prop == Value.UPCOMING_JAM && !(value instanceof Jam)) {
             value = new JamImpl(this, getCurrentPeriod().getCurrentJam());
+        } else if (prop == Value.NO_MORE_JAM) {
```
brian-brazil, line 88: The logic isn't quite right here, if I stop on 30 it goes red rather than allowing another jam.

brian-brazil (commented):
I'll leave the review until the timeout stuff is in.

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
frank-weinberg: There are exactly 30s between the moment when the display flips to 30 and the moment when the display flips to 0. So if you stop when the 30 is already showing, there are only 29.x seconds remaining and there should not be another Jam.

But I've changed the code to give you another Jam when you hit exactly 30.0s.

brian-brazil:
Can you rebase this off dev?

frank-weinberg:
Done.

html/controls/operator.js
```diff
@@ -1078,6 +1084,13 @@ function createJamDialog() {
 					row.find("td.Duration").text(_timeConversions.msToMinSec(v));
 				}
 			}
+			if (key == 'PeriodClockElapsedEnd') {
+				if (WS.state[prefix + '.WalltimeEnd'] == 0 && WS.state[prefix + '.WalltimeStart'] > 0) {
+					row.find("td.PC").text("running");
+				} else {
+					row.find("td.PC").text(_timeConversions.msToMinSec(v));
```
brian-brazil: This doesn't seem to be working, it's always empty.

html/controls/operator.js
```diff
@@ -1006,13 +1009,15 @@ function createPeriodDialog() {
...
-	var headers = $("<tr><td/><td/><td/><td/><td/></tr>").appendTo(tableTemplate);
+	var headers = $("<tr><td/><td/><td/><td/><td/><td/></tr>").appendTo(tableTemplate);
 	$("<a>").text("Nr").addClass("Title")
 		.appendTo(headers.find("td:eq(0)").addClass("Title"));
 	$("<a>").text("Points").addClass("Title")
 		.appendTo(headers.find("td:eq(1)").addClass("Title"));
 	$("<a>").text("Duration").addClass("Title")
 		.appendTo(headers.find("td:eq(2)").addClass("Title"));
+	$("<a>").text("PC at end").addClass("Title")
+	.appendTo(headers.find("td:eq(3)").addClass("Title"));
```
brian-brazil: prefer children to find when you can, it's faster.

brian-brazil (commented):
The new logic should have a unittest, covering OTO+TO cases preferably too.

html/controls/operator.js
frank-weinberg: While I was at it I did some other straightforward optimizations as well, like using k.Period etc.

html/controls/operator.js
brian-brazil: This is showing elapsed time, we probably want whatever the operator is seeing in terms of inversion.

frank-weinberg: True. Changed. (Also unit tests in.)

brian-brazil:
Hmm, points isn't showing on the Jams popup but I presume that's due to something else.

frank-weinberg:
Ah, damn, that was one `children` too much.

brian-brazil:
Changing those is always fun ;)

PR #288, opened 2019-05-24
Fix autoselecting new teams

frank-weinberg:


brian-brazil:
This needs more work (it's racy), I intend to use the #282 approach once that's in.

PR #291, opened 2019-05-25
Ignore accidental double clicks on Jam/TO controls

frank-weinberg:
They mess up undo and confuse people.
This also prevents the lineup clock from starting when the operator
tries to end the last jam after the software already did so.

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
@@ -280,6 +280,7 @@ public class ScoreBoardImpl extends ScoreBoardEventProviderImpl implements Score
             Clock jc = getClock(Clock.ID_JAM);
             Clock lc = getClock(Clock.ID_LINEUP);
             Clock tc = getClock(Clock.ID_TIMEOUT);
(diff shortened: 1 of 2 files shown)


src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
```diff
@@ -291,7 +292,10 @@ public void stopJamTO() {
                 setLabels(ACTION_START_JAM, ACTION_NONE, ACTION_TIMEOUT);
                 _endTimeout(false);
                 finishReplace();
-            } else if (!lc.isRunning()) {
+            } else if (!lc.isRunning() && (!ic.isRunning() || ic.getTimeElapsed() > 1000L)) {
```
brian-brazil, line 295: I think you meant `<`

brian-brazil (commented):
Unittests would be good

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
frank-weinberg: No. This is the condition under which we process the input.

Unit tests that show it works as intended are in.

PR #292, opened 2019-05-26
Don't show Lead after halftime

frank-weinberg:
When the intermission clock has run down and the new period hasn't
started, current HEAD shows "Lead" for the team that had it in the last
Jam of the previous period.


brian-brazil:
That'd explain a report I had, but they only mentioned between jams.

PR #294, opened 2019-05-29
Add option to assume penatlies are served

brian-brazil:
Useful when you've a PT, but no LT

Which can never be null, simplify code


html/controls/plt/plt-input.js
```diff
@@ -481,6 +480,8 @@ function preparePenaltyEditor() {
 		WS.Set(prefix + '.Code', $('#PenaltyEditor .Codes .Active').attr('code'));
 		if ($('#PenaltyEditor #served').prop('checked') != wasServed) {
 			WS.Set(prefix + '.ForceServed', !wasServed);
+		} else if (isTrue(WS.state['ScoreBoard.Settings.Setting(ScoreBoard.Penalties.ForceServed)'])) {
+			WS.Set(prefix + '.ForceServed', wasServed);
```
frank-weinberg, line 484: This should set to true.

brian-brazil:
You might still want to track something that wasn't served, such as across
jams. This allows the forced to be unticked manually still.

PR #295, opened 2019-05-29
Add rosters to the copy page.

brian-brazil:
Use unicode strikethrough so it'll copy properly.

### DIFF
--- a/html/views/roster-sheet.js
+++ b/html/views/roster-sheet.js
@@ -1,0 +1,72 @@
+function prepareRosterSheetTable(element, teamId, mode, statsbookPeriod) {
+	
+	/* Values supported for mode:
+	 * copyToStatsbook: Only roster cells for IGRF.
+	 */
+	
+	'use strict';
+	$(initialize);
+
+	var alternateName = "operator";
+	var tbody;
+
+	function initialize() {
+
+		var table = $('<table cellpadding="0" cellspacing="0" border="1">').addClass('Roster Team').addClass("AlternateName_" + alternateName).appendTo(element);
+		var thead = $('<thead>').appendTo(table);
+		$('<tr>').appendTo(thead)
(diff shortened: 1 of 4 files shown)


html/views/roster-sheet.js
```diff
@@ -0,0 +1,87 @@
...
+		}
+		row.attr("flags", WS.state[prefix + ".Flags"]);
+		var name = WS.state[prefix + ".Name"];
+		var number = String(WS.state[prefix + ".Number"]);
+		// Use Unicode's Combining Long Stroke Overlay so strikethrough
+		// will copy to the statsbook.
+		if (row.attr("flags") == "ALT" || row.attr("flags") == "BC") {
+			var n = "";
+			for (var i = 0; i<name.length; i++) {
+				n += name[i] + "\u0336";
```
frank-weinberg: This could break automated processing of the statsbook (if the roster names/numbers are compared against other documents/databases). I'd prefer a strikethrough that doesn't copy to one that edits the string itself.

html/views/roster-sheet.js
```diff
@@ -0,0 +1,87 @@
...
+			}
+			name = n;
+			n = "";
+			for (var i = 0; i<number.length; i++) {
+				n = n + number[i] + "\u0336";
+			}
+			number = n;
```
frank-weinberg: There needs to be an asterisk appended to the number (See statsbook manual page 29).

brian-brazil: But not on page 7 where I was looking :(

brian-brazil:
Okay, I thought they'd changed the statsbook. This is simpler.

html/views/roster-sheet.js
frank-weinberg: That's the In-Game procedures. For this screen we should be looking at the postgame part. (The same mistake caused me to report a not-a-bug in the statsbook checker recently.)

PR #297, opened 2019-05-29
Fix period start key name.

brian-brazil:
This gets us back to one file per game, rather than
one per jam.


frank-weinberg:
Didn't we say we want to have one file per period (with each containing the game data of previous periods as well)?

brian-brazil:
So we did, updated.

PR #299, opened 2019-05-30
It turns out there's already a way to clear a key, so expose it.

brian-brazil:
I was going to add this, but turns out it's just not discoverable.


frank-weinberg:
I did find this at some point.

PR #301, opened 2019-05-31
Attach new Jams to the upcoming period after half of intermission

frank-weinberg:
This is more natural than tipping over 1 minute before intermission is over.
(I wanted to go to the middle already when I moved to the 1 minute spot but determining the middle of intermission was not reliably possible with the code back then.)


brian-brazil:
I'd expect a unittest to get updated with this.

Isn't this the old behaviour?

frank-weinberg:
> I'd expect a unittest to get updated with this.

The ones in the last minute obviously still work and the opposite tests are all within the first half of intermission (which is where they also happen in practice).

> Isn't this the old behaviour?

old old behaviour (pre 3.9.3) was to only tip over when the intermission had completely run down (leading to problems when JTs tried to start on the spot and the Jam was started a tad before it ticked over).

PR #304, opened 2019-05-31
Make Start New Game the only way to initilise current teams.

brian-brazil:
The assign logic locks things up. If anyone wants to copy back changes,
the new copy&paste can be used with the rosters on the copy to statsbook page.

This also gets us closer to moving this to ScoreBoard.


frank-weinberg:
copy&paste does not copy colors or alternate names.

brian-brazil:
No, but copying those back to the main roster is probably not what you want
to do anyway as they change game to game. In any case, this feature
currently breaks the scoreboard.

brian-brazil:
Let's do this in the other PR.

PR #309, opened 2019-06-07
Fix and reorganize overlay code.

brian-brazil:
Correctly handle skaters changing number, skaters
with no name, skaters with no number, ignore foexp (for now),
count penalties correctly when colouring, handle null logo.

Break down big functions into smaller more understandable parts,
and use the new WS functionality to simplify.

There's more to be done, but this is a big step in the right direction.

### DIFF
--- a/html/views/overlays/interactive/admin.js
+++ b/html/views/overlays/interactive/admin.js
@@ -22,9 +22,6 @@ $(initialize)
 
 function initialize() {
 
-	WS.Connect();
-	WS.AutoRegister();
-
 	WS.Register(['ScoreBoard.Settings.Setting(Overlay.Interactive.Clock)', 
 		     'ScoreBoard.Settings.Setting(Overlay.Interactive.Score)', 
 		     'ScoreBoard.Settings.Setting(Overlay.Interactive.ShowJammers)',
@@ -82,6 +79,8 @@ function initialize() {
 
 	WS.Register('ScoreBoard.Settings.Setting(Overlay.Interactive.LowerThird.Style)', function(k,v) { $('#LowerThirdStyle option[value="'+v+'"]').attr('selected', 'selected'); });
 
+	WS.Connect();
+	WS.AutoRegister();
 }
 
 $('#Controls input, #Controls .Selector').change(function() {
@@ -90,9 +89,9 @@ $('#Controls input, #Controls .Selector').change(function() {
 	if ($(this).attr("type") == "color") {
 		$(this).attr("cleared", "false");
 	}
-	if (v == '' && t.endsWith('.Name')) {
+	if (v == '' && t.endsWith('AlternateName(overlay)')) {
 		// Delete the AlternateName
-		WS.Set(t.substr(0, t.length-5), null);
+		WS.Set(t, null);
 	} else if(t) {
 		WS.Set(t, v);
 	}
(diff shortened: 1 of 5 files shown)


html/views/overlays/interactive/admin.js
```diff
@@ -133,7 +132,7 @@ $('select#Skaters').change(function(e) {
 	team = $( 'option[value=' + v + ']', $t ).attr('data-team');
 	name = $( 'option[value=' + v + ']', $t ).attr('data-name');
 	tnam = WS.state['ScoreBoard.Team(' + team + ').AlternateName(overlay)'];
-	tnam = tnam ? tnam : WS.state['ScoreBoard.Team(' + team + ').Name'];
+	tnam = (tnam != "") ? tnam : WS.state['ScoreBoard.Team(' + team + ').Name'];
```
frank-weinberg: Isn't this `null` if the AlternateName isn't set?

html/views/overlays/interactive/index.js
```diff
@@ -1,38 +1,101 @@
...
+	WS.Register( ['ScoreBoard.Period(*).Jam(*).TeamJam(*).JamScore',
+			'ScoreBoard.Period(*).Jam(*).TeamJam(*).Lead',
+			'ScoreBoard.Period(*).Jam(*).TeamJam(*).Lost'], function(k,v) { jamData(k,v); } );
+
+	WS.Register( [
+			'ScoreBoard.Team(*).Skater(*).Name',
+			'ScoreBoard.Team(*).Skater(*).Number',
+			'ScoreBoard.Team(*).Skater(*).Flags' ], function(k,v) {
+				var me = '.RosterTeam' + k.Team + ' .Team' + k.Team + ' .Skater[data-skaterId=' + k.Skater + ']';
+				var mb = '.PenaltyTeam' + k.Team + ' .Team' + k.Team + ' .Skater[data-skaterId=' + k.Skater + ']';
```
frank-weinberg, line 33: What does `me`or `mb` mean? Can't we use more readable variable names? (This applies to much of the code here.)

html/views/overlays/interactive/index.js
```diff
@@ -1,38 +1,101 @@
...
+		if (v == null) {
+			$('.Number-'+k.Penalty, sel).remove();
+			$(sel).attr('data-count', $('.Penalty', sel).length);
+			return;
+		}
+		ensureSkaterExists(k.Skater, k.Team);
+		createPenalty(sel, k.Penalty, v);
+	} );
+
+	WS.Register( 'ScoreBoard.Team(*).Color(*)', function(k,v) {
```
frank-weinberg: The function only acts on `overlay_bg` and `overlay_fg` - wouldn't it make more sense to only register those keys.

html/views/overlays/interactive/admin.js
brian-brazil: Hmm, thought I'd already undone that one.

html/views/overlays/interactive/index.js
brian-brazil: This is all pre-existing, there's more changes to fix this up and I felt the PR was already big enough.

html/views/overlays/interactive/index.js
brian-brazil: Done

html/views/overlays/interactive/index.js
```diff
@@ -1,38 +1,102 @@
...
+			var ns,r;
+			var rule;
+			// chrome seems to like things in lowercase
+			var rd = '#sb .colourteam'+k.Team;
 
-	WS.Register( 'ScoreBoard.Team(1)', function(k,v) { teamData(1, k,v); } );
-	WS.Register( 'ScoreBoard.Team(2)', function(k,v) { teamData(2, k,v); } );
+			if(k.Color == 'overlay_bg') ns = 'background-color';
+			if(k.Color == 'overlay_fg') ns = 'color';
+			if(ns) {
```
frank-weinberg: This condition is now always true.

brian-brazil: Fixed

PR #313, opened 2019-06-10
Restructure SkaterImpl constructor

frank-weinberg:
They now follow the scheme of the other elments. (Previous structure was
differing in order to not break compatibility with XML team assignment which is now gone.)


brian-brazil:
That's cleaner alright.

Issue #4, opened 2017-05-07
Performance issue with large numbers of teams


jaredquinn:
When large numbers of teams are in use for a tournament performance issues have been noticed when utilising the Team screen; these may affect other viewers of the scoreboard and cause delays in clients updating.

Investigate improving performance of team data.

Bud-Herz:
Suggestion - Could a short term fix be file upload of team that overwrites any existing when a certain capacity is reached? 

Just spitballin'

official-sounding:
I did a bit of profiling on this, and it appears that the slowdown is happening because of two problems:

1. A Team's roster table takes about 500ms to generate
2. Every Team table gets generated when the page first loads

Problem 2 is pretty easy to resolve, I am finishing up testing a pull request that will defer generation of the table until the team is selected from the dropdown for the first time.  that will keep the load time of the team page very stable at about 600 ms, with a small pause the first time you select a team.

Problem 1 is happening mostly because a team roster table is very dense with $sbControl() calls -- building up an input that is backed by the XML model is quite expensive, and there are a lot of model-backed inputs on the page.  Specifically, a lot of time is taken determining the element type -- over 100ms is spent just on calls to jQuery's .is() function.  A braver soul than I would need to take that up, but it may be moot with fixing problem 2.

official-sounding:
@brian-brazil  this one is also done

brian-brazil:
My read of it is that it's mitigated, but the underlying excessive calls to sbControl are still there. 500ms is fairly slow to load a team.

PR #317, opened 2019-06-12
Move teams to WS

brian-brazil:
I also fixed some issues on the media page that I noticed in passing.


brian-brazil:
I'm seeing <150ms time to switch teams even with a "busy" scoreboard, so this also resolves #4.

PR #318, opened 2019-06-13
Match full components of WS paths

frank-weinberg:
This avoids getting e.g. updates for StarPassTrip when StarPass was
registered.
As a side effect it significantly reduces the number of trie lookups,
which should give a slight performance increase.

### DIFF
--- a/html/json/WS.js
+++ b/html/json/WS.js
@@ -278,8 +278,9 @@ var WS = {
 	},
 
 	_addToTrie: function(t, key, value) {
-		for (var i = 0; i < key.length; i++) {
-			var c = key.charAt(i);
+		var p = key.split(/[.(]/);
+		for (var i = 0; i < p.length; i++) {
+			var c = p[i];
 			t[c] = t[c] || {};
 			t = t[c];
 		}
@@ -288,20 +289,25 @@ var WS = {
 	},
 
 	_getMatchesFromTrie: function(t, key) {
-		var result = t.values || [];
-		for (var i = 0; i < key.length; i++) {
-			var c = key.charAt(i);
-			t = t[c];
-			if (t == null) {
-				break;
+		function matches(t, p, i) {
+			var result = t.values || [];
+			for (; i < p.length; i++) {
+				if (t["*)"] != null) {
+					// Allow Blah(*) as a wildcard.
+					var j;
+					for (j = i; j < p.length && !p[j].endsWith(")"); j++);
+					result = result.concat(matches(t["*)"], p, j+1) || []);
+				}
+				t = t[p[i]];
+				if (t == null) {
+					break;
+				}
+				result = result.concat(t.values || []);
 			}
-			if (c == "(" && t["*"] != null) {
-				// Allow Blah(*) as a wildcard.
-				result = result.concat(WS._getMatchesFromTrie(t["*"], key.substring(key.indexOf(")", i))) || []);
-			}
-			result = result.concat(t.values || []);
-		}
-		return result;
+			return result;
+		};
+		
+		return matches(t, key.split(/[.(]/), 0);
 	},
 
 	getPaths: function(elem, attr) {
--- a/html/json/WS.js
+++ b/html/json/WS.js
@@ -295,6 +295,7 @@ var WS = {
 				if (t["*)"] != null) {
 					// Allow Blah(*) as a wildcard.
 					var j;
+					// id captured by * might contain . and thus be split - find the end
 					for (j = i; j < p.length && !p[j].endsWith(")"); j++);
 					result = result.concat(matches(t["*)"], p, j+1) || []);
 				}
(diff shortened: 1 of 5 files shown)


html/json/WS.js
```diff
@@ -278,8 +278,9 @@ var WS = {
 	},
 
 	_addToTrie: function(t, key, value) {
-		for (var i = 0; i < key.length; i++) {
-			var c = key.charAt(i);
+		var p = key.split(/[.(]/);
```
brian-brazil, line 281: What if there's a . inside an id, which does happen for media?

brian-brazil (commented):
I'm pretty sure this will break something somewhere.

I think we should have a general glob so the old behaviour can be obtains, I'd suggest `**` as a suffix.

html/json/WS.js
frank-weinberg: You can still match the whole id. As a side effect you can also add a * after any . in the id and it will match anything after that.
See the new unit test for an example.

frank-weinberg:
Can you give any scenario where ** would be a significant improvement?
And the way we are currently using WS paths, there shouldn't be any breakage. If there is, it's most likely something that should be fixed anyway.

tests/com/carolinarollergirls/scoreboard/jetty/WSTest.java
```diff
@@ -73,11 +73,25 @@ public void path_trie_non_id_glob() {
...
+        pt.add("ScoreBoard.Rulesets.Rule(Jam.*)");
+        pt.add("ScoreBoard.Rulesets.Rule(Intermission*)");
+        
+        assertTrue(pt.covers("ScoreBoard.Rulesets.Rule(Period.Duration)"));
+        assertTrue(pt.covers("ScoreBoard.Rulesets.Rule(Jam.Foo)"));
+        assertTrue(pt.covers("ScoreBoard.Rulesets.Rule(Jam.Foo.Bar)"));
+        
+        assertFalse(pt.covers("ScoreBoard.Rulesets.Rule(Period.Direction)"));
+        assertFalse(pt.covers("ScoreBoard.Rulesets.Rule(Jam)"));
+        assertFalse(pt.covers("ScoreBoard.Rulesets.Rule(Intermission.Direction)"));
```
brian-brazil, line 95: Would `Rule(*)` cover this?

brian-brazil (commented):
> Can you give any scenario where ** would be a significant improvement?

It's useful for development/exploration to watch/dump an entire prefix.

> If there is, it's most likely something that should be fixed anyway.

I'm pretty certain there's cases in the JS somewhere.

tests/com/carolinarollergirls/scoreboard/jetty/WSTest.java
frank-weinberg: Yes, of course.

brian-brazil: Can you add a test for that?

frank-weinberg:
> > Can you give any scenario where ** would be a significant improvement?
> 
> It's useful for development/exploration to watch/dump an entire prefix.

You can still get `ScoreBoard` or `ScoreBoard.Period` (== `ScoreBoard.Period(*)`). The only thing you can't get anymore would be things like `ScoreBoard.P` to get Periods and PenaltyCodes at once but nothing else. But that doesn't seem to be particularly useful.

> I'm pretty certain there's cases in the JS somewhere.

Ok, I went through all 86 WS.Register calls and indeed there were two instances in the Overlay code, which you wanted to revisit anyways.

tests/com/carolinarollergirls/scoreboard/jetty/WSTest.java
frank-weinberg: Done

src/com/carolinarollergirls/scoreboard/jetty/WS.java
```diff
@@ -289,56 +289,53 @@ private void sendWSUpdatesForPaths(PathTrie watchedPaths, Set<String> changed) {
...
+                if (head.exists) {
+                    return true;
+                }
+                if (i >= p.length) {
+                    return false;
+                }
+                // Allow Blah(*).
+                if (head.trie.containsKey("*)")) {
+                    int j;
+                    for (j = i; j < p.length && !p[j].endsWith(")"); j++);
```
brian-brazil, line 330: A comment or empty {} would be useful here, it's unusual to use a for like this.

html/json/WS.js
```diff
@@ -288,20 +289,25 @@ var WS = {
...
-			if (c == "(" && t["*"] != null) {
-				// Allow Blah(*) as a wildcard.
-				result = result.concat(WS._getMatchesFromTrie(t["*"], key.substring(key.indexOf(")", i))) || []);
+		function matches(t, p, i) {
+			var result = t.values || [];
+			for (; i < p.length; i++) {
+				if (t["*)"] != null) {
+					// Allow Blah(*) as a wildcard.
+					var j;
+					for (j = i; j < p.length && !p[j].endsWith(")"); j++);
```
brian-brazil, line 299: Same here.

brian-brazil (commented):
Ah, I had thought you were going for full match rather than per-part match.

src/com/carolinarollergirls/scoreboard/jetty/WS.java
frank-weinberg: Comment added

PR #319, opened 2019-06-14
Switch ScoreBoard View to WS

brian-brazil:
I've extracted some utility functions, in preparation
for doing the main operator tab.


frank-weinberg:
Those utility functions look like they should be useful in other screens as well. We should probably factor them into a separate file (or add to an existing lib).

brian-brazil:
Some of them, others are less useful than you'd think. One step at a time though.

PR #329, opened 2019-06-16
Make row colours on SK/LT match the paperwork.

### LINKED ISSUE #325: SK/LT line colours don't match paperwork
Both show two lines the same when there's a star pass - which doesn't follow paperwork.

Also SK is counting lines from the top rather than the bottom, so the top line is always white rather than the bottom line of a period always being white.

brian-brazil:
Indicate which rows are related by starpasses by using
border styles.

As a side benefit, this approach means the Copy to Statsbook page now
works for star passes if you've saved a copy without the CSS.

Fixes #325

### DIFF
--- a/html/controls/lt/lt-sheet.css
+++ b/html/controls/lt/lt-sheet.css
@@ -22,9 +22,13 @@ div.LT { font-size: 200%; text-align: center; background-color: #cccccc; color: 
 .LT.Period thead td { font-size: 95%; background-color: #cccccc; color: #333333; }
 .LT.Period td { text-align: center; vertical-align: middle; }
 .LT.Period tbody tr { background-color: #FFF; }
-.LT.Period tbody tr:nth-child(4n+0), .LT.Period tr:nth-child(4n-1) { background-color: #e7f3ff; }
+.LT.Period.Forewards tbody tr:nth-child(even) { background-color: #e7f3ff; }
+.LT.Period.Backwards tbody tr:nth-last-child(even) { background-color: #e7f3ff; }
 .LT.Period tbody tr .Darker { background-color: #e7f3ff; }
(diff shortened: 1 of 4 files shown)


html/controls/lt/lt-sheet.css
```diff
@@ -22,9 +22,13 @@ div.LT { font-size: 200%; text-align: center; background-color: #cccccc; color:
 .LT.Period thead td { font-size: 95%; background-color: #cccccc; color: #333333; }
 .LT.Period td { text-align: center; vertical-align: middle; }
 .LT.Period tbody tr { background-color: #FFF; }
-.LT.Period tbody tr:nth-child(4n+0), .LT.Period tr:nth-child(4n-1) { background-color: #e7f3ff; }
+.LT.Period.Forewards tbody tr:nth-child(even) { background-color: #e7f3ff; }
+.LT.Period.Backwards tbody tr:nth-last-child(even) { background-color: #e7f3ff; }
 .LT.Period tbody tr .Darker { background-color: #e7f3ff; }
-.LT.Period tbody tr:nth-child(4n+0) .Darker, .LT.Period tr:nth-child(4n-1) .Darker { background-color: #cee7ff; }
+.LT.Period.Forewards tbody tr:nth-child(odd) .Darker { background-color: #cee7ff; }
+.LT.Period.Backwards tbody tr:nth-last-child(odd) .Darker { background-color: #cee7ff; }
```
frank-weinberg: These 2 should be `even`, not `odd`.

brian-brazil: It looks right to me, the first line is white.

frank-weinberg: But the first two columns are inverted.

brian-brazil: I see what you mean now, fixed.

PR #330, opened 2019-06-17
Switch main operator tab to WS.

brian-brazil:
Cleanup dead code.

This roughly halves the load time for a new scoreboard, we're now sub-second.


html/controls/operator.js
```diff
@@ -256,36 +245,26 @@ function createMetaControlTable() {
...
-				return value;
-			}
-		}, sbcontrol: {
-			button: true
-		} })
-		.appendTo(periodEndTd);
+	WS.Register(["ScoreBoard.Clock(Period).Time",
+			"ScoreBoard.Clock(Period).Number",
+			"ScoreBoard.Rulesets.CurrentRule(Period.Number)"], function(k, v) {
+				var under30 = (Number(WS.state["ScoreBoard.Clock(Period).Time"]) < 30000);
```
frank-weinberg, line 251: This could use `ScoreBoard.NoMoreJam`.

html/controls/operator.js
```diff
@@ -414,18 +398,25 @@ function createPeriodEndTimeoutDialog(td) {
...
-		if (currentSecs < targetSecs)
-			$sb("ScoreBoard.Clock(Period).Time").$sbSet(_timeConversions.secondsToMs(targetSecs));
-		$sb("ScoreBoard.Clock(Intermission).Stop").$sbSet("true");
-		$(this).unbind(event);
+		}
+		if (currentSecs < targetSecs) {
+			WS.Set("ScoreBoard.Clock(Period).Time", _timeConversions.secondsToMs(targetSecs));
+		}
+		WS.Set("ScoreBoard.Timeout", true);
+		WS.Set("ScoreBoard.Clock(Intermission).Stop", true);
```
frank-weinberg: This is already done in the backend when starting a timeout.

html/controls/operator.js
```diff
@@ -699,10 +714,11 @@ function createTeamTable() {
...
 
 		var scoreChangeTimeout;
-		sbTeam.$sb("TripScore").bind("sbchange", function(event,s) {
-			var c = (s<0 ? "#800" : s>0 ? "#080" : "#008");
-			var addTrip = $sb("ScoreBoard.InJam").$sbIsTrue();
-			scoreChange.stop(true).text(s).last().css({ opacity: "1", color: c });
+		WS.Register("ScoreBoard.InJam");
+		WS.Register(prefix + ".TripScore", function(k, v) {
+			var c = (v>0 ? "#080" : "#008");
+			var addTrip = isTrue(WS.state["ScoreBoard.InJam"]);
```
frank-weinberg: This variable is not used. (It's left over from when the extra trip was added in the frontend.)

html/controls/operator.js
```diff
@@ -699,10 +714,11 @@ function createTeamTable() {
 		var jamScore = $("<a>").appendTo(scoreSubTr.children("td:eq("+(first?"0":"2")+")")).addClass("JamScore");
 
 		var scoreChangeTimeout;
-		sbTeam.$sb("TripScore").bind("sbchange", function(event,s) {
-			var c = (s<0 ? "#800" : s>0 ? "#080" : "#008");
-			var addTrip = $sb("ScoreBoard.InJam").$sbIsTrue();
-			scoreChange.stop(true).text(s).last().css({ opacity: "1", color: c });
+		WS.Register("ScoreBoard.InJam");
```
frank-weinberg: This value is only used to set the unused variable below.

html/controls/operator.js
```diff
@@ -726,154 +742,91 @@ function createTeamTable() {
...
-					firstOption: { text: '?', value: "" }
-				} }).addClass(pos+" By"+elem+" "+sort+"Sort");
+			var select = $("<select>").append($("<option value=''>?</option>"));
+			WS.Register(prefix + ".Skater(*).Number", function(k, v) {
+				select.children("[value='"+k.Skater+"']").remove();
+				if (v != null) {
+					var option = $("<option>").attr("number", v).val(k.Skater).text(v);
+					_windowFunctions.appendAlphaSortedByAttr(select, option, "number", 1);
+					select.val(WS.state[prefix + ".Position("+pos+").Skater"]);
+				}
```
frank-weinberg, line 809: These dropdowns don't filter out the not skating skaters.

html/controls/operator.js
```diff
@@ -1235,92 +1201,92 @@ function createTimeoutDialog() {
...
-				.append($('<td>').addClass('PerClock').text(pc))
-				.append($('<td>').addClass('Type').append(typeDropdownTemplate.clone().val(type).change(function() {
-					var parts = $(this).val().split('.');
-					WS.Set(prefix+'.Owner', parts[0]);
-					WS.Set(prefix+'.Review', isTrue(parts[1]));
-				})))
-				.append($('<td>').addClass('Retained').append($('<button>').toggleClass('Hide', !review).toggleClass('Active', retained)
-						.text('Retained').button().click(function () {
-							WS.Set(prefix+'.RetainedReview', !isTrue(WS.state[prefix+'.RetainedReview']));
+				var row = table.find('tr.Timeout[toId='+id+']');
```
frank-weinberg: Why does the indentation change here?

brian-brazil: I reindented the entire file, so anything that was inconsistent would also have changed.

html/controls/operator.js
brian-brazil: Not quite, it won't flash once the period ends

html/controls/operator.js
brian-brazil: Done

html/controls/operator.js
brian-brazil: Done

html/controls/operator.js
brian-brazil: Done

html/controls/operator.js
brian-brazil: Done

html/controls/operator.js
frank-weinberg: But this line should be at the same level as the line above.

html/controls/operator.js
```diff
@@ -726,154 +739,92 @@ function createTeamTable() {
...
-					optionNameElement: elem,
-					optionFilterElement: "Role",
-					optionChildFilter: function(node) { return (node.$sb("Role").$sbGet() != 'NotInGame'); },
-					compareOptions: function(a, b) { return sortFunc("text", a, b); },
-					firstOption: { text: '?', value: "" }
-				} }).addClass(pos+" By"+elem+" "+sort+"Sort");
+			var select = $("<select>").append($("<option value=''>?</option>"));
+			WS.Register([prefix + ".Skater(*).Number", prefix + ".Skater(*).BaseRole"], function(k, v) {
+				select.children("[value='"+k.Skater+"']").remove();
+				if (v != null && WS.state[prefix + ".Skater("+k.Skater+").BaseRole"] == "Bench") {
```
frank-weinberg: This will remove and thus unset a skater as they foul out. It is also inconsistent with all the other screens. Use `.Role != 'NotInGame'`.

html/controls/operator.js
brian-brazil: Sometimes it does this, I'm not sure why. Worked around it.

html/controls/operator.js
brian-brazil: Done, plus fixed a small bug on the Teams tab.

frank-weinberg: It should be `.Role`, not `.BaseRole`. If a Skater is marked as being in the current lineup when the Not Skating flag is set, we give precedence to the former everywhere else.

brian-brazil: Done

PR #332, opened 2019-06-19
Add program version number to state

### LINKED ISSUE #239: Add program version number to state
In order to make things easier on API users in the future, it will be helpful if there's a clear way to determine which version of the scoreboard was used both from the output file and during the game.  This might actually warrant a top-level variable "ScoreBoard.Version:", but if not, it could probably get thrown in with "ruleset".

frank-weinberg:
closes #239


src/com/carolinarollergirls/scoreboard/core/ScoreBoard.java
```diff
@@ -109,6 +110,7 @@
     public XmlScoreBoard getXmlScoreBoard();
 
     public enum Value implements PermanentProperty {
+        VERSION(String.class, ScoreBoardManager.getVersion()),
```
brian-brazil: There's more to the version than just that, I'd dump all the properties.

It'd also be better to pass these in when instantiating the scoreboard.

frank-weinberg: > There's more to the version than just that, I'd dump all the properties.

Commit and timestamp are already in there. What use do you expect from build username and host?

> It'd also be better to pass these in when instantiating the scoreboard.

Why would that be better? It'd be more code, use more memory, and it'd diverge from every other instance where this value is used.

brian-brazil: > Commit and timestamp are already in there

Only if it's a custom build. Better to expose it all, just in case.

> Why would that be better?

It avoids a layering violation, the ScoreBoard shouldn't know that ScoreBoardManager exists.

frank-weinberg: > Only if it's a custom build. Better to expose it all, just in case.

If we can't trust our release builds to be unambiguously numbered, we have a different level of problem.

> It avoids a layering violation, the ScoreBoard shouldn't know that ScoreBoardManager exists.

a) The ScoreBoard already does know ScoreBoardManager since you introduced it when moving Media over.
b) We have already discussed that the code is a mess in terms of layering and fixing that would amount to more or less a rewrite from the gorund up. In the meantime this one line doesn't add much damage but will make Smasher's life easier with the coversion tool.

brian-brazil: > If we can't trust our release builds to be unambiguously numbered, we have a different level of problem.

I'm not worried about our release builds, I'm worried about random builds.

> The ScoreBoard already does know ScoreBoardManager since you introduced it when moving Media over.

That should be fixed then.

> We have already discussed that the code is a mess in terms of layering and fixing that would amount to more or less a rewrite from the gorund up.

It's a lot better now that XML is gone.

frank-weinberg: >I'm not worried about our release builds, I'm worried about random builds.

But those will have the git revision in there unless someone intentionally fiddles with the build process. And we can't protect against that.

Regarding the layering violation, I have moved the Version handling to utils, which I understand to be accessible from everywhere.

src/com/carolinarollergirls/scoreboard/core/ScoreBoard.java
```diff
@@ -106,6 +107,7 @@
     public Media getMedia();
 
     public enum Value implements PermanentProperty {
+        VERSION(String.class, Version.get()),
```
brian-brazil: Hmm, is there a race here? If this class is loaded before loadVersion runs it'll be empty. I'd suggest passing the version down as an argument when the ScoreBoard is created.

frank-weinberg: No need to use an argument when the value is available through a static function.

brian-brazil: That basically the long way of having a global variable, pass it down would be cleaner I think.

frank-weinberg: Well, the value is inherently global. If you want to pass it as parameter, you also have to pass it through multiple layers to the version servlet. And then do the same with the base path and the logger. A huge amount of extra code just to follow some dogma.

brian-brazil: The interplay between the ScoreBoardManager and Version smells a bit. If all the version parsing code was in the class that'd be okay, the split seems arbitrary currently.

frank-weinberg: The scope creep continues, but this should now resolve any and all layering violations that involve ScoreBoardManager.
(I had to split off the Logger as the Version parsing code used it and after that there wasn't much left anyways.)

brian-brazil: Thanks. Logging needs a complete re-work, that's something for another day.

Issue #269, opened 2019-05-15
Adding old LT fielding box trips is buggy


brian-brazil:
Attempting to add a box trip to an old jam seems to be quite buggy. If you add a skater and a box trip in one step, the skater isn't created yet (as that happens only when you hit submit) so works oddly. Relatedly hitting "abort" does not remove any box trips you created.

brian-brazil:
Playing around a little, I think the issue is related to when the start of the penalty is advanced into the next jam (jam 0), which you can't then undo.

PR #334, opened 2019-06-27
Fix various corner cases around box trips.

brian-brazil:
Handle adding past box trips seperate from new box trips via
PT controls. This fixes issues introduced around fielding advances
and box trip tracking against the LT Box button.

Remove notion of ending box trips between jams, this doesn't work as
it'd require released skaters to be fielded in the upcoming jam.
It's also not something I've ever seen needed.

We still need to remove the Submit and make all changes immediate, but this fixes all the other issues around this.

### DIFF
--- a/html/controls/lt/lt-sheet.js
+++ b/html/controls/lt/lt-sheet.js
@@ -362,11 +362,9 @@ function prepareFieldingEditor(teamId) {
 					+ (afterSP?' after SP':''));
 		}
 		if (['EndJamNumber', 'EndBetweenJams', 'EndAfterSP'].includes(key)) {
-			var between = isTrue(WS.state[prefix+'EndBetweenJams']);
 			var afterSP = isTrue(WS.state[prefix+'EndAfterSP']);
 			var jam = WS.state[prefix+'EndJamNumber'];
-			row.find('.tripEndText').text((between?' Before ':' ') + (jam == 0 ? 'ongoing' : 'Jam ' + jam) 
-					+ (afterSP?' after SP ':' '));
+			row.find('.tripEndText').text((jam == 0 ? 'ongoing' : 'Jam ' + jam) + (afterSP?' after SP ':' '));
 		}
 		if (key == 'Fielding') {
 			row.addClass(v);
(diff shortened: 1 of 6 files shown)


frank-weinberg:
> Remove notion of ending box trips between jams

So how do you record it when a skater goes to the box during a jam and after the jam their partially served penalty is overturned by an OR?

brian-brazil:
The same as if they exited the box exactly when the jam ended, plus it'd be in the OR paperwork.

frank-weinberg:
Which would be wrong. They weren't released during the jam. (And if the refs remove a penalty without an OR or a self-report is sorted out between jams, there is no OR paperwork.)

brian-brazil:
The current data model isn't setup for this, and it's a really rare scenario. This is the sort of thing where you'd make a note that the penalty was overturned/self-report and leave it at that, which we don't record either despite that being clearly documented.

frank-weinberg:
> The current data model isn't setup for this, and it's a really rare scenario.

I disagree with both statements.

brian-brazil:
> The current data model isn't setup for this

That it requires that the skater who was released to skate in the next jam is a data modelling issue. I tried to keep the existing behaviour, but that made it impossible. This is the best way to get this functionality functional, so we can do a beta.

frank-weinberg:
> That it requires that the skater who was released to skate in the next jam is a data modelling issue.

The model doesn't require that. If the implementation does, it's an implementation bug.

brian-brazil:
The model as implemented does require it, and if it's a bug then it's one of many in this bit of code - which I'm fixing here as this code was unusably buggy.

frank-weinberg:
> The model as implemented does require it,

Neither model nor implementation do require it. I've just taken current HEAD and created a penalty that ends between jams without the skater being fielded in the next jam. And there was not a single problem doing so.

brian-brazil:
I'm unable to create such a BoxTrip on HEAD, I get NPEs and if I try to do it for the just ended jam the end can't be changed from Ongoing.

frank-weinberg:
Screenshot_20190706_214347

So how about you provide a proper bug report instead of removing working features?

brian-brazil:
That'd be #269, which I'm trying to fix here. Behaviour is now fully consistent.

frank-weinberg:
#269 does not even mention box trips ending between jams.

brian-brazil:
No, it covers box trip editing being generally broken in the fielding editor. This was what I came up with as part of fixing all of that.

How about you build on this PR? The unittests should give you a good base to work from.

PR #336, opened 2019-06-28
Reorganize HTML and add tab/screen to download game data files

frank-weinberg:
- Add ruleset editor as a tab
- Merge Media Management screen and Save/Load tab; use in both places
- Add a tab to download game data files to merged screen/tab
- Add teams editor and Scoreboard view settings as standalone screens
- Place in-game-screens under `/nso/<position>`
- Place config screens under '/settings'
- Move overlay to `/view/overlay` - it's the only one, so no need for the `overlays/<name>` structure anymore
- Components (tabs/sheets/input areas) used by multiple screens are placed under '/components'
- Move remaining screens to separate (sub)directiories, so all URLs are folder names and don't have `.html` suffix for consistency.


html/settings/view/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<title>ScoreBoard View Settings</title>
```
brian-brazil: Just ScoreBoard Settings here and in the index would make more sense, given the left column affects non-view things.

src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
```diff
@@ -122,6 +122,14 @@ protected void init() {
         c.addFilter(mf, "/*", 1);
         c.setResourceBase((new File(ScoreBoardManager.getDefaultPath(), "html")).getPath());
 
+        sh = new ServletHolder("game-data", new DefaultServlet());
+        sh.setInitParameter("cacheControl", "no-cache");
+        sh.setInitParameter("etags", "true");
+        sh.setInitParameter("resourceBase", (new File(ScoreBoardManager.getDefaultPath(), "game-data")).getPath());
+        sh.setInitParameter("dirAllowed", "true");
+        sh.setInitParameter("pathInfoOnly", "true");
+        c.addServlet(sh, "/game-data/*");
```
brian-brazil: Rather than doing this, I'd suggest moving the game-data directory under html.

html/components/data-management-tab.js
```diff
@@ -0,0 +1,249 @@
...
+		var sbUploadTable = $("<table>").addClass("Upload")
+			.appendTo($("<td>").appendTo($("<tr>").appendTo(table)));
+		$("<tr>").addClass("Name").appendTo(sbUploadTable)
+			.append("<td>Upload ScoreBoard JSON</td>");
+		var contentTd = $("<td>")
+			.appendTo($("<tr>").addClass("Content").appendTo(sbUploadTable));
+	
+		var iframeId = "SaveLoadUploadHiddenIframe";
+		var uploadForm = $("<form method='post' enctype='multipart/form-data' target='"+iframeId+"'/>")
+			.append("<iframe id='"+iframeId+"' name='"+iframeId+"' style='display: none'/>")
```
brian-brazil, line 247: I'd prefer that we use completely separate pages and use browser tabs as you previously suggested. This means that refresh keeps you on the same page, and avoids two layers of tabs when looking at media. Iframes are generally something to avoid (and may break with future browser security changes) - plus it makes page loads slower.

html/components/data-management-tab.js
```diff
@@ -0,0 +1,249 @@
...
+				return;
+			}
+			div.dialog("option", "buttons", buttonsUploadClose);
+		}).change();
+	}
+	
+	function createGameDataTab(tab) {
+		var iframe = $('<iframe src="/game-data">').attr('id', 'GameData');
+		tab.append($('<table>')
+				.append($('<td>').addClass('Info').text('To download, right-click and Save - to view JSON, left-click'))
```
brian-brazil, line 197: Let's just have a link to the actual directory, and add a README if docs are needed. No need for a reload button if you can use the browser's.

html/components/data-management-tab.js
frank-weinberg: Using browser tabs has downsides, that I didn't immediately see:
- If you open all of the tabs at once, the user will have to close multiple tabs despite possibly only using one of them.
- If you open the tabs on demand, their order will vary. Also users might see the links that open the tabs as being the default way to switch, but they will trigger a reload, which is undesirable.
- If you run the page in full screen, you won't see the tab bar.

I see the iframe as a temporary solution until someone implements a nicer interface.

src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
frank-weinberg: In which case we could simply integrate it as a new type of Media, which would also address your concerns regarding refresh and iframe.

html/components/data-management-tab.js
brian-brazil: > Also users might see the links that open the tabs as being the default way to switch, but they will trigger a reload, which is undesirable.

The `target` attribute can handle that.

I'm not worried about full screen for any of these, and if we have a consistent ordering in the list then users will generally open them in that order.

src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
brian-brazil: I'm not suggesting it be a type of media, just a link like we do for /custom/

html/components/data-management-tab.js
frank-weinberg: `target` will ensure that we don't open multiple tabs for the same screen but it will still reload the tab every time the link is clicked.

And I'm pretty sure most users will by default only open the main operator screen, opening other ones as they need them.

frank-weinberg:
I have gone forward and added game data and custom screens to the data management interface.

For game data this removes the need for refreshing the list of files and provides some other convenience functions (e.g. renaming - and if we are adding support for prepared game, upload might become useful as well)
For custom screens this allows users to install screens by simply uploading a zip file through the interface without the need to fiddle around in the file system, as long as screen creators adhere to the principle of having all the files for a screen in one directory. Adding a launch menu would also be a straightforward task from this.

html/settings/view/index.html
frank-weinberg: That's done now. (I've shortened to 'Settings' for the operator interface.)

html/components/data-management-tab.js
brian-brazil: This leaves all the problems that the existing system has, actual tabs would make more sense.

brian-brazil:
> I have gone forward and added game data and custom screens to the data management interface.

This doesn't really work. You can't download the game data from there, there's no notion of "statsbook" in the scoreboard, and the Names have no relevance either. This is a square peg in a round hole, a link is enough.

> For custom screens this allows users to install screens by simply uploading a zip file through the interface without the need to fiddle around in the file system

That's handy, but doesn't resolve the above.

In addition the index page links no longer take you to the index page, `Up/Download` is how to spell that (or Up-/Down-load), and the nested tab colours are a bit garish.

frank-weinberg:
> You can't download the game data from there,

There's a download button for each file. And differing from a flat directory listing, it will ask the browser to save the file by default instead of opening a page of gibberish (from the viewpoint of an average user).

> there's no notion of "statsbook" in the scoreboard

For now that's a folder where people can store their prepared statsbook in order to have it available at the event. Or it could be used to distribute completed statsbooks at tournaments. And if it isn't used, it doesn't take away any significant amount of resources.

> and the Names have no relevance either.

That's a feature we got for free. If nobody finds a good use for it, we're still at net zero for this. Compared to a flat listing, we got the option to show new files as soon as they are created without the need for refresh. We got download instead of display on left click. We got an interface that allows users to clean up their game data folder without the need to navigate around the file system. And it's all integrated into the look and feel of the application.

> In addition [...]

These are fixed.

html/components/data-management-tab.js
frank-weinberg: > all the problems that the existing system has

You named refresh keeping you on the same page, the two layers of tabs and the iframe. - The iframe has since gone. For the two layers of tabs I I think the different colors work really well. This leaves just the refresh. But with the other tabs being available as standalone screens, it's likely that these tabs on the operator screen will only be used for quick changes just before or during the game or downloading the game data afterwards. In all of these cases, the user doesn't want to be on a non-operator tab for a prolonged time, so the chance of a reload taking them  to the 'wrong' tab is pretty slim. Plus, if this is actually a significant problem, it's rather easy to store the current tab as a page parameter and read it on reload (like we do with the operator name).

In my opinion the problems I mentioned above for the browser tab solution are way more serious.

brian-brazil:
> For now that's a folder where people can store their prepared statsbook in order to have it available at the event. Or it could be used to distribute completed statsbooks at tournaments. And if it isn't used, it doesn't take away any significant amount of resources.

I think that will cause confusion, as if it's there it'd be presumed that it means something to the scoreboard - which it doesn't.

> That's a feature we got for free. If nobody finds a good use for it, we're still at net zero for this. 

It's a wart at least, I'd consider it a bug. In all the other media types it has a UI meaning, but here it has none so once again we're looking at confusion. Also lacking the original filename I can't tell what the file actually is.

> the nested tab colours are a bit garish.

This is still the case

html/components/data-management-tab.js
brian-brazil: > In all of these cases, the user doesn't want to be on a non-operator tab for a prolonged time, so the chance of a reload taking them to the 'wrong' tab is pretty slim.

It depends on what they're doing, if everything is on the operator screen would everyone realise these other screens exist? I still think actual separate pages makes more sense, and limit future issues.

html/nso/sbo/index.js
```diff
@@ -101,7 +106,9 @@ function logout() {
 }
 
 function createTab(title, tabId) {
-	if (typeof title == "string") title = $("<a>").html(title);
+	if (typeof title == "string") title = $("<a>").html(title).click(function() {
+		location.hash = '#'+tabId;
```
brian-brazil: This isn't working with nested tabs. I'd suggest removing up/download from the operator screen (they can do logos from the teams page now anyway and that's the thing changed most often). I'd also put rulesets last rather than 2nd, it's pretty rare that they need changing.

frank-weinberg: > This isn't working with nested tabs.

Neither would a separate browser tab.

> I'd suggest removing up/download from the operator screen (they can do logos from the teams page now anyway and that's the thing changed most often).

I'd anticipate that downloading JSON files will become regularly used. (Plug in USB key, go to download tab and save to the key - no need to find the correct folder via explorer etc)

> I'd also put rulesets last rather than 2nd, it's pretty rare that they need changing.

Usage frequency has never been a factor in the tab order (or Teams would be 2nd). This is the place where historically the Policies tab has been, which has been replaced by the rulesets. So it's one less item that people skipping several versions will have to adjust to.

brian-brazil: > Neither would a separate browser tab.

It'd work fine by default.

> I'd anticipate that downloading JSON files will become regularly used.

By the operator during a game though?

> Usage frequency has never been a factor in the tab order (or Teams would be 2nd).

Maybe it's time to switch that then?

frank-weinberg: > It'd work fine by default.

How could it? It'd jump to the first (sub)tab just as the operator screen did all the time.

> By the operator during a game though?

Right after the game or between games of a double or triple header seems likely.

> Maybe it's time to switch that then?

If you want to. I suggest that you start by collecting representative usage stats.
But I'm not willing to have this PR sit there and collect more merge conflicts in the meantime.

brian-brazil: > How could it? It'd jump to the first (sub)tab just as the operator screen did all the time.

If it's a browser tab, then refresh leaves you where you were.

> Right after the game or between games of a double or triple header seems likely.

By a game data manager or stats crew maybe, but not the operator.

frank-weinberg: > If it's a browser tab, then refresh leaves you where you were.

Yeah, sure. That's why I had to add code so that the operator screen remembers its opened tab on refresh.

> By a game data manager or stats crew maybe, but not the operator.

HNSO or home league rep come to the SBO and ask: "Can you quickly put the game data file on this stick so I can generate the statsbook at home?" Or they do it themselves but that doesn't change the fact that it is more convenient to be able to do it from the screen that's already open than having to open another one.

brian-brazil: > That's why I had to add code so that the operator screen remembers its opened tab on refresh.

That doesn't work for the nested tabs on the Up/Download on the operator screen.

I'd suggest removing the Up/Download screen (and Rulesets), as use of them will be relatively niche and resolves this issue. In such a setup the HNSO likely has network access anyway, and can pull it themselves.

frank-weinberg: > That doesn't work for the nested tabs on the Up/Download on the operator screen.

We're going in circles here - I never claimed it would. You requested it to remember the opened tab on reload as it would do with separate browser tabs. I implemented that. If you want more features, I suggest you open a feature request in the issue tracker instead of trying to continuously scope creep my PR.

brian-brazil: It doesn't do that though, which your original proposal would have. The creep is due to changing from the original proposal.

frank-weinberg: Ok, I've reverted the behaviour to current HEAD. If you want any changes to how tabs behave, bring those up separate from this PR.

brian-brazil:
I'm generally in favour of what this PR does, but I don't like the changes to the operator screen. Rulesets are an advanced feature, and do not deserve prominence that they've never had before - they've always lived on only the index page. Similarly I don't think linking to all the media options is needed, particularly as you can upload a logo directly from the teams page these days.

I think this should keep the operator screen as it is in 3.9.5 and current head: team/time, settings, teams and save/load. The UI an operator has to deal with should be made simpler, not more complex.

frank-weinberg:
Rulesets show up in the "Start New Game" dialog. It hardly gets more prominent than that. And they replace Preferences that have always existed as a tab.

Save/Load no longer even exists as a separate component. And one more tab in the existing tab bar is insignificant in terms of complexity.

In the end your reasons come down to personal taste. And I don't see why the personal taste of the reviewer should trump the preference of the person who put in the work.

brian-brazil:
> Rulesets show up in the "Start New Game" dialog. It hardly gets more prominent than that. And they replace Preferences that have always existed as a tab.

An option that you usually don't need to care about doesn't make it prominent, and the closest successor to policies are now on the settings page.

> And one more tab in the existing tab bar is insignificant in terms of complexity.

One more tab with subtabs lots of buttons, that also aren't usually relevant -  which would confuse new users.

> And I don't see why the personal taste of the reviewer should trump the preference of the person who put in the work.

That approach would mean we should accept all PRs. My reasoning is based on keeping clutter off the operator screen, what you and I can effectively use is much broader than what a first time user can handle.

frank-weinberg:
> the closest successor to policies are now on the settings page.

The commit was titled: "Move Policies to Rules" - and that is what it did.

> One more tab with subtabs lots of buttons, that also aren't usually relevant - which would confuse new users.

New users wouldn't see those subtabs - why should they open the tab in the first place? And even if - it's straightforward to return to the main tab if you don't understand the other ones. No need to add inconvenience for experienced users.

> That approach would mean we should accept all PRs.

Because requiring it to not break existing functionality is also just a personal preference?

brian-brazil:
> The commit was titled: "Move Policies to Rules" - and that is what it did.

That doesn't justify the change.

> New users wouldn't see those subtabs - why should they open the tab in the first place?

Maybe because they're clicking everything to figure out what it's all for. Every button we have is additional cognitive overhead for all users.

> Because requiring it to not break existing functionality is also just a personal preference?

I'm not following your argument.

frank-weinberg:
> Every button we have is additional cognitive overhead for all users.

So you're proposing to make the mobile operator the default?

PR #344, opened 2019-07-09
Fix UI when new ruleset created and viewed

### LINKED ISSUE #340: Ruleset Editor Blank Values on Initial Ruleset creation
When first entering the ruleset editor, if you create a new ruleset that has another existing ruleset as it's parent, all of the editors are active, and their values are blank (as if you had created a new ruleset without a parent).

Additionally, once an existing ruleset has been opened, the ruleset editor exhibits the correct behavior upon creating a ruleset (all values populated, all editor check boxes are not checked).

brian-brazil:
Fixes #340

Simplify and cleanup code.

### DIFF
--- a/html/controls/rulesets.js
+++ b/html/controls/rulesets.js
@@ -1,15 +1,15 @@
-initialize();
 var rulesets = {};
+var activeRuleset = null;
 var definitions = {};
-var knownRulesets = {};
+initialize();
 
-function loadRulesets(displayId) {
+function loadRulesets() {
 	var rs = $(".rulesets div");
 	$("#new_parent").empty();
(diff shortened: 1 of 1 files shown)


html/controls/rulesets.js
```diff
@@ -1,15 +1,15 @@
-initialize();
 var rulesets = {};
+var activeRuleset = null;
 var definitions = {};
-var knownRulesets = {};
+initialize();
 
 function loadRulesets(displayId) {
```
frank-weinberg: The parameter is now unused

brian-brazil: Fixed

PR #345, opened 2019-07-10
Remove properties file, hardcode instead.

brian-brazil:
The port and binding property still exist, and
can be set on the command line with -D.


src/com/carolinarollergirls/scoreboard/Main.java
```diff
@@ -52,11 +52,6 @@ private void parseArgv(String[] argv) {
             else if(arg.equals("--nogui") || arg.equals("-G")) {
                 gui = false;
             }
-
-            else if(arg.startsWith("--port=") || arg.startsWith("-p=")) {
-                String port = arg.split("=")[1];
-                ScoreBoardManager.setPropertyOverride("com.carolinarollergirls.scoreboard.jetty.JettyServletScoreBoardController.port", port);
```
frank-weinberg, line 58: Why remove this? `-p=` is a lot easier to remember and type than `-Dcom.carolinarollergirls.scoreboard.jetty.JettyServletScoreBoardController.port=`. I'd rather think about adding something like `-l` and `--localhost` for that property.

brian-brazil: It's a fairly advanced feature that few users need (or are even aware exists), plus two ways to configure things is asking for confusion about which takes precedence.

frank-weinberg: "Few users need it, so we can make it significantly more obscure" is not a convincing argument. And there's no need to even document the `-D` way - we can just treat it as something that happens to have that effect but is not an official feature at which point everybody who is able to figure out that it exists should also be able to figure out that `-p`/`-l` takes precedence (or that they should just leave it alone).

brian-brazil: Updated to take a different approach.

src/com/carolinarollergirls/scoreboard/ScoreBoardManager.java
```diff
@@ -118,41 +111,9 @@ private static void loadVersion() {
         printMessage("Carolina Rollergirls Scoreboard version "+getVersion());
     }
 
-    private static void loadProperties() {
```
frank-weinberg, line 119: This leaves `FileInputStream`, `FileNotFoundException`, `HashMap`, `Map`, and `ConcurrentHashMap` unused.

src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
```diff
@@ -67,27 +67,12 @@ public JettyServletScoreBoardController(ScoreBoard sb, JSONStateManager jsm) {
...
-            ScoreBoardManager.printMessage("ScoreBoard configured to listen ONLY on localhost interface.");
-            SocketConnector sC = new SocketConnector();
-            sC.setHost("localhost");
-            sC.setPort(port);
-            server = new Server();
-            server.addConnector(sC);
-        } else {
-            server = new Server(port);
-        }
+    protected void init(String host, int port) {
```
frank-weinberg, line 69: This shadows a class attribute (that appears to be unused)

src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
```diff
@@ -128,27 +113,11 @@ protected void init() {
         HttpServlet ljs = new LoadJsonScoreBoard(scoreBoard);
         c.addServlet(new ServletHolder(ljs), "/LoadJSON/*");
 
-        Enumeration<?> keys = ScoreBoardManager.getProperties().propertyNames();
```
frank-weinberg, line 131: `Enumeration` is now unused

src/com/carolinarollergirls/scoreboard/jetty/MediaServlet.java
```diff
@@ -32,18 +33,18 @@
...
+    public MediaServlet(ScoreBoard sb, String dir) {
+        scoreBoard = sb;
+        htmlDirName = dir;
     }
 
-    @Override
-    public String getPath() { return "/Media"; }
-
-    @Override
     protected void doPost(HttpServletRequest request, HttpServletResponse response) throws ServletException,IOException {
```
frank-weinberg, line 45: This method is still `@Override`

brian-brazil:
All fixed.

PR #347, opened 2019-07-11
Add support for penalty box substitutions

### LINKED ISSUE #269: Adding old LT fielding box trips is buggy

frank-weinberg:
Also:

- color the (P)LT box field, when a Skater has an unserved penalty
- make the buttons etc. in the popups in the P/LT and SK screens submit immediately
- fix handling of adding past box trips and moving their start/end around (closes #269)
- add possibility to unend box trips

### DIFF
--- a/html/controls/lt/lt-sheet.js
+++ b/html/controls/lt/lt-sheet.js
@@ -249,6 +249,7 @@ function openFieldingEditor(p, j, t, pos, upcoming) {
 	var notFieldedField = fieldingEditor.find('#notFielded').attr('checked', isTrue(WS.state[prefix+'NotFielded']));
 	var sitFor3Field = fieldingEditor.find('#sitFor3').attr('checked', isTrue(WS.state[prefix+'SitFor3']));
 	var annotationField = fieldingEditor.find('#annotation').val(WS.state[prefix+'Annotation']);
+	fieldingEditor.find('#notFielded').toggleClass('Hide', isTrue(upcoming));
 	fieldingEditor.find('.BoxTrip').addClass('Hide');
 	fieldingEditor.find('.'+WS.state[prefix+'Id']).removeClass('Hide');
 	fieldingEditor.data('prefix', prefix);
--- a/html/controls/lt/lt-sheet.js
+++ b/html/controls/lt/lt-sheet.js
@@ -248,6 +248,7 @@ function openFieldingEditor(p, j, t, pos, upcoming) {
 	var skaterField = fieldingEditor.find('#skater').val(WS.state[prefix+'Skater']);
 	var notFieldedField = fieldingEditor.find('#notFielded').attr('checked', isTrue(WS.state[prefix+'NotFielded']));
 	var sitFor3Field = fieldingEditor.find('#sitFor3').attr('checked', isTrue(WS.state[prefix+'SitFor3']));
+	var annotationField = fieldingEditor.find('#annotation').val(WS.state[prefix+'Annotation']);
 	fieldingEditor.find('.BoxTrip').addClass('Hide');
 	fieldingEditor.find('.'+WS.state[prefix+'Id']).removeClass('Hide');
 	fieldingEditor.find('#addTrip, #submit').unbind('click');
@@ -258,6 +259,7 @@ function openFieldingEditor(p, j, t, pos, upcoming) {
 		WS.Set(prefix+'Skater', skaterField.val());
 		WS.Set(prefix+'NotFielded', notFieldedField.attr('checked') != null);
 		WS.Set(prefix+'SitFor3', sitFor3Field.attr('checked') != null);
+		WS.Set(prefix+'Annotation', annotationField.val());
 		fieldingEditor.dialog('close');
 	});
 	
@@ -271,27 +273,38 @@ function prepareFieldingEditor(teamId) {
 
 	function initialize() {
 		var table = $('<table>').appendTo($('#FieldingEditor'));
+		
 		var row = $('<tr>').addClass('Skater').appendTo(table);
 		$('<td>').append($('<select>').attr('id', 'skater').append($('<option>').attr('value', '').text('None/Unknown'))).appendTo(row);
 		var notFieldedField = $('<td>').append($('<button>').attr('id', 'notFielded').button().text('No Skater fielded')).appendTo(row).children('button');
 		notFieldedField.click(function(){notFieldedField.attr('checked', notFieldedField.attr('checked') == null);});
 		var sitFor3Field = $('<td>').append($('<button>').attr('id', 'sitFor3').button().text('Sit out next 3')).appendTo(row).children('button');
 		sitFor3Field.click(function(){sitFor3Field.attr('checked', sitFor3Field.attr('checked') == null);});
+		
+		row = $('<tr>').addClass('Skater').appendTo(table);
+		$('<td>').attr('colspan', '3')
+			.append($('<input type="text">').attr('size', '40').attr('id', 'annotation'))
+			//.append($('button').attr('id', 'setAnnotation').text('Set').button())
+			.appendTo(row);
+		
 		row = $('<tr>').addClass('tripHeader').appendTo(table);
 		$('<td>').attr('colspan', '2').text('Box Trips').appendTo(row);
 		$('<td>').appendTo(row);
+		
 		row = $('<tr>').addClass('tripHeader').appendTo(table);
 		$('<td>').text('Start').appendTo(row);
 		$('<td>').text('End').appendTo(row);
 		$('<td>').appendTo(row);
+		
 		row = $('<tr>').attr('id', 'tripFooter').appendTo(table);
-		$('<td>').append($('<button>').attr('id', 'addTrip').text('Add Box Trip')).appendTo(row);
-		$('<td>').addClass('ButtonCell').append($('<button>').attr('id', 'abort').text('Abort')).click(function() {
+		$('<td>').append($('<button>').attr('id', 'addTrip').text('Add Box Trip').button()).appendTo(row);
+		$('<td>').addClass('ButtonCell').append($('<button>').attr('id', 'abort').text('Abort').button()).click(function() {
 			fieldingEditor.dialog('close');
 		}).appendTo(row);
-		$('<td>').addClass('ButtonCell').append($('<button>').attr('id', 'submit').text('Submit')).appendTo(row);
+		$('<td>').addClass('ButtonCell').append($('<button>').attr('id', 'submit').text('Submit').button()).appendTo(row);
 
-		WS.Register(['ScoreBoard.Team('+teamId+').Skater'], function(k,v) { processSkater(k,v); })
+		WS.Register(['ScoreBoard.Team('+teamId+').Skater(*).Role',
+			'ScoreBoard.Team('+teamId+').Skater(*).Number'], function(k,v) { processSkater(k,v); })
 		WS.Register(['ScoreBoard.Team('+teamId+').BoxTrip'], function(k,v) { processBoxTrip(k,v); })
 
 		fieldingEditor = $('#FieldingEditor').dialog({
@@ -299,21 +312,19 @@ function prepareFieldingEditor(teamId) {
 			closeOnEscape: false,
 			title: 'Fielding Editor',
 			autoOpen: false,
-			width: '500px',
+			width: '450px',
 		});
 	}		
 	
 	function processSkater(k, v) {
-		if (k.Skater == null || k.parts[3] != "Number") return;
-
 		var role = WS.state['ScoreBoard.Team('+teamId+').Skater('+k.Skater+').Role'];
 		var number = WS.state['ScoreBoard.Team('+teamId+').Skater('+k.Skater+').Number'];
 		var playing = (role != null && role != 'NotInGame'); 
 
 		var option = $("#FieldingEditor #skater option[value='"+k.Skater+"']");
 		var inserted = false;
-		if (v != null && option.length == 0) {
-			var option = $('<option>').attr('value', k.Skater).text(v);
+		if (number != null && option.length == 0) {
+			var option = $('<option>').attr('value', k.Skater).text(number);
 			$('#FieldingEditor #skater').children().each(function (idx, s) {
 				if (s.text > number && idx > 0) {
 					$(s).before(option);
@@ -337,19 +348,19 @@ function prepareFieldingEditor(teamId) {
 		var row = $('#FieldingEditor .BoxTrip[id='+k.BoxTrip+']');
 		if (v != null && row.length == 0) {
 			row = $('<tr>').addClass('BoxTrip').attr('id', k.BoxTrip).insertBefore('#FieldingEditor #tripFooter');
-			$('<td>').append($('<button>').addClass('tripModify').text('-').click(function() {
+			$('<td>').append($('<button>').addClass('tripModify').text('-').button().click(function() {
 				WS.Set(prefix+'StartEarlier', true);
 			})).append($('<span>').addClass('tripStartText'))
-			.append($('<button>').addClass('tripModify').text('+').click(function() {
+			.append($('<button>').addClass('tripModify').text('+').button().click(function() {
 				WS.Set(prefix+'StartLater', true);
 			})).appendTo(row);
-			$('<td>').append($('<button>').addClass('tripModify').text('-').click(function() {
+			$('<td>').append($('<button>').addClass('tripModify').text('-').button().click(function() {
 				WS.Set(prefix+'EndEarlier', true);
 			})).append($('<span>').addClass('tripEndText').text('ongoing'))
-			.append($('<button>').addClass('tripModify').text('+').click(function() {
+			.append($('<button>').addClass('tripModify').text('+').button().click(function() {
 				WS.Set(prefix+'EndLater', true);
 			})).appendTo(row);
-			$('<td>').addClass('Col3').append($('<button>').addClass('tripRemove').text('Remove').click(function() {
+			$('<td>').addClass('Col3').append($('<button>').addClass('tripRemove').text('Remove').button().click(function() {
 				WS.Set(prefix+'Delete', true);
 			})).appendTo(row);
 		}
--- a/html/controls/lt/lt-sheet.js
+++ b/html/controls/lt/lt-sheet.js
@@ -281,9 +281,8 @@ function prepareFieldingEditor(teamId) {
 		
 		row = $('<tr>').addClass('Skater').appendTo(table);
 		$('<td>').attr('colspan', '3')
-			.append($('<input type="text">').attr('size', '35').attr('id', 'annotation'))
-			.append($('<button>').attr('id', 'setAnnotation').text('Set').button().click(function() {
-				WS.Set(fieldingEditor.data('prefix')+'Annotation', fieldingEditor.find('#annotation').val());
+			.append($('<input type="text">').attr('size', '40').attr('id', 'annotation').change(function() {
+				WS.Set(fieldingEditor.data('prefix')+'Annotation', $(this).val());
 			})).appendTo(row);
 		
 		row = $('<tr>').addClass('Skater').appendTo(table);
--- a/html/controls/lt/lt-sheet.js
+++ b/html/controls/lt/lt-sheet.js
@@ -120,6 +120,8 @@ function prepareLtSheetTable(element, teamId, mode) {
 						setBoxTripSymbols(jamRow, '.Box'+k.Fielding, v);
 					} else if (k.BoxTripSymbolsAfterSP != null) {
 						if (isTrue(WS.state[prefix+'StarPass'])) { setBoxTripSymbols(spRow, '.Box'+k.Fielding, v); }
+					} else if (k.Annotation != null) {
+						jamRow.children('.'+k.Fielding).toggleClass('hasAnnotation', v!= null && v != '');
 					}
 				}
 				break;
--- a/html/controls/lt/lt-sheet.js
+++ b/html/controls/lt/lt-sheet.js
@@ -41,11 +41,15 @@ function prepareLtSheetTable(element, teamId, mode) {
 				}
 			});
 
-			WS.Register(['ScoreBoard.Team('+teamId+').Position(*).Number', 'ScoreBoard.Team('+teamId+').Position(*).CurrentBoxSymbols'], function(k, v) {
+			WS.Register(['ScoreBoard.Team('+teamId+').Position(*).Number',
+				'ScoreBoard.Team('+teamId+').Position(*).CurrentBoxSymbols',
+				'ScoreBoard.Team('+teamId+').Position(*).Annotation'], function(k, v) {
 				if (k.field == "Number") {
 					element.find("#upcoming .Skater."+k.Position).text(v);
 				} else if (k.field == "CurrentBoxSymbols") {
 					element.find("#upcoming .Box.Box"+k.Position).text(v);
+				} else if (k.field == "Annotation") {
+					element.find("#upcoming .Skater."+k.Position).toggleClass('hasAnnotation', v != '');
 				}
 			});
 		}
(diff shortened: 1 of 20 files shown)


html/controls/plt/plt-input.js
```diff
@@ -600,4 +623,113 @@ function preparePenaltyEditor() {
...
+		var table = $('<table>').appendTo($('#AnnotationEditor'));
+		var row = $('<tr>').addClass('Box').appendTo(table);
+		$('<td>').append(subDropdown)
+				.append($('<button>').text('Substitute').button().click(function() {
+					var prefix = jamPrefix + annotationEditor.data('position') + ').';
+					WS.Set(prefix + 'Skater', subDropdown.val());
+					WS.Set(prefix + 'Annotation', 'Substitute for #' + annotationEditor.data('skaterNumber'));
+					annotationEditor.dialog('close');
+				})).appendTo(row);
+		$('<td>').append($('<button>').text('Self Report').button().click(function() { leaveBox('Self Report');})).appendTo(row);
```
brian-brazil: No penalty and Penalty overturned are what the statsbook manual has

(and Left early to be complete)

html/controls/plt/plt-input.js
```diff
@@ -600,4 +623,113 @@ function preparePenaltyEditor() {
...
+	}
+	
+	function processSkater(k, v) {
+		var role = WS.state['ScoreBoard.Team('+teamId+').Skater('+k.Skater+').Role'];
+		var number = WS.state['ScoreBoard.Team('+teamId+').Skater('+k.Skater+').Number'];
+		var playing = (role != null && role != 'NotInGame'); 
+
+		var option = $("#AnnotationEditor #subDropdown option[value='"+k.Skater+"']");
+		var inserted = false;
+		if (number != null && option.length == 0) {
```
brian-brazil: You can use the appendAlpha functions for this, rather than hand-coding

src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
```diff
@@ -231,7 +231,7 @@ public void execute(CommandProperty prop) {
     public void end() {
         set(Value.IS_CURRENT, false);
         set(Value.WALLTIME_END, ScoreBoardClock.getInstance().getCurrentWalltime());
-        if (getTeam().hasFieldingAdvancePending()) {
+        if (getTeam().hasFieldingAdvancePending() && getCurrentFielding().getTeamJam().isRunningOrUpcoming()) {
```
brian-brazil: There's far more broken than this, and this needs unittests. You should start with the ones I wrote, that covers all the corner cases I found.

html/controls/sk/sk-sheet.js
```diff
@@ -306,13 +288,29 @@ function prepareTripEditor() {
...
-						.append($('<td width="40%">').append($('<button>').attr('id','insert_before').text('Insert Before')))
-						.append($('<td width="30%">').append($('<button>').attr('id','remove').text('Remove')))));
+						.append($('<td width="30%">').append($('<button>').attr('id','submit').text('Submit').button().click(function() {
+							if (tripEditor.find('#score').val() == '') {
+								WS.Set(tripEditor.data('prefix')+'Remove', true);
+							}
+							tripEditor.dialog('close');
+						})))
+						.append($('<td width="40%">').append($('<button>').attr('id','insert_before').text('Insert Before').button().click(function() {
+							WS.Set(tripEditor.data('prefix')+'InsertBefore', true);
```
brian-brazil: Insert Before is across two lines for me. For consistency with the jam/period editor, Delete should be first.

html/controls/sk/sk-sheet.js
```diff
@@ -306,13 +288,29 @@ function prepareTripEditor() {
...
+								.append($('<button>').attr('id', 'afterSP').text('SP in this or prior trip').button().click(function() {
+									var check = !$(this).hasClass('checked');
+									$(this).toggleClass('checked', check);
+									WS.Set(tripEditor.data('prefix')+'AfterSP', check);
+								}))))
 				.append($('<tr class="buttons">')
-						.append($('<td width="30%">').append($('<button>').attr('id','submit').text('Submit')))
-						.append($('<td width="40%">').append($('<button>').attr('id','insert_before').text('Insert Before')))
-						.append($('<td width="30%">').append($('<button>').attr('id','remove').text('Remove')))));
+						.append($('<td width="30%">').append($('<button>').attr('id','submit').text('Submit').button().click(function() {
```
brian-brazil: It's no longer submitting anything, so Close would make more sense

html/controls/sk/sk-sheet.js
```diff
@@ -306,13 +288,29 @@ function prepareTripEditor() {
...
 				.append($('<tr>')
 						.append($('<td colspan="3">')
-								.append($('<input type="number" min="0">').attr('id', 'score'))
-								.append($('<input type="checkbox">').attr('id', 'afterSP'))
-								.append($('<span>').addClass('Infotext').text('SP in this or prior trip'))))
+								.append($('<input type="number" min="0">').attr('id', 'score').keydown(function(event) {
+									if (event.which == 13) tripEditor.find('#submit').click(); }).change(function() {
+										WS.Set(tripEditor.data('prefix')+'Score', $(this).val());
+									}))
+								.append($('<button>').attr('id', 'afterSP').text('SP in this or prior trip').button().click(function() {
```
brian-brazil, line 300: This state isn't propagating correctly on subsequent trips in the UI

html/controls/plt/plt-input.js
```diff
@@ -442,17 +455,28 @@ function preparePenaltyEditor() {
...
-		$('<td width="33%">').append($('<span>').text('Period: ')).append($('<button>').addClass('period_minus').text('-1'))
-			.append($('<select>').addClass('Period')).append($('<button>').addClass('period_plus').text('+1')).appendTo(tr);
-		$('<td width="33%">').append($('<span>').text('Jam: ')).append($('<button>').addClass('jam_minus').text('-1'))
-			.append($('<select>').addClass('Jam')).append($('<button>').addClass('jam_plus').text('+1')).appendTo(tr);
-		$('<td width="33%">').append($('<button>').addClass('clear').text('Clear')).appendTo(tr);
+		$('<td width="35%">').append($('<span>').text('Period: ')).append($('<button>').addClass('period_minus').text('-1').button())
+			.append($('<select>').addClass('Period')).append($('<button>').addClass('period_plus').text('+1').button()).appendTo(tr);
+		$('<td width="35%">').append($('<span>').text('Jam: ')).append($('<button>').addClass('jam_minus').text('-1').button())
+			.append($('<select>').addClass('Jam')).append($('<button>').addClass('jam_plus').text('+1').button()).appendTo(tr);
+		$('<td width="15%">').append($('<button>').addClass('set').text('Set Jam').button()).appendTo(tr);
```
brian-brazil: This shouldn't be required, changes should be submitted immediately.

html/controls/plt/plt-input.js
```diff
@@ -442,17 +455,28 @@ function preparePenaltyEditor() {
...
+		$('<td>').append($('<button>').text('Served').attr('id', 'served').button().click(function() {
+			var active = $(this).hasClass('checked');
+			$(this).toggleClass('checked', !active);
+			if (!isTrue(penaltyEditor.data('new'))) {
+				var teamId = penaltyEditor.data('team');
+				var skaterId = penaltyEditor.data('skater');
+				var penaltyNumber = penaltyEditor.data('pnr');
+				var prefix = 'ScoreBoard.Team(' + teamId + ').Skater(' + skaterId + ').Penalty(' + penaltyNumber + ')';
+				WS.Set(prefix + '.ForceServed', active);
+				penaltyEditor.dialog('close');
```
brian-brazil, line 479: This shouldn't close the dialog.

We also need an explicit close button now

html/controls/plt/plt-input.js
```diff
@@ -442,17 +455,28 @@ function preparePenaltyEditor() {
...
+		$('<td width="35%">').append($('<span>').text('Jam: ')).append($('<button>').addClass('jam_minus').text('-1').button())
+			.append($('<select>').addClass('Jam')).append($('<button>').addClass('jam_plus').text('+1').button()).appendTo(tr);
+		$('<td width="15%">').append($('<button>').addClass('set').text('Set Jam').button()).appendTo(tr);
+		$('<td width="15%">').append($('<button>').addClass('clear').text('Clear').button()).appendTo(tr);
 		
 		var tr2 = $('<tr>').appendTo(topTable);
-		$('<td>').append($('<input type="checkbox">').attr('id', 'served'))
-			.append($('<span>').text(' has been served')).appendTo(tr2);
+		$('<td>').append($('<button>').text('Served').attr('id', 'served').button().click(function() {
+			var active = $(this).hasClass('checked');
```
brian-brazil: This isn't right, it's not showing the WS state

html/controls/plt/plt-input.js
```diff
@@ -338,9 +341,16 @@ function preparePltInputTable(element, teamId, mode, statsbookPeriod, alternateN
 			p.append(boxCell);
 			
 			var advanceCell = $('<td>').addClass('Advance').attr('rowspan', 2).click(function() {
-				WS.Set('ScoreBoard.Team('+t+').AdvanceFieldings', true);
+				if (isTrue(WS.state['ScoreBoard.Team('+t+').FieldingAdvancePending'])) {
+					WS.Set('ScoreBoard.Team('+t+').AdvanceFieldings', true);
+				} else if (advanceCell.hasClass('OnTrack')) {
+					openAnnotationEditor(t, id);
```
brian-brazil, line 347: You can't edit annotations from a jam that has just ended from here, as the advance is showing.

There's also no indication on the LT sheet that there's an annotation.

html/controls/lt/lt-sheet.js
```diff
@@ -271,49 +262,69 @@ function prepareFieldingEditor(teamId) {
...
 		var row = $('<tr>').addClass('Skater').appendTo(table);
-		$('<td>').append($('<select>').attr('id', 'skater').append($('<option>').attr('value', '').text('None/Unknown'))).appendTo(row);
-		var notFieldedField = $('<td>').append($('<button>').attr('id', 'notFielded').button().text('No Skater fielded')).appendTo(row).children('button');
-		notFieldedField.click(function(){notFieldedField.attr('checked', notFieldedField.attr('checked') == null);});
-		var sitFor3Field = $('<td>').append($('<button>').attr('id', 'sitFor3').button().text('Sit out next 3')).appendTo(row).children('button');
-		sitFor3Field.click(function(){sitFor3Field.attr('checked', sitFor3Field.attr('checked') == null);});
+		$('<td>').append($('<select>').attr('id', 'skater').append($('<option>').attr('value', '').text('None/Unknown')).change(function() {
+			WS.Set(fieldingEditor.data('prefix')+'Skater', $(this).val());
+		})).appendTo(row);
+		$('<td>').append($('<button>').attr('id', 'notFielded').button().text('No Skater fielded').click(function() {
```
brian-brazil, line 278: This doesn't show n/a for the upcoming jam until the jam starts

html/controls/plt/plt-input.js
```diff
@@ -600,4 +623,113 @@ function preparePenaltyEditor() {
...
+		var row = $('<tr>').addClass('Box').appendTo(table);
+		$('<td>').append(subDropdown)
+				.append($('<button>').text('Substitute').button().click(function() {
+					var prefix = jamPrefix + annotationEditor.data('position') + ').';
+					WS.Set(prefix + 'Skater', subDropdown.val());
+					WS.Set(prefix + 'Annotation', 'Substitute for #' + annotationEditor.data('skaterNumber'));
+					annotationEditor.dialog('close');
+				})).appendTo(row);
+		$('<td>').append($('<button>').text('Self Report').button().click(function() { leaveBox('Self Report');})).appendTo(row);
+		$('<td>').append($('<button>').text('Penalty Revoked').button().click(function() { leaveBox('Penalty Revoked');})).appendTo(row);
```
brian-brazil: The revoked doesn't remove the penalty, so it still shows as waiting to serve.

Self report also removes the LT entry, which is incorrect.

src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
frank-weinberg: Instead of vaguely alluding to stuff that's allegedly broken, you should give specific bug reports. And if you want unit tests, provide some that test the intended functionality and not some that test for a feature to be missing.

brian-brazil: The tests I wrote are pretty extensive. It's so broken as is, that 2-3 clicks is enough to show a bug when editing a box trip.

html/controls/plt/plt-input.js
frank-weinberg: Since the only way to submit changes to a penalty is via their position in the ordered list and changing the jam number may change that position, changing the jam immediately could result in subsequent changes being submitted for the wrong penalty unless we close the window. But closing the window on every +1/-1 would be incredibly annoying.

html/controls/plt/plt-input.js
frank-weinberg: When editing a penalty that already exists, it's way more likely that only one attribute is supposed to be changed. So closing on edit is serving the more common workflow.

html/controls/plt/plt-input.js
frank-weinberg: It's behaving exactly as it was before.

html/controls/plt/plt-input.js
frank-weinberg: Advance overriding the annotation field was a tradeoff I consciously chose:
- The cases where you need to enter an annotation just as the jam ends are exceedingly rare. And if one arises, you can still enter the annotation in the fielding editor.
- If a user were to enter a substitution before pressing "Advance", the substitution would be entered for the previous interjam. This user error is prevented here.
- There is no other reasonable spot to put either element.

html/controls/plt/plt-input.js
brian-brazil: Then the set jam shouldn't show when initially creating a penalty. And it should be jam/period.

html/controls/plt/plt-input.js
brian-brazil: No it's not, the button is not showing the state any more.

src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
brian-brazil: Put another way, run those unittests by hand and see if they do what you expect. The removal of that behaviour was incidental and was just the way that PR worked out, my goal was to fix the general broken box trip editor.

html/controls/plt/plt-input.js
frank-weinberg: It's not possible to always correctly remove the penalty: It could be that the trip was for two penalties of which only one was removed. Also if PT and LT are staffed separately, this would allow the LT to meddle with the PT sheet, which I don't think is a good idea.

Self report simply ends the box trip at the point you press it. The only case where a box trip would vanish is if a skater enters and leaves in the same interjam as that means that it never started.

brian-brazil: > It's not possible to always correctly remove the penalty: It could be that the trip was for two penalties of which only one was removed.

In either case it should remove them from the Box.

html/controls/sk/sk-sheet.js
frank-weinberg: If you removed the score, it still does submit that - the auto submit does not remove the trip (as that would disallow entering a new score to replace the removed one).

html/controls/sk/sk-sheet.js
frank-weinberg: What does it do and what did you expect?

src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
frank-weinberg: "Your code is a pile of shit" is not exactly motivating me to put any effort into following up. I will not do any work on this unless I get specific bug reports (that are reproducible with the changes from this PR applied). (And if I do change anything, it would be in a separate PR.)

html/controls/sk/sk-sheet.js
brian-brazil: Hmm, maybe require an explicit remove now?

html/controls/sk/sk-sheet.js
brian-brazil: If I click on a subsequent trip I expect the SP button to be green, but it isn't.

html/controls/lt/lt-sheet.js
frank-weinberg: It can't be n/a before the jam starts - the team still has the chance to field someone. Thus I've removed the button before Jam start.

html/controls/plt/plt-input.js
frank-weinberg: > In either case it should remove them from the Box.

It does.

html/controls/plt/plt-input.js
frank-weinberg: The display was correct. It was setting the value that was broken.

html/controls/plt/plt-input.js
frank-weinberg: Those are only defined for PBT. But we might as well use them here.

html/controls/plt/plt-input.js
brian-brazil: Hmm, can't reproduce now.

src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
brian-brazil: It's a complex thing to fix so doing it in another PR is fine, though this one should not be marked as fixing #269 as it doesn't.

frank-weinberg: Actually it does. #269 is about adding a skater and a box trip in the same go which was broken as the former was delayed and the latter wasn't. This PR makes adding Skaters immediate, so it addresses this specific scenario. If you have other scenarios, report them properly.

brian-brazil: https://github.com/rollerderby/scoreboard/issues/269#issuecomment-500606708 covers another. And the unittests cover the rest.

frank-weinberg: I don't have the faintest idea what that comment is talking about. But if you insist on keeping that thing open, so be it. I'd rather get this functionality in.

html/controls/plt/plt-input.js
brian-brazil: Can the Set Jam button be grouped more with the jam/period? As-is isn't not obvious that they're related. Maybe have Served/Clear on one row? For consistency, Clear should probably be Delete too.

html/controls/plt/plt-input.js
brian-brazil: Okay. Given you're probably on a tablet, typing in "Penalty Overturned" would be tricky. We should probably have buttons for those on the fielding editor too.

html/controls/plt/plt-input.js
brian-brazil: This still applies.

src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
brian-brazil: Create a penalty in the previous jam, advance the start time forward. It'll say Jam 0 at some point.

Considering it's a bugfix, can you add a unittest for this particular change? (Or leave it all for another PR)

brian-brazil:
Something else I noticed is that when a jam is ongoing and you click Box, the lineup symbols flicker to S before becoming -.

src/com/carolinarollergirls/scoreboard/core/impl/BoxTripImpl.java
frank-weinberg: Ok, I can reproduce that. I have an idea what the cause is, but that's definitely it's own PR.

Everything else should be addressed.

brian-brazil: That's just one form of it, my PR handles all of them so I'd suggest taking a look as I've done most of the leg work already.

html/controls/lt/lt-sheet.js
```diff
@@ -271,61 +263,81 @@ function prepareFieldingEditor(teamId) {
...
+		$('<td>').append($('<button>').attr('id', 'sitFor3').button().text('Sit out next 3').click(function() {
+			var check = $(this).attr('checked') == null;
+			$(this).attr('checked', check);
+			WS.Set(fieldingEditor.data('prefix')+'SitFor3', check);
+		})).appendTo(row);
+		
+		row = $('<tr>').addClass('Skater').appendTo(table);
+		$('<td>').attr('colspan', '3')
+			.append($('<input type="text">').attr('size', '35').attr('id', 'annotation'))
+			.append($('<button>').attr('id', 'setAnnotation').text('Set').button().click(function() {
```
brian-brazil: Do we really need this Set button? I'd expect this to be submitted automatically via an oninput handler.

brian-brazil (commented):
The S is still flickering when you press Box during a jam.

I also noticed that if you're between periods and the intermission clock has expired, you can't create new penalties - but this appears to be pre-existing.

frank-weinberg:
I've removed the set buttons. But the flickering S is an artefact of the fact that the EventListeners are run before `valueChanged`. Changing that order would be guaranteed to break stuff elsewhere in the software.

brian-brazil:
That sounds like the lack of a batch somewhere.

brian-brazil:
...and that looks like something I've already on my todo list.

brian-brazil:
There's still no indication on the LT sheet that a given fielding has an annotation.

When a skater is not in the box, and you click the ... the Penalty Overturned etc. aren't present.

frank-weinberg:
> That sounds like the lack of a batch somewhere.

This has nothing to do with batches.

> There's still no indication on the LT sheet that a given fielding has an annotation.

Now there is.

> When a skater is not in the box, and you click the ... the Penalty Overturned etc. aren't present.

Of course not. There's no penalty that could be overturned.

brian-brazil:
> Of course not. There's no penalty that could be overturned.

The skater may have already left the box, but you still want to add the annotation. I'd expect the interface to be the same whether they've left the box or not.

frank-weinberg:
The LT input interface is used to maintain the current game state. If you want to edit past events, you use the sheet part.

brian-brazil:
The current jam is current state though, and annotations are tied to fieldings not penalties.

frank-weinberg:
The semantics of the buttons is "End the current box trip and add the reason as an annotation." That is tied to the box trip.

brian-brazil:
Having one place where you can always press a button to add to the text, and another place where you sometimes can and sometimes can't is likely to cause confusion. For as long as these are tied to fieldings (which I think is the way to handle it) whether they're in the box or not shouldn't matter.

frank-weinberg:
Ok, fine. So now both sets of buttons are available if and only if there is a box trip associated with the fielding.

And since they were cumbersome to do in parallel, I also included the box trip fixes in here and even copied your unit tests (though I don't think checking the paperwork symbols is the best way to do them).

tests/com/carolinarollergirls/scoreboard/core/impl/FieldingImplTests.java
```diff
@@ -85,4 +86,52 @@ public void testSortBoxTripSymbols() {
...
+       t.field(s, Role.PIVOT);
+       Fielding f2 = s.getFielding(t.getRunningOrUpcomingTeamJam());
+       assertEquals(0, f2.getAll(Fielding.Child.BOX_TRIP).size());
+
+        // Box trip added to upcoming jam.
+       s.setPenaltyBox(true);
+       assertEquals(1, f1.getAll(Fielding.Child.BOX_TRIP).size());
+       assertEquals(1, f2.getAll(Fielding.Child.BOX_TRIP).size());
+
+        // And remed again.
```
brian-brazil: removed

brian-brazil (commented):
I was away the past week, so am only getting to this now.

I only see one bug from testing now, the red dot indicating an annotation doesn't show for the upcoming jam.

brian-brazil:
> though I don't think checking the paperwork symbols is the best way to do them

Feel free to do something better, though the tests for the symbols themselves were also pretty light.

frank-weinberg:
Annotation indicator for the upcoming jam has been added.

PR #351, opened 2019-07-24
Add unit tests for the event package

frank-weinberg:
This probably misses some special case behaviour that other classes rely upon but
should cover the most commonly used cases.

### DIFF
--- a/tests/com/carolinarollergirls/scoreboard/event/TestNumberedScoreBoardEventProviderImpl.java
+++ b/tests/com/carolinarollergirls/scoreboard/event/TestNumberedScoreBoardEventProviderImpl.java
@@ -1,0 +1,17 @@
+package com.carolinarollergirls.scoreboard.event;
+
+public class TestNumberedScoreBoardEventProviderImpl
+        extends NumberedScoreBoardEventProviderImpl<TestNumberedScoreBoardEventProvider> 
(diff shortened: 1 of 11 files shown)


tests/com/carolinarollergirls/scoreboard/event/TestNumberedScoreBoardEventProviderImpl.java
```diff
@@ -0,0 +1,17 @@
+package com.carolinarollergirls.scoreboard.event;
+
+public class TestNumberedScoreBoardEventProviderImpl
+        extends NumberedScoreBoardEventProviderImpl<TestNumberedScoreBoardEventProvider> 
+        implements TestNumberedScoreBoardEventProvider {
+    public TestNumberedScoreBoardEventProviderImpl(TestScoreBoardEventProvider parent, int number) {
+        super(parent, number, TestScoreBoardEventProvider.NChild.NUMBERED,
+                TestNumberedScoreBoardEventProvider.class, Value.class, Child.class);
+        if (parent instanceof TestNumberedScoreBoardEventProvider) {
```
brian-brazil: Isn't this always false?

frank-weinberg: It is. Fixed.

PR #357, opened 2019-08-01
Add SBO button behaviours anticipated by users

frank-weinberg:
Built on top of #336 because I'm not going to do another merge of Head into that one.


frank-weinberg:
This has been sitting almost two mmonths without any objection being raised. I'm going to merge it.

brian-brazil:
This is sitting around as it is based on #336, which is not resolved. The actual changes in the PR sound fine, the other ones it pulls in are not.

PR #358, opened 2019-09-28
Add a dialog to select the team on (P)LT screens

frank-weinberg:
Open the dialog if on loading no team is selected.

### DIFF
--- a/html/components/plt-input.js
+++ b/html/components/plt-input.js
@@ -25,7 +25,7 @@ function preparePltInputTable(element, teamId, mode, statsbookPeriod, alternateN
 		var thead = $('<tr>').appendTo($('<thead>').appendTo(table));
 		if (mode == 'plt' || mode == 'lt') {
 			if (mode == 'lt') {
-				$('<td>').attr('colspan', '5').attr('id', 'head').text('Team ' + teamId).appendTo(thead);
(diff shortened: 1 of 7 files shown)


html/components/plt-input.js
```diff
@@ -727,4 +727,41 @@ function prepareAnnotationEditor(teamId) {
...
+			} else {
+				optionsDialog.dialog('close');
+			}
+		})).appendTo(row);
+	});
+
+	WS.Register(['ScoreBoard.Team(*).Name', 'ScoreBoard.Team(*).AlternateName(operator)'], function(k, v) {
+		var displayName = 'Team ' + k.Team + ': ' + WS.state['ScoreBoard.Team('+k.Team+').Name'];
+		var altName = WS.state['ScoreBoard.Team('+k.Team+').AlternateName(operator)'];
+		if (altName != null) { displayName = displayName + ' / ' + altName; }
```
brian-brazil, line 755: We only ever display one name, so this should be the operator falling back to the team name as we have elsewhere.

brian-brazil (commented):
Hmm, something is making loads very slow. Probably the NIO change. I'll dig in when I have a chance.

html/components/plt-input.js
frank-weinberg: That was an intentional diversion from the usual behaviour due to a feature request on FB a while ago that pointed out a potential problem when operator names are colours and tablets from multiple tracks are picked up from the same spot.

PR #360, opened 2019-09-28
Revert changes from #336 that weren't agreed upon.

brian-brazil:
Undo just the contentious bit. I didn't put team before settings as we'd mentioned, but am happy to make that tweak.

I had a quick peek at the #337 changes (no point reviewing something that's dependant on a big PR that's not yet merged), and it looks okay. Only comment I'd have is that mentioning the team/position for the skater select in the title.

### DIFF
--- a/html/components/data-management-tab.css
+++ b/html/components/data-management-tab.css
@@ -1,14 +1,14 @@
-#DataManagementTab #saveLoad table { width: 100%; border-spacing: 1px 0px; }
-#DataManagementTab #saveLoad table tr>td { text-align: center; background: #eee; }
-#DataManagementTab #saveLoad table tr.Name>td { background: #ccc; border-radius: 12px 12px 0px 0px; }
-#DataManagementTab #saveLoad table tr.Content>td:first-of-type { border-radius: 0px 0px 0px 12px; }
-#DataManagementTab #saveLoad table tr.Content>td:last-of-type { border-radius: 0px 0px 12px 0px; }
-#DataManagementTab #saveLoad table tr.Content>td:only-of-type { border-radius: 0px 0px 12px 12px; }
+#saveLoad table { width: 100%; border-spacing: 1px 0px; }
+#saveLoad table tr>td { text-align: center; background: #eee; }
+#saveLoad table tr.Name>td { background: #ccc; border-radius: 12px 12px 0px 0px; }
+#saveLoad table tr.Content>td:first-of-type { border-radius: 0px 0px 0px 12px; }
+#saveLoad table tr.Content>td:last-of-type { border-radius: 0px 0px 12px 0px; }
+#saveLoad table tr.Content>td:only-of-type { border-radius: 0px 0px 12px 12px; }
 
-#DataManagementTab #GameData { width: 100%; height: 70vh; }
-#DataManagementTab #gameData table { background: #ddd; font-size: 125%; font-weight: bold; } 
-#DataManagementTab #gameData .Info { width: 90%; }
-#DataManagementTab #gameData .Reload { width: 10%; text-align: right; }
+#GameData { width: 100%; height: 70vh; }
+#gameData table { background: #ddd; font-size: 125%; font-weight: bold; } 
+#gameData .Info { width: 90%; }
+#gameData .Reload { width: 10%; text-align: right; }
 
 /* Tab divs */
 #DataManagementTab #tabsDiv>div { padding: 0px; }
--- a/html/components/data-management-tab.js
+++ b/html/components/data-management-tab.js
@@ -203,57 +203,58 @@ function createDataManagementTab(tab) {
 		
 	}
 
-	function createSaveLoadTab(tab) {
-		var table = $("<table>").attr("id", "SaveLoad").appendTo(tab);
-		
-		// Download table
-		var sbDownloadTable = $("<table>").addClass("Download")
-			.appendTo($("<td>").appendTo($("<tr>").appendTo(table)));
-		$("<tr>").addClass("Name").appendTo(sbDownloadTable)
-			.append("<td colspan='4'>Download ScoreBoard JSON</td>");
-		var contentRow = $("<tr>").addClass("Content").appendTo(sbDownloadTable);
-	
-		var links = [
-		{ name: "All data", url: "" },
-		{ name: "Teams", url: "teams.json?path=ScoreBoard.PreparedTeam" }
-		];
-		$.each( links, function() {
-			$("<td><a download/></td>").appendTo(contentRow)
-				.children("a").text(this.name).button()
-				.attr('href', '/SaveJSON/'+this.url);
-		});
-		var allDataA = contentRow.find(">td:eq(0)>a");
-		var updateAllUrl = function() {
-			var d = new Date();
-			var name = $.datepicker.formatDate("yy-mm-dd_", d);
-			name += _timeConversions.twoDigit(d.getHours());
-			name += _timeConversions.twoDigit(d.getMinutes());
-			name += _timeConversions.twoDigit(d.getSeconds());
-			allDataA.attr("href", "/SaveJSON/scoreboard-"+name+".json");
-		};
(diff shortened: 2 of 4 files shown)


frank-weinberg:
This isn't reverting the tab list to pre-336 state - pre 336 there was a "Save/Load" tab.

How about creating a second version of the operator screen that removes the tab-bar altogether instead?

brian-brazil:
True, it's been a while. Updated.

> How about creating a second version of the operator screen that removes the tab-bar altogether instead?

A typical random game is going to need the Teams page at least. I'd prefer not to have two operator screens too, unless we're going back to breaking everything out.

frank-weinberg:
> A typical random game is going to need the Teams page at least.

Not during the game. And your hypothetical inexperienced SBO will be just as confused by the Teams or Scoreboard Settings tab as they will be by the Rulesets tab. So if you want to cater to them, you have to remove all the tabs. And if you are catering to experienced users, you want everything that might be needed before, during, or after the game in one screen. What you propose here is combining the downsides of both approaches without either of the benefits.

Plus, the data you can download in the operator screen not being the data that can be used to generate a stats book is bound to generate a ton of support questions.

brian-brazil:
> Not during the game.

I wish, particularly at lower levels.

> So if you want to cater to them, you have to remove all the tabs.

I'm fine with that, this was your original proposal. I can make that change no problem.


What I'm doing in this PR currently is reverting us back to where we were, so we can discuss this further rather than making changes without agreement.

> Plus, the data you can download in the operator screen not being the data that can be used to generate a stats book is bound to generate a ton of support questions.

It's all the same data and format, the only difference is that the operator screen version will have some extra data like keyboard shortcuts. So it'll just work if you grab it once the score is final.

frank-weinberg:
> this was your original proposal

It was not.

Also, this PR introduces a violation to the component structure.

brian-brazil:
> It was not.

Hmm, yours had links on the top of the pages wasn't it?

> Also, this PR introduces a violation to the component structure.

A structure only introduced by your PR. I can remove the saving bit from the Up/Download page if you like, and I'm also happy to split the operator screen out into separate pages.

frank-weinberg:
My original idea (!= proposal) was to replace jQuery tabs with browser tabs. And I have explained at length, why I have come to the conclusion that this would introduce more downsides than upsides. It was never about reducing the number of tabs.

> A structure only introduced by your PR.

Which is still better established than the layering you occasionally claim violations to, which has never been either clearly spelled out or adhered to by existing code.

Historically, CRG has always had an operator screen that allowed all the tasks potentially needed on game day. Not including a tab for Ruleset editor and Media Management when they were introduced in 3.9.x was arguably a bug. #336 finally fixed that bug.

#336 also enables us to do a paradigm shift to an operator screen that only allows SBO in game data entry (in line with the other positional screens). But introducing this as the only option should only be done with community support. Until that support is established I'm fine with introducing it as an option, either as a separate link on the index page or by allowing the tab bar to be hidden in the screen.

But reverting to the buggy state because you don't believe in catering to any use cases beyond your own makes no sense at all.

brian-brazil:
What I'm against is a) merging changes without consensus and b) making the operator screen busier than it needs to be.

> But introducing this as the only option should only be done with community support.

I'm saying the same applies to your change. As it is we don't have a single issue requesting that rulesets and media be manageable via the operator screen. I don't find retroactively claiming that you were fixing a bug very convincing, particularly for a non-urgent change.

Let's go back to where we were UI wise, so that we can move on.

brian-brazil:
Also I just noticed that the rulesets page isn't rendering properly inside the operator screen (not surprising, it's an odd page in terms of implementation).

brian-brazil:
What do you propose we do here? To be clear merging of controversial changes without agreement is the main issue for me here, as that makes it difficult to work together.

frank-weinberg:
Well, refusing to merge a PR based on subjective criteria neither agreed-upon nor supported by anyone else in the community and then simply ceasing to discuss for months isn't exactly helping collaboration either.

And in light of that behaviour your demand to revert can be easily perceived as: "Put me back in a position where I can simply get what I want by doing nothing."

> I'm saying the same applies to your change.

I reintroduced a tab that has only been removed with 3.9.5 and extended another one. That's hardly comparable to removing tabs at all in terms of scope.

> What do you propose we do here?

I propose we add a per operator setting to show/hide the (full) tab bar (default: hidden), release 4.0 with documentation that points at the option of opening browser tabs for just those parts that you need and let the community form an opinion on the different options.

brian-brazil:
We've had many disagreements which we managed to resolve over the past year-ish, often on subjective topics. When things degrade into back and forth arguments, there's not much point in continuing discussion. Instead giving things some time to cool off is not a bad idea.

> nor supported by anyone else in the community

I don't believe this argument makes sense here, as I'm not aware of community support for your change either.

If the community was clamouring for something and I was saying no, then you'd have a case. Otherwise "the community didn't really say anything either way, so I can go ahead" is a general argument that'd justify any change.

> And in light of that behaviour your demand to revert can be easily perceived as

That's not what I'm saying. It's lets build consensus, rather than forcing through changes for which a developer can't get others to agree are a good idea.

> I propose

What I meant is what we do governance wise? A situation where maintainers can't build consensus and then merge what they like is not ultimately sustainable, so some ground rules are required.

frank-weinberg:
> so some ground rules are required.

How about: 

A PR that introduces new functionality is only not merged if it
- violates coding standards, UI principles, or similar rules that the existing code adheres to,
- breaks or notably complicates an existing use case, or
- goes against a consensus that has been established amongst developers or in the FB groups

and it does not get more developer votes saying the benefits outweigh these problems than votes opposing it. An objection has to clearly name the rule/use case/consensus that stands against the PR.

A submitter may (only) merge their own PR if another developer approved it and noone oppesed or if it sat for a full month without any objection.

brian-brazil:
Expressing such things negatively is a bit odd, I've never come across a governance like that. It's also quite subjective, would lead to scope creep, and makes it more challenging to gradually improve code health.

Usually it's something about lazy concensus, and some form of conflict resolution rules when that doesn't work. The general presumption is that the person who wants to make a change has to argue for it, not that everyone else has to argue against it. No is temporary, yes is forever.

frank-weinberg:
So you're proposing a policy that supports your stance in the underlying discussion, while I propose one that supports mine. Surprise.

And after we have established that the meta discussion leads to the same deadlock, I'd rather return to the original discussion. I believe #367 adds a strong argument in favor of having the ruleset editor as a tab, as it now supports modifying the behaviour of the scoreboard for the game in progress.

brian-brazil:
> And after we have established that the meta discussion leads to the same deadlock

This needs to be handled first, which is why I brought it up. We need to have an agreed way to resolve disagreements rather than letting them fester.

The standard way of handling things when lazy consensus doesn't suffice is voting, with a majority  being required.

frank-weinberg:
>This needs to be handled first

Why? If we were to agree on a compromise regarding the specific issue, we could then look into the meta discussion without it being bogged down by that additional baggage.

> The standard way of handling things when lazy consensus doesn't suffice is voting

... which has historically proven a rather ineffective way of resolving a conflict involving only two parties with equal votes.

brian-brazil:
> Why? If we were to agree on a compromise regarding the specific issue, we could then look into the meta discussion without it being bogged down by that additional baggage.

We could do that. Having the prospect of controversial (non-urgent) PRs being merged off the table would be nice.

In this particular case I note that it was common practice to tell new users to completely ignore the existence of the policy page, as they often got confused by and asked questions in the user group. I'm also not sure I've ever heard of save/load being used outside of entering teams in big events.

> ... which has historically proven a rather ineffective way of resolving a conflict involving only two parties with equal votes.

Indeed. Thus why I suggest consensus building, I'd rather not have to get into a big formal governance for what is ultimately a small (but disproportionately used) hobby project.


I presume we're planning on using whatever latest master is for Sunday's games? I'll have my laptop with me anyway.

brian-brazil:
A way forward has been agreed, so closing this.

PR #361, opened 2019-09-28
Fix a pile of warnings Java is complaining about.

brian-brazil:


src/com/carolinarollergirls/scoreboard/utils/ClockConversion.java
```diff
@@ -17,11 +17,11 @@ public static Long fromHumanReadable(String v) {
                 par = Long.parseLong(parPad.substring(0, 3));
             }
 
-            return new Long(((sec + (min * 60)) * 1000) + par);
+            return Long.valueOf(((sec + (min * 60)) * 1000) + par);
         }
 
         try {
-            return new Long(Long.parseLong(v));
+            return Long.valueOf(Long.parseLong(v));
```
frank-weinberg: valueOf() can directly take a String argument.
And I think many of the other instances can be removed due to autoboxing.

brian-brazil: Updated. I'd done them in batch.

We should really get off the hand-edited JSON library...

frank-weinberg: > We should really get off the hand-edited JSON library...

As we should probably update to upstream supported versions of most of our dependencies.

brian-brazil: Indeed, it's somewhere on my todo list.

PR #363, opened 2019-10-01
Switch fully to jackson-jr for our JSON needs.

brian-brazil:
Remove now unused json.org library.


frank-weinberg:
I have added some cleanup - doing it myself was way faster and less error prone than commenting on all those places. Feel free to merge, if you have no objections to those changes.

brian-brazil:
Thanks

PR #367, opened 2019-10-02
Allow the Current Ruleset to be changed

frank-weinberg:
This allows e.g. changing intermission duration or en/disabling auto start/stop rules without a reset of the teams.

### DIFF
--- a/html/components/rulesets-tab.js
+++ b/html/components/rulesets-tab.js
@@ -258,15 +258,15 @@ function createRulesetsTab(tab) {
 		});
 
 		if (rs.Immutable) {
-			tab.find("#name").attr('readonly', true);
+			tab.find("#name").attr('disabled', true);
 			definitions.find(".definition *").prop("disabled", rs.Immutable);
 			definitions.find(".Update, .Delete, .EditNote").hide();
 		} else if (rs.Effective) {
-			tab.find("#name").removeAttr('readonly');
+			tab.find("#name").removeAttr('disabled');
 			definitions.find(".Update, .EditNote").show();
 			definitions.find(".Delete").hide();
 		} else {
-			tab.find("#name").removeAttr('readonly');
+			tab.find("#name").removeAttr('disabled');
 			definitions.find(".Update, .Delete").show();
 			definitions.find(".EditNote").hide();
 		}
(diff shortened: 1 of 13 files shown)


src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
```diff
@@ -62,6 +63,15 @@ else if (flag != Flag.INTERNAL) {
         if (prop == Value.INVERTED_TIME) {
             return getMaximumTime() - getTime();
         }
+        if (prop == Value.MAXIMUM_TIME && getId() == ID_INTERMISSION) {
+            long duration = DEFAULT_MAXIMUM_TIME;
```
brian-brazil: This means I can no longer override the maximum time for an intermissions from the operator screen, which is often required when the schedule needs moving forward.

brian-brazil:
That updating/deleting rulesets that are in use will affect the in-progress game may not be completely obvious to users.

frank-weinberg:
I have removed the delete button (because it's completely unclear what should be done if the active ruleset is deleted) and added a note that editing will affect the current game for rulesets currently in effect.
(I've also changed handling of max intermission time so manual overrides are possible again.)

brian-brazil:
That works.

A few small issues:
- Changing the current ruleset dosen't update the Delete button/warning
- If you create a ruleset that's only whitespace, things get odd
- The Name of "WFTDA Sanctioned" shouldn't be in a textbox, as it's not editable

I suspect the last two are existing.

frank-weinberg:
They were existing but sufficiently easy fixes/workarounds.

html/components/rulesets-tab.js
```diff
@@ -261,6 +302,9 @@ function createRulesetsTab(tab) {
 				}
 				WS.Set("ScoreBoard.Rulesets.Ruleset(" + activeRuleset.Id + ").Rule(" + val.Fullname + ")", value);
 			});
+			var newName = tab.find("#name").val();
+			if (newName.trim() === '') { newName = 'Unnamed'; }
```
brian-brazil: Elsewhere (colors) we ignore such requests, I think we should do the same here.

html/components/rulesets-tab.js
```diff
@@ -225,10 +258,17 @@ function createRulesetsTab(tab) {
 		});
 
 		if (rs.Immutable) {
+			tab.find("#name").attr('readonly', true);
```
brian-brazil: I think disabled would be a bit clearer

brian-brazil (commented):
Grand, saves me from fixing them.

frank-weinberg:
Addressed the last comments and fixed an NPE discovered in the process.

PR #368, opened 2019-10-13
Turn "Auto End Jam" off by default

frank-weinberg:
Rationale:

Auto ending adds additional problems if the JT misses calling of the Jam at 2:00, as the lineup clock will have run up prematurely. It also makes it harder to notice at a glance that the jam clock has run down, as there is a non-zero value displayed in the spot of the jam clock. (With the change, the jam clock will idle at 0 until "Stop Jam" is pressed.)

There is also little advantage to having the automation - it is easier for the SBO to just react to the calloff whistles by reflex than to try to remember to skip that button press on 2 minute jams.

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
@@ -308,12 +308,11 @@ public class ScoreBoardImpl extends ScoreBoardEventProviderImpl implements Score
     @Override
     public void stopJamTO() {
         synchronized (coreLock) {
-            Clock jc = getClock(Clock.ID_JAM);
             Clock lc = getClock(Clock.ID_LINEUP);
             Clock tc = getClock(Clock.ID_TIMEOUT);
             Clock ic = getClock(Clock.ID_INTERMISSION);
 
-            if (jc.isRunning()) {
+            if (isInJam()) {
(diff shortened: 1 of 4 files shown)


brian-brazil:
I think it'd cause a lot of confusion to do this, this behaviour has been there a long time and SBOs likely rely on it.

bullseye555:
As an SBO, I'm with Brian on this.

It's so ingrained that a 2min jam will automatically end that you don't
think to react to it

frank-weinberg:
I'm an SBO myself and I've never treated 2min jams special, so this is not universally ingrained.

But this PR was very much intended as a "What do you think of it?" and the reactions so far are quite clear.

JeneralPain:
I am against removing 'AUTO STOP', it's expected that at 2min the jam
ends... it should NEVER continue. There is no Casebook or rule that
justifies it continuing.

frank-weinberg:
> it should NEVER continue

Yes. But the Jam isn't ended by the scoreboard switching to lineup, it's ended by (in this case) the JT whistling it off. Jam Timers are human. Humans make mistakes. I've seen jams longer than 2 minutes even at Playoffs. If the scoreboard goes to lineup but the JT only whistles several seconds later, the effects of the error are worsened because now the lineup clock doesn't correlate to the time that actually passed since the Jam ended (which could e.g. affect a team's decision to call or not call a timeout).

PR #373, opened 2019-10-22
Auto-mark some more cases

frank-weinberg:
Mark Lost
- when a Jammer completes their initial trip and isn't declared lead
despite being eligible.
- when a Penalty is recorded for an eligible or Lead Jammer.

Make Calloff and Injury mutually exclusive when entered through the
operator controls. (In the rare case where both should be marked, this
can be done on the SK sheet.)

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/Jam.java
+++ b/src/com/carolinarollergirls/scoreboard/core/Jam.java
@@ -10,6 +10,8 @@ import com.carolinarollergirls.scoreboard.event.ScoreBoardEventProvider;
 public interface Jam extends NumberedScoreBoardEventProvider<Jam> {
     public void setParent(ScoreBoardEventProvider p);
 
+    public boolean isOvertimeJam();
+    
     public long getDuration();
     public long getPeriodClockElapsedStart();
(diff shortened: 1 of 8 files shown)


JeneralPain:
You don’t mark lost for lead eligible when you don’t get lead.

Lost has two cases:
1. Eligible for lead and LOST eligibility due to a penalty

2. Had lead and LOST due to penalty.

frank-weinberg:
The rules disagree:

> A Jammer becomes ineligible to earn Lead during a Jam if they commit a penalty during that Jam, **exit the front of the Engagement Zone without having earned Lead**, remove their helmet cover, or have their helmet cover removed by a teammate.

(emphasis mine)

See also the Statsbook Manual:

> Situations where “Lost” should be marked: [...] Jammer exits the Engagement Zone at the end of the Initial Trip and the Jammer Referee signals not Lead, even though Lead has not yet been declared.

brian-brazil:
These should be predicated on it not being an overtime jam, and these should have unittests.

frank-weinberg:
I have gone with blocking lost in overtime jams in general instead of adding an extra condition to every item that sets it.

PR #374, opened 2019-10-22
Make Tab Bar Hidable

frank-weinberg:


brian-brazil:
The initial state's not right.

frank-weinberg:
Turns out that initializing operator settings was severly broken and we just didn't notice as all defaulted to false.

brian-brazil:
That's still not right, plus every 3rd click flickers instead of applying.

frank-weinberg:
should work now.

PR #375, opened 2019-10-23
Show all skater numbers for jammer/pivot selection.

brian-brazil:
No more dropdowns to get in the way of keyboard shortcuts,
plus it should be faster to use.

I think this works nicely, though this bit of the screen might need a small reorg now. These don't support key controls.

Screenshot_2019-10-23_18-57-46

### DIFF
--- a/html/components/team-time-tab.css
+++ b/html/components/team-time-tab.css
@@ -47,6 +47,12 @@
 #TeamTime table.Team tr.Jammer td.ui-buttonset label.Lead.ui-state-active { background: #3f3; }
 #TeamTime table.Team tr.Jammer td.ui-buttonset label.StarPass.ui-state-active { background: #3f3; }
 
+#TeamTime .skaterSelector>button { border-radius: 0; }
+#TeamTime .skaterSelector>button:first-of-type { border-radius: 12px 0px 0px 12px; border-top: none; }
+#TeamTime .skaterSelector>button:last-of-type { border-radius: 0px 12px 12px 0px; }
+#TeamTime .skaterSelector>button:only-of-type { border-radius: 12px; }
+#TeamTime .skaterSelector>button>span { padding-top: 0.3em; padding-bottom: 0.3em; }
+
 #TeamTime table.Team button.Box.Active { color: #000; background: #F33; }
 #TeamTime table.Team button.Active:not(.Box) { color: #000; background: #3F3; }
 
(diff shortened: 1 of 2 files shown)


frank-weinberg:
The selector area is already crowded on this wide screen. On a smaler screeen, this will be too large. How about this layout:

Screenshot_20191026_111531

On half the width this still looks reasonable:

Screenshot_20191026_111243

> These don't support key controls.

Why? Not having to take the hand off the keyboard would be very useful.

brian-brazil:
> How about this layout:

That works. If all skater numbers are 4 digit on a widescreen the Box wraps, so that seems fine.

> Why? Not having to take the hand off the keyboard would be very useful.

Unlike every other control these are dynamic, so in the best case the semantics would be non-obvious to users (is it tied to a skater? a number? an index in the list?).

frank-weinberg:
> in the best case the semantics would be non-obvious to users (is it tied to a skater? a number? an index in the list?).

Since the key assignments are displayed on the buttons, this will be easy enough to figure out. The people who are the audience for this feature will happily learn that. And those that don't want to use the feature don't need to care about it.

What I'd do is tie the key to the skater and drop it if the skater is removed from the roster or the team is replaced. That happens to also be the easiest option to implement.

brian-brazil:
> Since the key assignments are displayed on the buttons, this will be easy enough to figure out.

Hmm, I wonder how that'd affect wrapping.

> What I'd do is tie the key to the skater and drop it if the skater is removed from the roster or the team is replaced.

That'd drop more than you'd expect, such as if a skater's name/number changes (button is deleted and re-created), or a skater is out for 3 jams. This is more complicated than it might seem.

frank-weinberg:
> Hmm, I wonder how that'd affect wrapping.

For every given team roster there will be a window size where adding in a key assignment will cause it to wrap. And a slightly smaller one where it will wrap anyways. Since many people are not running the operator screen in fullscreen, speculating about window sizes is moot. We should just make sure that the screen doesn't become unusable on a reasonable range of window sizes.

> That'd drop more than you'd expect, such as if a skater's name/number changes (button is deleted and re-created),

I've seen the delete-recreate part. It should be possible to read the key before and reassign after.

>  or a skater is out for 3 jams.

Out for 3 jams does not remove a skater from the list.

> This is more complicated than it might seem.

I'm fine with implementing it myself as long as I don't have to expect a block based on "I don't want that feature".

brian-brazil:
> I'm fine with implementing it myself as long as I don't have to expect a block based on "I don't want that feature".

That works for me.

PR #376, opened 2019-10-25
Fix writing game data files

frank-weinberg:
- Trying to overwrite the file by renaming the temp file fails on some systems.
- Force using UTF-8 for writing the file
- Detect jam end by ScoreBoard.InJam

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
+++ b/src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
@@ -3,7 +3,9 @@ package com.carolinarollergirls.scoreboard.json;
 import io.prometheus.client.Histogram;
 
 import java.io.File;
-import java.io.FileWriter;
+import java.io.FileOutputStream;
+import java.io.OutputStreamWriter;
+import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.util.Iterator;
 import java.util.Map;
@@ -25,14 +27,14 @@ public class JSONStateSnapshotter implements JSONStateListener {
     public synchronized void sendUpdates(Map<String, Object> state, Set<String> changed) {
         if (state.get("ScoreBoard.CurrentPeriodNumber") != "0") {
             // If the jam has just ended or the score is now official, write out a file.
-            if ((inJam && !bool(state.get("ScoreBoard.Clock(Jam).Running")))
+            if ((inJam && !bool(state.get("ScoreBoard.InJam")))
                     || (!officialScore && bool(state.get("ScoreBoard.OfficialScore")))) {
                 writeFile(state);
             }
         }
 
         officialScore = bool(state.get("ScoreBoard.OfficialScore"));
-        inJam = bool(state.get("ScoreBoard.Clock(Jam).Running"));
+        inJam = bool(state.get("ScoreBoard.InJam"));
     }
 
     public void writeFile(Map<String, Object> state) {
@@ -49,10 +51,10 @@ public class JSONStateSnapshotter implements JSONStateListener {
         String name = new SimpleDateFormat("yyyy-MM-dd HH_mm_ss").format(startTime)
         + " - " + state.get("ScoreBoard.Team(1).Name")
         + " vs " + state.get("ScoreBoard.Team(2).Name")
-        + " P" + period
-        + ".json";
+        + " P" + period;
         name = name.replaceAll("[^a-zA-Z0-9\\.\\- ]", "_");
-        File file = new File(new File(directory, "html/game-data/json"), name);
+        File file = new File(new File(directory, "html/game-data/json"), name + ".json");
+        File prev = new File(new File(directory, "html/game-data/json"), name + "_prev.json");
         file.getParentFile().mkdirs();
 
         // The state includes secrets (Twitter auth) and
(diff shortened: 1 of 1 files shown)


brian-brazil:
> Trying to overwrite the file by renaming the temp file fails on some systems.

Can you give more details of when this happens, and which systems?

frank-weinberg:
I observed it on Windows 7. Whenever the target file already exists, the call to `renameTo()` will fail. As the javadoc for that function suggests could happen:

> Many aspects of the behavior of this method are inherently platform-dependent: The rename 
 operation might not be able to move a file from one filesystem to another, it might not be 
 atomic, and it might not succeed if a file with the destination abstract pathname already exists. 
 The return value should always be checked to make sure that the rename operation was 
 successful.

(The temp file would still be deleted in the `finally` clause.)

brian-brazil:
Hmm, it should work on Windows as long as the file isn't open. Did you have it open in another process, or are we not closing it somewhere?

frank-weinberg:
I did not do anything but run the scoreboard. And, as cited above, the documentation of the function says that it's not guaranteed to work on all file systems, so even if it should have worked on my system in theory, we should still switch to the proposed implementation as that is more robust in general.

brian-brazil:
This code has the exact same potential problems as the current version.

This is a problem we run into every now and then on the main project I work
on, it's always due to the file being opened by something.

frank-weinberg:
> This code has the exact same potential problems as the current version.

After yesterday's game I can add a second computer (Win 10, Java downloaded and installed on game day) where current HEAD failed to update the file and the code from this PR worked. And the new code uses the same approach that has proven itself for years with the autosave files. So the data doesn't support your hypothesis at all.

brian-brazil:
> So the data doesn't support your hypothesis at all.

This is not a hypothesis, this is the well known behaviour. What we're doing here only makes whatever race take longer, as the file is getting deleted a bit later. We need to fix whatever fd leak is happening.

Do you have a traceback before this one? Do you known if a GC is occurring between autosaves?

frank-weinberg:
> as the file is getting deleted a bit later

This isn't about deletion. Deletion works fine. This is about `renameTo()` not being able to overwrite files on Windows. Which is in line with the documentation and - as a quick search shows - consistently observed by a lot of people.

brian-brazil:
Ah, I'm used to Go which handles this under the covers. Looks like https://stackoverflow.com/a/10532430/83591 is how to handle it.

frank-weinberg:
Except the first comment on that answer cites the documentation stating that the two options cannot be combined. And without an atomic replacing rename the original code is vulnerable to having no valid file, if a crash happens during the operation. The code in the PR guarantees that there always is at least one valid file on the disk.

brian-brazil:
> Except the first comment on that answer cites the documentation stating that the two options cannot be combined.

But that it works in practice. If this wasn't workable at the syscall level, I'd have heard about it.

frank-weinberg:
What this PR does is exactly what we have been doing successfully for the autosaves all along. I don't see the need to bring in another way.

brian-brazil:
My aim here was standard atomic replacement of files, what you're proposing is a more complicated way of doing that which takes more thinking to verify correctness. It's also not the same control flow as autosaves.

frank-weinberg:
> My aim here was standard atomic replacement of files

Which, according to the documentation, we won't get reliably in Java.

> what you're proposing is a more complicated way

So far it's the simplest proposed way that reliably gives the desired outcome.

> It's also not the same control flow as autosaves.

Both delete the last file in the chain using `delete()`, then move the others up one spot using `renameTo()`. So regarding the relevant operations, they work the same way.

brian-brazil:
> Which, according to the documentation, we won't get reliably in Java.

We will on all platforms we care about. I'm not aware of a platform (which Go supports - which is ~everything we care about) where there is not an atomic rename.

> So regarding the relevant operations, they work the same way.

You're doing stuff with prev in a different order than the other code.

frank-weinberg:
> We will on all platforms we care about. 

We don't know what os/file system combos people will use CRG on. And given that we're talking about a potential data loss and have an alternative that does not bring this risk, I don't see why we should not use that alternative.

> You're doing stuff with prev in a different order than the other code.

Huh? It's remove prev, rename next file to prev, repeat. There isn't even a theoretical option to loop two operations in a different order.

brian-brazil:
> We don't know what os/file system combos people will use CRG on.

Which kernel beyond OSX, Linux, Windows, NetBSD, FreeBSD, OpenBSD, and Dragonfly are you worried about?

> And given that we're talking about a potential data loss and have an alternative that does not bring this risk, I don't see why we should not use that alternative.

It's quite a non-standard way to do things, whereas what I'm proposing is typical.

frank-weinberg:
> Which kernel

It's not just kernels it's also file systems and Java implementation. That's a lot of possible combinations to stay on top of.

> whereas what I'm proposing is typical.

We have already worked out that what you're proposing comes with a remaining risk. I'm not willing to take that risk just to satisfy your sense of code aesthetics.

brian-brazil:
> It's not just kernels it's also file systems and Java implementation.

If you're running on NFS you've already lost - and your code is likely more susceptible to that set of issues.

> I'm not willing to take that risk just to satisfy your sense of code aesthetics.

This isn't about code, this is about semantics and engineering. It's confusing that you're doing this in a roundabout way, rather than the typical way for a simple use case. That's a smell, and hint that there's an unhandled corner case in there somewhere as it's quite rare that someone manages to implement something bespoke like this without a bug - and I believe there is indeed a bug or at least dependence on unspecified behaviour in this one. Can you spot it?

frank-weinberg:
I don't think this kind of arrogance is called for. Even more so given that this PR is about fixing a data loss bug that you introduced.

brian-brazil:
This isn't worth extended arguing about, it'll do for now.

PR #378, opened 2019-10-30
Allow keys to be assigned to Jammer/Pivot buttons

frank-weinberg:
Key assignments are tied to SkaterId and Team 1 or 2 and will persist if the team is
reassigned to the same side or if the Skater is temporarily removed
from the roster.

### DIFF
--- a/html/components/team-time-tab.js
+++ b/html/components/team-time-tab.js
@@ -712,10 +712,6 @@ function createTeamTable() {
 								WS.Set(prefix + '.Position('+pos+').Skater', k.Skater);
 						}).button();
 						_crgKeyControls.setupKeyControl(button, _crgKeyControls.operator);
-						var key = WS.state[button.attr('_crgKeyControls_prop')];
-						button.toggleClass('HasControlKey', (key?true:false))
-							.find('span.Key').text(key?key:'')
-							.attr('data-keycontrol', String(key?key.charCodeAt(0):''));
 					_windowFunctions.appendAlphaSortedByAttr(container, button, 'number', 1);
 				}
 				setValue(WS.state[prefix + '.Position('+pos+').Skater']);
--- a/html/components/team-time-tab.js
+++ b/html/components/team-time-tab.js
@@ -693,7 +693,7 @@ function createTeamTable() {
 		var makeSkaterSelector = function(pos) {
 			var container = $('<span class="skaterSelector">')
 
-			var none = $('<button>').text("?").attr("skater", "").button();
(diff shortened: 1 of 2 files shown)


html/components/team-time-tab.js
```diff
@@ -706,10 +706,16 @@ function createTeamTable() {
...
+					var button = $('<button>').attr('number', number).attr('skater', k.Skater)
+							.attr('id', 'Team'+team+pos+k.Skater).addClass('KeyControl').text(number)
 							.click(function() {
-						WS.Set(prefix + '.Position('+pos+').Skater', k.Skater);
-					}).button();
+								WS.Set(prefix + '.Position('+pos+').Skater', k.Skater);
+						}).button();
+						_crgKeyControls.setupKeyControl(button, _crgKeyControls.operator);
+						var key = WS.state[button.attr('_crgKeyControls_prop')];
+						button.toggleClass('HasControlKey', (key?true:false))
```
brian-brazil: Why do you need this? setupKeyControl should already have us covered here.

brian-brazil (commented):
I believe this will be the last UI change before v4?

I also noticed that we're not excluding the bench coach from this list.

html/components/team-time-tab.js
frank-weinberg: It didn't work in all cases. (IIRC when the button was removed and readded the key wasn't activated, while it did work when reloading the screen.) 

> I believe this will be the last UI change before v4?

I don't have anything else in the pipeline.

> I also noticed that we're not excluding the bench coach from this list.

Can add that tonight.

brian-brazil: I think we'd be better off fixing this over in setupKeyControl, rather than spreading the key control code around.

frank-weinberg: I was shying away from touching this code to avoid introducing some subtle bug shortly before release. But before this turns into another tug of war...

> I also noticed that we're not excluding the bench coach from this list.

Actually, we are. Bench coach is mapped to NotInGame.

brian-brazil: > I was shying away from touching this code to avoid introducing some subtle bug shortly before release.

I'd be more worried about longer term maintenance, duplicating code like that is asking for one copy not to be updated down the line. No obvious issues from a quick test, and I'll be using this at a tournament and it'll get a full run through when I'm doing the videos.

> Actually, we are. Bench coach is mapped to NotInGame.

So it is, it looks like it doesn't interact too well with such a skater being Jammer/Pivot when their state is changed.

PR #379, opened 2019-11-01
Re-Field Skaters penalized between Jams

frank-weinberg:


src/com/carolinarollergirls/scoreboard/core/impl/SkaterImpl.java
```diff
@@ -113,16 +113,23 @@ public ValueWithId create(AddRemoveProperty prop, String id) {
...
-        } else if (prop == Child.FIELDING && ((Fielding)item).isCurrent()) {
-            set(Value.CURRENT_FIELDING, item, Flag.INTERNAL);
-        } else if (prop == Child.FIELDING && team.hasFieldingAdvancePending() &&
-                ((Fielding)item).getTeamJam().isRunningOrEnded()) {
-            setRole(((Fielding)item).getCurrentRole());
+        if (prop == NChild.PENALTY) {
+            Penalty p = (Penalty)item;
+            if (FO_EXP_ID.equals(p.getProviderId())) {
+                updateEligibility();
+            } else if (!p.isServed() && getRole() == Role.JAMMER && getCurrentFielding() != null
```
brian-brazil: This is making it not work for Jammers

frank-weinberg: This should not fire between jams - fixed.

PR #381, opened 2019-11-04
Allow inserting a Jam between last and upcoming

frank-weinberg:
If the SBO realizes a missing Jam before the subsequent one has started.

### DIFF
--- a/html/components/team-time-tab.js
+++ b/html/components/team-time-tab.js
@@ -893,6 +893,8 @@ function createPeriodDialog() {
 }
 
 function createJamDialog() {
+	var currentPeriod;
+	var nextJam;
 	var dialog = $("<div>").addClass("NumberDialog");
 	var tableTemplate = $("<table>").addClass("Period");
 	var headers = $("<tr><td/><td/><td/><td/><td/><td/></tr>").appendTo(tableTemplate);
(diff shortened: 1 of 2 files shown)


html/components/team-time-tab.js
```diff
@@ -904,7 +906,13 @@ function createJamDialog() {
 		.appendTo(headers.children("td:eq(2)").addClass("Title"));
 	$("<a>").text("PC at end").addClass("Title")
 		.appendTo(headers.children("td:eq(3)").addClass("Title"));
-	var currentPeriod;
+	var footer = $('<tr><td colspan="4"></td><td></td><td></td>')
+		.addClass("Jam").attr("nr", 999).appendTo(tableTemplate);
+	$('<span>').text('Upcoming').appendTo(footer.children('td:eq(0)'));
+	$('<button>').text('Insert Before').button().click(function () {
+		console.log('ScoreBoard.Jam(' + nextJam + ').InsertBefore');
```
brian-brazil: Debug code left behind?

html/components/team-time-tab.js
```diff
@@ -981,7 +989,11 @@ function createJamDialog() {
 		currentPeriod = v;
 		dialog.find("table.Period.Show").removeClass("Show");
 		dialog.find("table.Period[nr="+v+"]").addClass("Show");
-	})
+	});
+	
+	WS.Register(['ScoreBoard.Jam(*).Number'], function(k, v) {
+		nextJam = v;
```
brian-brazil: Hmm, is this safe? It's going to get deleted jams wrong at least.

frank-weinberg: `ScoreBoard.Jam()` is the upcoming Jam, so this will not be triggered by any changes to past jams (unless they change the number of the upcoming Jam).

But a guard for deletion makes sense - added (and removed the debug code).

PR #383, opened 2019-11-08
Make scoreboard-Linux work better on OSX, and rename.

brian-brazil:
Remove the OSX app, this will replace it and work on both 32/64bit systems.
Rename to scoreboard.sh.

GeneralPain has verified this on a Mac.


JeneralPain:
I'm fine with this, I know it works because I've tested it.

PR #384, opened 2019-11-10
Add a scoreboard setting for displaying the logo.

brian-brazil:
This allows for not using the logo on the scoreboard
to make the team name more readable, while leaving it
accessible for the overlay.


frank-weinberg (commented):
When the setting is not set (e.g. starting from an older autosave), the button has no text.

brian-brazil:
Fixed down in the library.

PR #386, opened 2019-11-11
Store all parts of the release info.

brian-brazil:
This will help track things down if
someone uses an unusual build.

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/utils/Version.java
+++ b/src/com/carolinarollergirls/scoreboard/utils/Version.java
@@ -3,6 +3,9 @@ package com.carolinarollergirls.scoreboard.utils;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Properties;
+import java.util.Collections;
+import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


src/com/carolinarollergirls/scoreboard/utils/Version.java
```diff
@@ -19,14 +22,19 @@ public static boolean load() throws IOException {
...
+        for (String k : versionProperties.stringPropertyNames()) {
+          m.put(k, versionProperties.getProperty(k));
+        }
+        props = Collections.unmodifiableMap(m);
         Logger.printMessage("Carolina Rollergirls Scoreboard version "+Version.get());
         return true;
     }
 
-    public static String get() { return versionRelease; }
+    public static String get() { return props.getOrDefault(VERSION_RELEASE_KEY, "unset"); }
```
frank-weinberg: `getOrDefault` is not available in Java 7.

brian-brazil: Fixed.

PR #391, opened 2019-11-21
Don't show penalty count on LT screen

brian-brazil:


frank-weinberg:
They are intentionally on there, being commonly tracked by LTs.

JeneralPain:
Agree, the totality of the penalties ... ie "310 (6)" for example, you know
they're on 6 or something?

but I wouldn't remove it.

brian-brazil:
PLTs track penalties, LTs don't - and there's no way on this screen for them to alter them.

frank-weinberg:
It's pretty standard for (paper) LTs to keep a tally of Penalties for quick cross-checking with PT/PBT. 
When people are running paper backup parallel to CRG (which I expect to happen relatively regularly early on until trust in the software has been built), having the counts on the screen allows these cross-checks to be done locally.

And for pure CRG-setups, having the info on the screen at the very least allows LTs to check the plausibility of their box trip count on their own.

And it's not as if this screen area is overcrowded with columns.

brian-brazil:
> It's pretty standard for (paper) LTs to keep a tally of Penalties for quick cross-checking with PT/PBT.

I've not seen this as a common thing, and there's nothing related to this in the statsbook manual.

This is overall confusing, as the only things an LT could be summing based on their paperwork are box trips - which are not the same as penalties. It's basically only useful in the context of PT or PBT paperwork.

> And for pure CRG-setups, having the info on the screen at the very least allows LTs to check the plausibility of their box trip count on their own.

We don't expose box trip counts anywhere, so I don't see how this would work. This also presumes that there's ePT.

> And it's not as if this screen area is overcrowded with columns.

That's not a reason to add information not directly relevant to the role. We should be aiming to keep things as minimal as practical, as devices don't have infinitely wide screens.

Particularly if someone is using this in portrait view, which I'd expect to be usually the case for this screen.

frank-weinberg:
> I've not seen this as a common thing, and there's nothing related to this in the statsbook manual.

I see this being told even to rookie LTs and have seen it being done at any level of tournament I've been to.

> box trips - which are not the same as penalties

Which is why it's a plausibility check - if the numbers are slightly off, you can check if there was a double penalty or a self-report that explains it, otherwise you knowthat a detailed cross check is advisable.

> That's not a reason to add information not directly relevant to the role.

Like the current lineup on the PT screen? They can't change it, they don't track it, so why should they see it?
Or having the box buttons on the SBO screen?

> We should be aiming to keep things as minimal as practical, as devices don't have infinitely wide screens. 

Even at 400px screen width the cells are still wider than they are high, so adding extra width is unlikely to make them notably easier to hit. (And checking it on my 4" smartphone the screen is barely usable in portrait mode due to row height; cell width is still perfectly fine.)

I'm not claiming that there is an important use case for this, but there is one. And as long as we don't have a better use for the screen space, I'll be blocking it's removal.

brian-brazil:
> Which is why it's a plausibility check - if the numbers are slightly off, you can check if there was a double penalty or a self-report that explains it, otherwise you knowthat a detailed cross check is advisable.

How would you check that though? There's no paper to do tracking/logging/checkpointing on. That's quite a stretch once you get beyond a handful of penalties total.

> Like the current lineup on the PT screen? They can't change it, they don't track it, so why should they see it?

This is relevant to the role, as you (usually) only assign penalties to those in the jam and it doesn't take up any space.

> Or having the box buttons on the SBO screen?

This interacts with Lead, Lost, Jammer, Pivot, and Star Pass - so directly provides functionality.

> Even at 400px screen width the cells are still wider than they are high, so adding extra width is unlikely to make them notably easier to hit.

That's not what I'm seeing - checking with the Chrome mobile thingy, the closest preset is 414pixels wide (an iPhone 8 Plus) and the cells are higher than they are tall due to this column.

As these are the primary buttons on this interface, the bigger we can have them the better.

> I'm not claiming that there is an important use case for this, but there is one. 

I'd argue that it actively is going to cause confusion, as it's not obvious what this means, its link to the role is fairly tenuous, and only of potential use to quite experienced officials who work in a certain way and I would expect have some impressive mental skills.

Either you don't have ePT and this column is useless, or you have ePT and the blue not-yet-served highlights provide largely the same benefits in a way that will work for everyone without having to mentally track how many box trips each individual skater has had.

The instant I saw this I personally was confused, and after figuring out what it was judged that it was a blatant accidental regression. The code about which columns are displayed in which mode is a bit fragile after all. Accordingly I fixed the regression before recording the training videos.

frank-weinberg:
> and after figuring out what it was judged that it was a blatant accidental regression. 

And of course, as you are the ultimate authority in those questions, no statement by the person who actually created the screen is going to change that judgement.

brian-brazil:
I had videos to record on a deadline, I couldn't hold up everything to wait on the PRs for every bug I found along the way to get merged. I still believe this is a bug, this information doesn't belong on this screen. If there actually exists some tiny subset of eLT users who could actually use this throughout a game but don't want to open the ePLT screen where this is actually tracked they can always have the IWB open in another tab or on another device (an IWB in the penalty box may be nearby too).

frank-weinberg:
How does your self-imposed deadline justify the arrogance of continuing to call something a bug after the person who designed it has repeatedly explained that it is working exactly as designed?

brian-brazil:
Having a differing opinion is not arrogance. That the code does some particular thing doesn't meant that that may then never be questioned, nor that people who challenge it are automatically arrogant. I'd request that you use less divisive language, as this doesn't help bring things to resolution.

Had I known this was going to be contentious I'd have had the videos containing this UI element and pursued the bugfix post-release, as my intention was and is that the videos match the actual release.

JeneralPain:
Please stop arguing, it's completely counterproductive.

Lets come to a solution.

frank-weinberg:
> Having a differing opinion is not arrogance.

No. But insisting that you know better than me how a screen that I designed is supposed to work is. And by using the word "bug" you're doing exactly that. If you want this to be less confrontative then how about you stop calling the PR a bugfix and instead treat it as the request for a design change it actually is.

> my intention was and is that the videos match the actual release.

Well, I'm not going to give up a feature just to save you from redoing a video.

brian-brazil:
> insisting that you know better than me how a screen that I designed is supposed to work is.

You're free to explain your thinking on how it is supposed to work, I'm free to argue that a) an explanation of a particular workflow with sufficient detail to verify plausibility has not been provided, b) I can come up with no plausible general workflow which uses this, and c) even if there could theoretically exist an official whose brain works in a way such that this is actually of benefit to them, it's still not worth it due to the additional cognitive overhead and physical effort for everyone else. The best I can come up with requires a person with working memory around 4 standard deviations above an average human. Presuming a normal distribution that's 1 in ~30k people, likely much much fewer given the time constraints also involved. (This is just a very rough estimate, the short version is you can't expect someone to remember more than about 7 digits at once. And in a game there's 15 skaters to track box trips/penalty counts for. Plus the other work an LT has to do that also require working memory, such as 3-5 slots for the lineups themselves).

I would like you to explain your thinking on how it is supposed to work, in sufficient detail that it's possible to evaluate if it's possible to do follow this workflow within the abilities of a typical human with the data volumes and time constraints of a typical game. Put another way, what procedure should I follow to determine "if the numbers are slightly off"?

If this is a good idea, then we need to be able to teach people that workflow. If there is no such general workflow, then this is what custom screens are for: things that some particular people/leagues find useful, but don't make sense for users or the scoreboard in general.

Given that I've yet to make my own paperwork PBT cross-check workflow fast enough to work and  that involves less computation on simpler data structures - plus I can employ additional data structures when working on paper - I don't think this is workable for eLT.
My expectation is that to be workable any such procedure would require adding features to eLT that amount to ePLT - and we already have a screen for that.

> And by using the word "bug" you're doing exactly that.

I can call it a wart (or possibly misfeature) instead. My argument remains the same.

> Well, I'm not going to give up a feature just to save you from redoing a video.

I'm not arguing that, I was explaining why I recorded with it removed.

frank-weinberg:
> I can come up with no plausible general workflow which uses this

You're the one requesting a change here, so you should be the one arguing how your change is improving things.

> Put another way, what procedure should I follow to determine "if the numbers are slightly off"?

Scenario 1, you have a paper-LT next to you, who does keep track of the tallies (as the majority of LTs I've worked with do): Compare the number on your screen with the tallies of the paper-LT. That's as simple as it gets.

Scenario 2, you're alone: Count the number of $ and + signs in your paperwork. Adjust for any double penalties or self-reports/revoked penalties you already know of (two numbers, which I've never seen exceed 5 in a game - and it's sufficient to remember their difference which makes it a single number in the range -5..5). Compare to the total number of penalties for your team. Even in a penalty heavy game this can be done in well under 30s.

> I can call it a wart

... because using more insulting language is surely going to help diffuse the situation.

JeneralPain:
Arguing isn’t helping us!

Seriously!!!

Can we settle this like adults please?

brian-brazil:
> You're the one requesting a change here, so you should be the one arguing how your change is improving things.

As I've said it's removing information not relevant to the role, so the argument is that it's bloating the UI for no good reason. The question then is why do we have this UI element given it's not strictly necessary, is there a reason to justify keeping it?

> Scenario 1, you have a paper-LT next to you, who does keep track of the tallies (as the majority of LTs I've worked with do): Compare the number on your screen with the tallies of the paper-LT. That's as simple as it gets.

So the LT wishes to look at the IWB, and is using the eLT for that purpose. That's not a use case that helps the eLT, you'd be better providing an eIWB to the LT.

> Count the number of $ and + signs in your paperwork. 

Towards the end of the game you've to scan across ~80 lines with ~400 fieldings, and of those ~1200 boxes of which say 50 are non-blank (40 trips, plus some occurring across jams) that you need to look closely to see whether they're -/+,  or S/$. I can count to 50 inside my head in about 10s, without having to do all the parsing.

Grabbing the first statsbook to hand (37 penalties for the team) I've found that a zig-zag pattern is best, and I'm coming in at around 40s to go through it if I want to get the right answer. Scrolling with my mouse wheel is making it a bit tricky in a way that wouldn't happen on a touch screen, so let's call it 30s.

You'd also need to spot any trips where they were released between jams which would require full parsing of each trip to spot. So you can't just look for $ and +, you need to spot unterminated S and - too which would be slower again.

> Even in a penalty heavy game this can be done in well under 30s.

You don't have that much time even. During the lineup you're getting fieldings (I'd say that it usually takes me 20s on a good day), and during the jams you should be looking at the game. So you've maybe 10-20s, and the longer it is the more likey you are to be interrupted and have to start again.

The scrolling involved is also going to make you much less responsive to performing the actual role, as every time you need to count again from P1J1.

Aside from this not working time wise, this workflow also only requires the total penalties. It does not need the entire column with the per-skater values. If you just want that total then that could go in the table header somewhere (break out the skater number column's header I guess) and skip the whole manual counting thing and present the delta between penalties and box trips to the eLT directly. If it's `box trips - penalties`, then we'd avoid a potentially confusing negative value when there's eLT but no ePT as it's just the number of box trips then.
This seems to me like a useful thing to have for both eLT and ePLT in all scenarios. Would that work for you?

> two numbers, which I've never seen exceed 5 in a game - and it's sufficient to remember their difference which makes it a single number in the range -5..5).

Agreed, it's going to be a small and easy to remember number. Plus the comment feature helps track that, and the little red dots are easy to scan for. We should probably also have a way to more easily track double penalties for a trip as a comment, but that's a topic for another day.

> ... because using more insulting language is surely going to help diffuse the situation.

Wart is the standard technical term for this: http://www.catb.org/jargon/html/W/wart.html

frank-weinberg:
> So the LT wishes to look at the IWB, and is using the eLT for that purpose. That's not a use case that helps the eLT, you'd be better providing an eIWB to the LT.

But that would require an extra device or for the eLT to switch to a different screen. With the current setup the LT can just take a glance at the screen.

> present the delta between penalties and box trips to the eLT directly. [...] Would that work for you?

For me personally, whom I have read this discussion, in the given scenario, yes. For the average NSO in general almost surely not. That's a number that's not used anywhere yet and its meaning is not immediately obvious without explanation. Penalty counts are something people have seen and know what they mean.
Plus there are other scenarios where it can be helpful to know penalty counts. E.g. an LT might be positioned well to make a team aware that they need to send a substitute. Having a penalty count helps them know when to check for a corresponding signal from the PBM. Minor and not their core duty but occasionally useful. I'm sure there are more such scenarios.


The most I'd agree to here is an option to hide the column in the event that a user is really scrambling for screen space (or explicitly wants the reduced interface for other reasons). But I won't agree with taking potentially useful info away from people as long as it doesn't interefere with something more important.

> Wart is the standard technical term for this

The given definition doesn't really make it better. For starters using the term still transmits the notion that you have chosen your stance and are unwilling to reconsider. And I don't see how a a term that's defined using words like "ugly" does not qualify as insulting on at least some level.

brian-brazil:
> But that would require an extra device or for the eLT to switch to a different screen. With the current setup the LT can just take a glance at the screen.

The question then is why does it go on this screen, rather than any other? This is very, very niche that you have both LT and eLT, and the LT is tracking penalties, but not acting as PLT, and you have an ePT, but you don't have any more devices available such as mobile phones (which are okay for eIWB in a pinch) nor can the LT look at an eIWB in the penalty box or team bench which they are likely to be beside at least one of.

> That's a number that's not used anywhere yet and its meaning is not immediately obvious without explanation.

Agreed, I was trying to find some solution though and this seemed to meet your stated needs much better than what you're proposing. By rejecting this though it's difficult not to conclude that making it possible cross check between penalties and box trips as eLT is not actually your reason for having such a strong belief that this UI element is so important.

> Plus there are other scenarios where it can be helpful to know penalty counts. E.g. an LT might be positioned well to make a team aware that they need to send a substitute. 

Once again this is a very niche scenario. You're talking a situation where you've an ePT, two eLTs but probably not an eIWB for the teams, and teams aren't tracking skaters close to foul out, and the box isn't on top things. Worst case it's an OTO to handle the foul out, which was probably going to happen anyway as it usually does.

> Having a penalty count helps them know when to check for a corresponding signal from the PBM. 

Confirming penalty counts is PT/PBM communication, not LT/PBM communication, as LTs don't officially track penalty counts and the eLT screen has no facility track per-skater penalty counts. In addition it requires that the eLTs aren't keeping an eye on the box, and if that's the case it seems pretty unlikely they're experienced enough to notice and then be able to take action to make this situation go a tiny bit more smoothly.

> I'm sure there are more such scenarios. 
> But I won't agree with taking potentially useful info away from people as long as it doesn't interefere with something more important.

As a reductio ad absurdum, why not show the full statsbook on all screens? Maybe the eLT, ePLT, or ePT is in a position to notice that the jammer was out of the engagement zone at the end of a jam, but the trailing 0 wasn't entered in the eSK. Maybe we should put the skater UUIDs back not only on the roster screen, but beside every other place there's a skater number in case they happen to help someone spot an botched skater number correction. The argument that some piece of information which is vaguely related to a position because there exist rare circumstances where it might be useful must be placed on a screen by default leads inevitably to that conclusion.

There has to be a bar, and when the only justifications involve chains of niche and/or unlikely circumstances that falls well below the bar in my view. UI elements should be tied to established roles and typical scenarios, and the presence of something niche should be due to the relevant rule/statsbook being themselves niche (e.g. overtime jams) rather than prioritising theoretical situations where they might potentially help without any consideration of how clutter could bring far more harm overall.

> For starters using the term still transmits the notion that you have chosen your stance and are unwilling to reconsider. And I don't see how a a term that's defined using words like "ugly" does not qualify as insulting on at least some level.

Any word in this space is going to have some notion that there is badness and we have to be able to express opinions that aspects of software are bad, otherwise we can never improve it.

frank-weinberg:
> LTs don't officially track penalty counts

They very often do in practice. At all levels. Just because the Statsbook manual doesn't mention it or you happend to not notice it doesn't mean it's not done. Please stop assuming that anything you havent seen in officiating must be a rare occurence.

>>as long as it doesn't interefere with something more important.

> As a reductio ad absurdum, why not show the full statsbook on all screens?

I wonder where that answer may be hidden. Maybe in plain sight?

brian-brazil:
>They very often do in practice. 

I'm not claiming otherwise, however the question here is what should an eLT screen whose purpose is to be used by a NSO doing eLT contain?

If your argument is that there should be a screen that does both eLT and allows the eLT person to track penalty counts then let's talk about if it makes sense to have such a screen. I'd argue that we should direct such users to ePLT, as it already does that. After all it's what I've seen many LTs do in terms of unofficial additional paperwork, and PLT is the future.

Wedging in one part of such a screen in to eLT is not a solution to anything though. This column does not help such an eLT track penalty counts, as once again the eLT has no way to enter or alter them, nor is it plausible for them to manually cross-check, and you rejected my suggestion on the one way I could think of to have a viable feature here. Thus it is useless for eLT, whether they're trying to track penalty counts or not.

This information is at best a distraction, taking up space that could be better used for the actual eLT, and in the worst case will cause confusion and problems during a game. It should never have been added to this screen, and should be removed now.

> Just because the Statsbook manual doesn't mention it or you happend to not notice it doesn't mean it's not done.

Extra paperwork/notes is possible on paper because there's blank space, you can write anything, and indeed it's encouraged. That doesn't work with electronic roles, where we have to explicitly design all the input and interactions upfront. While I wouldn't mind a few more notes fields, it's never going to be as expressive or have as many options as what you can do on paper.

In practice this means the screens need to focus on the core of the roles. Where there's existing "unofficial" practices which help the electronic role (e.g. the SK dot system) we have incorporated them, but most of them aren't going to work out due to the change in medium and single source of truth nature of the scoreboard (e.g. ePT isn't going to be tracking jammers/TTOs/ORs, whereas a PT usually would).

>> As a reductio ad absurdum, why not show the full statsbook on all screens?
> I wonder where that answer may be hidden. Maybe in plain sight?

That would be my point, it's an absurd idea. Adding arbitrary elements to screens what have no plausible workflow is a bad idea. There's basically as much justification for adding per-player scores to eLT as there is per-player penalty counts, which is to say essentially none.

The eLT's purpose is to serve the eLT. If someone wants to experiment with a non-standard disposition of responsibilities among the NSOs, they're free to build custom screens.

frank-weinberg:
>If your argument is that there should be a screen that does both eLT and allows the eLT person to track penalty counts

My argument is that eRole should have all info available that Role would typically have available. Most LT track penalty counts, so that info is typically available.

brian-brazil:
> My argument is that eRole should have all info available that Role would typically have available.

I don't disagree in general...

> Most LT track penalty counts, so that info is typically available.

...however I think you're misinterpreting the situation here. We have cross-role information elsewhere (e.g. ePT gets the current lineups as a hint, which has no basis in the paperwork at all - but is clearly useful), but this doesn't fit that mould.


The observation is that many experienced LTs will produce extra paperwork, from which per-player penalty counts are readily available. We both agree this is a thing.

Your argument is then that based on this we should make per-player penalty counts available on the eLT screen, however I have demonstrated above (and you've not provided any counter-arguments/evidence) that such information has no plausible use.

So having this information on the screen, and indeed the paperwork, smells like cargo culting to me. While that appears be the case electronically, it'd seem odd (though not entirely unexpected give the spread of officiating practices in the past) that so many experienced LTs would end up doing pointless work that doesn't benefit either them or the game.

So let's say the goal of this extra paperwork was indeed to have per-player penalty counts. We know that's not really useful for the LT paperwork per above, however knowing those numbers would allow a quick cross-check with the PT to see if a penalty was missed by the PT and for which skater. From there you can take an OTO if it risks being game impacting, and trawl through the paperwork to get it all resolved.
For eLT/ePT that's pointless though, as the penalty counts on the eLT screen came from the ePT so you'd just be giving the ePT back their own numbers.

So that's one potential viable use on paper, but which doesn't apply electronically. What if having per-player penalty counts readily available isn't actually the goal of the extra paperwork, but a mere side effect of something else? This is what I believe is primarily happening here.

My belief is that it is not the per-player counts themselves that are important, but the process of recording the per-player penalties that is the useful thing to some experienced LTs. By writing down the penalties as they hear them, they a) force themselves to be more aware of penalty calls and thus b) may notice if they've forgotten to write down a box trip on the LT paperwork, c) may have additional personal annotations/mnemonics around when a penalty/box trip is "done", d) other stuff that varies per person.

That this produces a per-player record and thus count of penalties is incidental. Just like a PT writing down the jammer numbers helps them track the jam number when there's no scoreboard - and incidentally produces a backup list of all the jammers in case the SK/LT missed them.


So in effect if we wanted the benefits of this extra paperwork, it would seem that you basically have to either switch to doing ePLT, or give the eLTs some paper.

frank-weinberg:
>I have demonstrated above (and you've not provided any counter-arguments/evidence) that such information has no plausible use.

I stopped reading at this point. I did provide examples for plausible use. If you decide to simply ignore them, there's no basis for further discussion. The line I've drawn above stands: I will not accept anything beyond an option to hide the column.

brian-brazil:
> I did provide examples for plausible use. If you decide to simply ignore them, there's no basis for further discussion. 

I didn't ignore them, I refuted them. They all either require some quite strained assumptions and/or don't benefit the eLT. The only one that might benefit the eLT I provided evidence that it was unworkable in practice, which you have not challenged.

I see no use cases that come close (even in aggregate) to justify taking up ~1/7th of the horizontal space on this screen. Would you like to provide any further arguments?

I'm happy to engage in actual debate so we can find the right way forward here. If you've got some workflow in mind that you've failed to communicate thus far but would make eLT better for everyone then let's talk about it. Taking the stance that you refuse to even discuss this is not however productive.

frank-weinberg:
> I provided evidence that it was unworkable

You tried it once, it took longer than you deemed worthwhile and from that you concluded that no one will ever want to do that. If that counts as evidence, then my sample of multiple NSOs who have said they don't want this column to be removed (vs. no one except from you preferring removal) is sufficient evidence to keep it.

brian-brazil:
> You tried it once, it took longer than you deemed worthwhile and 

When I do a benchmark, I do try to do it fairly.

As you didn't share a specific technique I tried it multiple times, trying to find the most efficient way to do it accurately, made allowance for my hardware/software being suboptimal, and doing it repeatedly on the same dataset would have naturally made me a little faster each time too. I probably did around 20 attempts in all.

Given the above, I would consider 30s as a good lower bound. If memory serves my actual best was 37s. 

> from that you concluded that no one will ever want to do that

I didn't come to that conclusion, as I don't control what people will want.
I concluded however that it takes too long to actually be of use during a game, which'd need to be nearer 10s.

In any case this particular argument doesn't support per-skater penalty counts, only the overall team penalty count.

> then my sample of multiple NSOs who have said they don't want this column to be removed 

Have any of them articulated a good reason to add this column? My own experience of getting feedback on this is that it's one of those ideas that sounds good on the surface, but that evaporates when you ask them why they want it. And that's before any potential costs/downsides are considered.

Per my response two back, I think it's largely cargo culting here. Probably with a bit of an undercurrent that more data is automatically better (an anti-pattern I often see in my day job, where the moral equivalent of putting the entire statsbook on every screen is not uncommon).
What makes sense on paper doesn't automatically make sense electronically.

frank-weinberg:
>Have any of them articulated a good reason to add this column?

Why would they? It's already there.

And let's go through your arguments for removal:

- On some screens space it at such a premium that the space this column is taking would be required for the other cloumns to be usable. - That's what an option to hide it is for.

- It might confuse people. - That's an unsubstantiated claim thus far. I have yet to see anybody being confused. And if it actually did, that would be a reason to change the default of the option but not to remove it altogether as long as other people find it useful.

- The people that find it useful are mistaken about that and just following a cargo cult. - That's you patronizing people about how they should do their job. We're providing a tool here. A good tool is adjusted or adjustable to the user's needs instead of forcing the user to adjust to the tool.

Did I miss anything?

brian-brazil:
> Why would they? It's already there.

I thought they might have requested it before it got added. That would mean then that this UI element is entirely your idea, and noone else had asked for to it to be added.

Did any of them give a to reason why they don't want it to be removed?
Or put more bluntly, has anyone presented a use case or workflow for this that didn't originate from you?

> On some screens space it at such a premium that the space this column is taking would be required for the other cloumns to be usable. - That's what an option to hide it is for.

It's not just about space being a premium, this is a continuum. Making everything else smaller makes things harder to read, and buttons smaller targets to hit. Both will slow people down, and/or lead to mistakes.

Every UI element has a cost. The question is is this one worth it?

In addition every option has a cost, both to users and for maintenance. It's usually better to hardcode than add an option.

> It might confuse people. - That's an unsubstantiated claim thus far. I have yet to see anybody being confused. And if it actually did, that would be a reason to change the default of the option but not to remove it altogether as long as other people find it useful.

Cognitive overload is a thing, and tends to be subtle. Every UI element confuses people, even if they don't realise it. The question is which ones are then worth it. Usually it doesn't come up as UI elements typically have a clear purpose with clear benefits that more than make up for this.

What's even more fun is that people can think something is useful, even though it demonstrably is far from optimal. The more complicated or convoluted something is the more likely they'll vehemently argue that it's essential.

> That's you patronizing people about how they should do their job.

No, it's me trying to explain the observation that this appears to have no notable real world uses but there's extremely strong sentiment for keeping it anyway. I'm trying to argue in good faith and find the most charitable interpretation of things, so I'm open to other explanations.


This is why I'm trying to tease out a workflow, as that's evidence which would show that this is actually a good idea and not merely blindly copying things from paperwork.

> A good tool is adjusted or adjustable to the user's needs instead of forcing the user to adjust to the tool.

A good tool also serves a user's actual needs, without extraneous aspects. What a user says they need versus what they actually need regularly differ, e.g. the XY problem.

A good tool does not cater to every whim of users, that's how we ended up with e.g. 30+ overlay screens.

Let's serve users actual needs.

> Did I miss anything?

I'm arguing not only that this should be removed, but that it should never have been added in the first place.

Your argument at this point appears to be primarily that it's here now, so we're not ever allowed remove it even if it turns out that there wasn't a good reason to add it in the first place.

I don't buy that argument, and am asking you to justify adding it in the first place. This slipped past code review (as I said it sounds good at first glance, so my mind skipped past that sentence) so we never had that discussion at the time. I would hope that we wouldn't be in a place where something missed in code review can never be subsequently challenged.

frank-weinberg:
>It's usually better to hardcode than add an option.

The more diverse a user base is, the less likely a one-size-fits-all approach is to be a good one. And our user base is very diverse in many aspects.

>Let's serve users actual needs.

You're not doing that by claiming you know their needs better than they do. You can try to explain why you think they actually need something else but it will still be them who know best what they need. If your arguments don't convince them, your assumption should be that you haven't understood them fully, not that they don't know what they need. Because that's - I repeat myself - patronizing.

brian-brazil:
> The more diverse a user base is, the less likely a one-size-fits-all approach is to be a good one.

This is true in general (e.g. the colourblind stuff we just added), but I'm not sure how it helps your argument here. Accessibility concerns would argue towards removing this, as arguments in this area are stronger forms of the arguments I've been making.

> If your arguments don't convince them, your assumption should be that you haven't understood them fully, not that they don't know what they need. 

I have attempted to understand you and have given you multiple chances to so do. I have even refuted stronger versions of arguments you've made, and come up with arguments on your side.
At this point any responsibility for failing to be understood is on you.

> it will still be them who know best what they need.

I've already explained that this is not how humans work. Cialdini's book is one of the standard texts if you want to read more, but the short version is that humans say they need isn't automatically what they actually need.

As another reductio ad absurdum, if someone says they don't need vaccines then they must be right even if all the evidence indicates otherwise.

>> Or put more bluntly, has anyone presented a use case or workflow for this that didn't originate from you?

I note a lack of an answer to this question, from which I infer that this UI element is entirely for you. 

I don't think it's wise to add UI elements that are only for the sake of a single user - even if that single user finds it to be the most useful thing ever -  that's what custom screens are for.

frank-weinberg:
>from which I infer

Infer whatever you want. I'm blocking this PR based on "It brakes a Use case". Unless you come up with an advantage this PR brings beyond "fewer UI elements are better per se" or change it to comply with the line I've stated, I won't reply further.

AlpenYeti:
Well, this certainly looks like an unhealthy way of ending things... I am not an experienced eLT so I will not be able to give a very well qualified point of view on this side, but as a curious and somewhat logical software user, I would like to add as an argument towards removing it that it feels weird to have an information on your screen that you can not interact with.

Counting trips as eLT would seem interesting because when you press the Box trip button, you see the counter go up. In the case we are currently calmly discussing, it means you have a piece of information that updates without you being able to fix it, act on it or anything.

For me, having an information on screen that I can not change myself means it should not be there. I don't want to go to Brian's lengths as far as _reductio ad absurdum_ goes, but I would say that it is as useful to keep as would the period clock on all the screens. It is always a nice info to have, it can help in some cases, but most of the time, you do not need it or if you want it you can have it on the screen in question.

I hope this might help you all go forward a bit further...

PR #392, opened 2019-11-21
Put twitter auth callback page back where it was.

brian-brazil:
Changing the URL breaks logins.


JeneralPain:
I'm happy to merge this, because we have to change the twitter dev api to show the right url's or it will break.

This also means, moving the twitter url around, all existing users will work, new users with new code will not.

PR #393, opened 2019-11-21
Update game data files after score is final

### LINKED ISSUE #377: Changes after declaring the score final don't get added to the game data file
Such changes are a regular occurence when PT/LT/PBT compare their notes, which usually isn't completed before "Final Score".
This can be worked around by unsetting and resetting the "Final Score" flag after all changes have been made, but there should be a more obvious way.

frank-weinberg:
closes #377 

This will write on any change to relevant data once the score is final but allows the user to
lock the data in once post-game updates are complete to protect against accidental changes after that point.

Also place restoring from autosave in a batch.

### DIFF
--- a/html/components/team-time-tab.js
+++ b/html/components/team-time-tab.js
@@ -145,8 +145,6 @@ function createMetaControlTable() {
 
 	var confirmedButton = toggleButton("ScoreBoard.OfficialScore", "Official Score", "Unofficial Score");
 	confirmedButton.appendTo(periodEndTd);
-	var statsLockedButton = toggleButton("ScoreBoard.StatsLocked", "Stats Locked", "Writing Stats");
-	statsLockedButton.appendTo(periodEndTd);
 	var periodEndTimeoutDialog = createPeriodEndTimeoutDialog(periodEndTd);
 	$("<button>").addClass("PeriodEndTimeout").text("Timeout before Period End")
 		.appendTo(periodEndTd)
--- a/html/components/team-time-tab.js
+++ b/html/components/team-time-tab.js
@@ -145,6 +145,8 @@ function createMetaControlTable() {
 
 	var confirmedButton = toggleButton("ScoreBoard.OfficialScore", "Official Score", "Unofficial Score");
 	confirmedButton.appendTo(periodEndTd);
+	var statsLockedButton = toggleButton("ScoreBoard.StatsLocked", "Stats Locked", "Writing Stats");
+	statsLockedButton.appendTo(periodEndTd);
 	var periodEndTimeoutDialog = createPeriodEndTimeoutDialog(periodEndTd);
 	$("<button>").addClass("PeriodEndTimeout").text("Timeout before Period End")
 		.appendTo(periodEndTd)
--- a/src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
+++ b/src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
@@ -28,13 +28,28 @@ public class JSONStateSnapshotter implements JSONStateListener {
         if (state.get("ScoreBoard.CurrentPeriodNumber") != "0") {
             // If the jam has just ended or the score is now official, write out a file.
             if ((inJam && !bool(state.get("ScoreBoard.InJam")))
-                    || (!officialScore && bool(state.get("ScoreBoard.OfficialScore")))) {
+                    || (bool(state.get("ScoreBoard.OfficialScore"))
+                            && !bool(state.get("ScoreBoard.StatsLocked"))
+                            && containsRelevantUpdate(changed))) {
                 writeFile(state);
             }
         }
 
-        officialScore = bool(state.get("ScoreBoard.OfficialScore"));
         inJam = bool(state.get("ScoreBoard.InJam"));
+    }
+    
+    private boolean containsRelevantUpdate(Set<String> keys) {
+        for (String key : keys) {
+            if (!key.startsWith("ScoreBoard.Clock")
+                    && !key.startsWith("ScoreBoard.Twitter.")
+                    && !key.startsWith("ScoreBoard.Media." )
+                    && !key.startsWith("ScoreBoard.Settings." )
+                    && !key.startsWith("ScoreBoard.Rulesets." )
+                    && !key.startsWith("ScoreBoard.PreparedTeams(" )) {
+                return true;
+            }
+        }
+        return false;
     }
 
     public void writeFile(Map<String, Object> state) {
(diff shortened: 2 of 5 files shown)


html/components/team-time-tab.js
```diff
@@ -145,6 +145,8 @@ function createMetaControlTable() {
 
 	var confirmedButton = toggleButton("ScoreBoard.OfficialScore", "Official Score", "Unofficial Score");
 	confirmedButton.appendTo(periodEndTd);
+	var statsLockedButton = toggleButton("ScoreBoard.StatsLocked", "Stats Locked", "Writing Stats");
```
brian-brazil: As per the issue, this should not involve operator action.

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
```diff
@@ -28,14 +28,29 @@ public synchronized void sendUpdates(Map<String, Object> state, Set<String> chan
         if (state.get("ScoreBoard.CurrentPeriodNumber") != "0") {
             // If the jam has just ended or the score is now official, write out a file.
             if ((inJam && !bool(state.get("ScoreBoard.InJam")))
-                    || (!officialScore && bool(state.get("ScoreBoard.OfficialScore")))) {
+                    || (bool(state.get("ScoreBoard.OfficialScore"))
```
brian-brazil, line 31: A simpler way to do this would be to make this just `bool(state.get("ScoreBoard.OfficialScore")` without the state. We can then filter out the Clock changes if we want to.

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
```diff
@@ -28,14 +28,29 @@ public synchronized void sendUpdates(Map<String, Object> state, Set<String> chan
...
             }
         }
 
-        officialScore = bool(state.get("ScoreBoard.OfficialScore"));
         inJam = bool(state.get("ScoreBoard.InJam"));
     }
+    
+    private boolean containsRelevantUpdate(Set<String> keys) {
+        for (String key : keys) {
+            if (!key.startsWith("ScoreBoard.Clock")
```
brian-brazil, line 42: Let's not have 2 lists of things we care about, we're already filtering below.

html/components/team-time-tab.js
frank-weinberg: Writing does not require operator action - this setting is off by default. If it's never touched and the only inputs between final score and Start New Game are corrections, then the Game data file will contain exactly the expected data.
This button merely adds the option to block (accidental) writes once the NSOs finished their cross-checks, if the next game isn't started right away..

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
frank-weinberg: What's the point of splitting the conditions on which to write the file to two places? Or do I misunderstand the suggestion?

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
frank-weinberg: The two lists are different and serve different purposes - this one is about WHEN not to write, the other is about WHAT not to write. (And while the what list will probaly always be a subset of the when list, the inverse does not hold.)

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
brian-brazil: My point is that rather than looking for the point that OfficialScore is changed and writing out only then, that we update the file continuously once the score is final.

html/components/team-time-tab.js
brian-brazil: The button doesn't lock stats though, they can still be changed - they'll just not get written out to one particular way of getting stats out. I think this would cause more confusion than anything.

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
brian-brazil: It feels unclean, and likely to cause maintenance issues. Writing the file is pretty cheap overall, particularly once the game is over.

html/components/team-time-tab.js
frank-weinberg: >one particular way of getting stats out

The one we advertise as the default way.
But I'm open for suggestions as to a better label.

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
frank-weinberg: Which is what I am doing. It used to look for a change in the old code.

html/components/team-time-tab.js
brian-brazil: I'd suggest not having it, among other things it's not in the videos I just recorded.

The original code missed score updates after the last jam ended. The current code misses paperwork updates after the score is final. What I propose won't miss anything, but if someone messes around might lose data through an overwrite. Let's keep it simple for now, and only add a button if we really need it. A game data manager can always download the data once it's final, and they're likely in a better position to judge that than the SBO (who has probably disappeared off by that stage).

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
frank-weinberg: If someone is running the scoreboard from a flash based medium, each write is eating into the lifetime of the medium. And while the chance of anything going wrong during a write is miniscule, this will add up if we do loads of unnecessary writes.

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
brian-brazil: Ah right, I'm not quite awake. I suggest not having the two clauses after this around locking and relevant updates.

src/com/carolinarollergirls/scoreboard/json/JSONStateSnapshotter.java
brian-brazil: We don't sync the file, so it'll only get flushed out to disk every minute or so. The regular autosaves every 10s will use up more write bandwidth.

html/components/team-time-tab.js
frank-weinberg: > among other things it's not in the videos I just recorded

Alright. So let's stop the bugfixing then.

brian-brazil: Yeah, best to avoid non-critical UI changes until the release.

frank-weinberg: > best to avoid non-critical UI changes

So I can expect you to withdraw #391 ?

brian-brazil: No, that's still a bugfix that should go in one way or the other.

PR #396, opened 2019-11-26
Don't show jammers when clocks aren't running.

brian-brazil:
This was causing jammers to appear when no clocks
were running, e.g. just before a period start when
the intermission clock wasn't running.

That isn't a problem in itself, however if you were coming
into P2 and had lead in the last jam of P1, then lead
would be incorrectly indicated before the jam start.

So go with consistency when we're in lineup,
and don't show the jammer name.

We could also go with altering the ShowIn logic, but that seems risky giving the release coming up as it touches everything related to clocks.


brian-brazil:
And as I prepare release notes, it looks like #292 was a previous attempt to fix this.

PR #400, opened 2019-11-27
Don't let clock sync interfere with restoring clock snapshots.

brian-brazil:
Also ensure all the clocks are gone from UpdateClockTimerTask
in tests by resetting the scoreboard in teardown.

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
@@ -157,7 +157,7 @@ public class ClockImpl extends ScoreBoardEventProviderImpl implements Clock {
         synchronized (coreLock) {
             if (s.getId() != getId()) { return; }
             setNumber(s.getNumber());
-            setTime(s.getTime());
+            set(Value.TIME, s.getTime(), Flag.INTERNAL);
             if (s.isRunning()) {
                 start();
             } else {
--- a/tests/com/carolinarollergirls/scoreboard/core/impl/ScoreboardImplTests.java
+++ b/tests/com/carolinarollergirls/scoreboard/core/impl/ScoreboardImplTests.java
@@ -88,6 +88,7 @@ public class ScoreboardImplTests {
     @After
     public void tearDown() throws Exception {
         ScoreBoardClock.getInstance().start(false);
+        sb.reset();
         // Check all started batches were ended.
         assertEquals(0, batchLevel);
     }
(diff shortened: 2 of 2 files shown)


src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
```diff
@@ -157,11 +157,10 @@ public void restoreSnapshot(ClockSnapshot s) {
...
             if (s.getId() != getId()) { return; }
             setNumber(s.getNumber());
+            stop();  // Avoid sync logic interfering.
             setTime(s.getTime());
             if (s.isRunning()) {
-                start();
-            } else {
-                stop();
+                start(false);
             }
```
frank-weinberg: Stopping and restarting a clock will fire two events. Have you verified that this does not cause unwanted effects anywhere?

brian-brazil: If it did, we should be seeing it in unittests.

frank-weinberg: We have less than 100% coverage at the unit test level. And this would only be caught at integration or system test level, where our coverage is way worse. Or it might even be in the frontend, which we don't have any automated tests for.

brian-brazil: The snapshot restore logic should undo anything that happened, and I don't see any issues from a quick peek in the code. The frontend logic shouldn't have anything where it depends on transitions rather than states, so this should have no impact (plus it'll all be in one batch).

Is there another way you'd suggest doing this?


Once this PR is in, I'm going to see if it's practical to turn on sync in unittests with the tearDown fix included here. That'll give us a lot more real coverage.

frank-weinberg: > The snapshot restore logic should undo anything that happened,

But if an event listener changes a clock that has already been restored, it won't be undone.

>  (plus it'll all be in one batch).

True. That will filter the events from the frontend.

>Is there another way you'd suggest doing this?

No. I just noticed this potential problem during review and wanted to be sure it's not an actual problem. And I didn't have time to check myself.

brian-brazil: > But if an event listener changes a clock that has already been restored, it won't be undone.

That's a possibility. Maybe I'll come back to this PR after doing the clock cleanup I've planned.

frank-weinberg: > Is there another way you'd suggest doing this?

After looking at the code more closely: just replace `setTime(s.getTime())` with `set(Value.TIME, s.getTime(), Flag.INTERNAL)` - that's how we already keep sync from interfering with timer ticks.

brian-brazil: I was just thinking that today actually, updated.

PR #401, opened 2019-11-28
Avoid reverse DNS resolution when looking for URLs.

brian-brazil:
At most events the scoreboard will not be connected to the internet,
so users end up having to wait for a DNS timeout before the URLs show
at startup or on the index page.

This loses us hostnames, but it's very unlikely that there was a DNS
setup that would have allowed them to work.

We'll still do forward resolution to bind to if hostname is provided on
the command line, but that's going to be rare and also likey something
present in /etc/hosts so it won't hit the network.

Also cleanup and modernise the code generally, and don't exclude v6
addresses.

Something I noticed over the weekend.

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
+++ b/src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
@@ -15,8 +15,7 @@ import io.prometheus.client.hotspot.DefaultExports;
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.SocketException;
-import java.net.URL;
-import java.util.Iterator;
+import java.util.Set;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.ServletException;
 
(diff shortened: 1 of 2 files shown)


src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
```diff
@@ -50,12 +50,12 @@ public JettyServletScoreBoardController(ScoreBoard sb, JSONStateManager jsm, Str
...
-            Iterator<URL> urls = urlsServlet.getUrls().iterator();
-            if (urls.hasNext()) {
+            Set<String> urls = urlsServlet.getUrls();
+            if (!urls.isEmpty()) {
                 Logger.printMessage("or try one of these URLs:");
             }
-            while (urls.hasNext()) {
-                Logger.printMessage("	"+urls.next().toString());
+            for (String u : urls) {
+                Logger.printMessage("	"+u.toString());
```
frank-weinberg: `toString()` is now redundant

src/com/carolinarollergirls/scoreboard/jetty/JettyServletScoreBoardController.java
```diff
@@ -16,7 +16,7 @@
 import java.net.MalformedURLException;
 import java.net.SocketException;
 import java.net.URL;
```
frank-weinberg: `URL` is no longer used

src/com/carolinarollergirls/scoreboard/jetty/UrlsServlet.java
```diff
@@ -16,11 +16,11 @@
 import java.net.SocketException;
 import java.net.URL;
 import java.net.UnknownHostException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
```
frank-weinberg: `Arrays`, `Iterator`, and `Inet4Address` are no longer used

brian-brazil:
All fixed.

PR #402, opened 2019-11-28
Remove minimum time, hardcode to 0.

brian-brazil:
The ruleset settings for it were removed in #176, but these
were still hanging aroung.

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
@@ -35,8 +35,7 @@ public class ClockImpl extends ScoreBoardEventProviderImpl implements Clock {
         } else {
             values.put(Value.NUMBER, 0);
         }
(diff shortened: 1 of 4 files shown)


src/com/carolinarollergirls/scoreboard/core/impl/ClockImpl.java
```diff
@@ -56,15 +55,15 @@ else if (flag != Flag.INTERNAL) {
             if ((flag == Flag.RESET && isCountDirectionDown()) || (Long)value > getMaximumTime() + 500 && (!isCountDirectionDown() || flag != Flag.RECALCULATE)) {
                 return getMaximumTime();
             }
-            if ((flag == Flag.RESET && !isCountDirectionDown()) || (Long)value < getMinimumTime() - 500) {
-                return getMinimumTime();
+            if ((flag == Flag.RESET && !isCountDirectionDown()) || (Long)value < 0 - 500) {
```
frank-weinberg, line 58: Why not `-500` directly?

brian-brazil: I did it this way so it matches the pattern of the if above it.

PR #407, opened 2019-12-04
Make it possible to see devices connecting to the scoreboard, and on what pages

brian-brazil:
As the scoreboard holds ever more game data and is more interesting to more people who will get access, there's increased chances of a mishap. This PR makes it possible to see current and past connections, what pages they're on, give indications of hardware, and if they've performed a write.

A future PR will add simple authorisation, so some things here make more sense in that contex e.g. you need the names in order to grant permission to the right device.
As this is a security feature, the unittests go much deeper than usual to ensure that someone who is permitted to write can't cover their tracks as even with auth being available it's expected that most won't enable it. Another future PR will add audit logging, which is the proper way to do this (and will help with debugging) but doesn't require deep skills to figure out if something might have happened.

A number of other improvements were made in passing, including:
* A save is now made on shutdown
* URLs are only shown in logs once the http server is running
* GUI text is monospace
* Logging in as SBO no longer causes writes
* Index page now uses WS for the version, letting us remove a custom servlet
* Throwing an exception if a set doesn't work due to programmer error, rather than silently failing


frank-weinberg:
I think that this is too big to be put in 4.0.1 as per our discussion. Thus I'll postpone reviewing it until after I've finished my list of bugfixes that should be in 4.0.1.

brian-brazil:
4.0.1 is out now, so this can be reviewed.

frank-weinberg:
Both the extra fields on the index page and the devices page are of interest to only a small subset of users and will look like gibberish to the majority of users. By the standard you're trying to establish in #391 these should be custom screens.

Other UI issues:

- The names are confusing in that they look like they have a derby meaning but don't. As they are centrally issued, why not just go "Device 1", "Device 2", ... and give the user the option to change the name?
- What do the comments do? They are not visible anywhere but on the page you enter them. And the shared column between comments and pages is not intuitive. If their purpose is just to ID devices, a changable name will achieve this in a more intuitive fashion. Also, toggling comments on the devices screen has no visible effect.
- Platform and address will only be of some use when you are trying to identify a rogue device that got into the network. And even then their use will be limited. For any known devices the name displayed on the device will be the way to identify them. -> These columns should probably be hidden by default. (Or, as you would presumably prefer, be removed altogether.)
- Logging in as a known operator still caused a write.

There's also some code changes that looked questionable on first read. But I'll have a closer look at those later on.

brian-brazil:
> Both the extra fields on the index page and the devices page are of interest to only a small subset of users and will look like gibberish to the majority of users.

The main page is not used during games, and that bit of the page already has dynamic content. I had tried having it down the bottom with the version, but didn't look as good.

If you want to argue this, then the debate should be about whether the scoreboard should have security-related features at all. Given our userbase and typical hardware I don't think things like reverse proxies, CA authorities, TLS client certs, 802.11x, and firewall rules are realistic options. I think this is as simple as I can get it while providing reasonable security when considered against physical threats such as stealing an authorised device or pulling the fire alarm.

> By the standard you're trying to establish in #391 these should be custom screens.

This is new feature for the scoreboard, so it cannot be done via custom screens as it requires backend changes. If someone wanted to view these in some other way, that'd be a case for a custom screen however this is the "default" view on to this information.The next stage is auth, which can also only be done in the backend and will need a screen (this one) to control.

> As they are centrally issued, why not just go "Device 1", "Device 2", ... and give the user the option to change the name?

The idea is that the name is an id which is both distinct and memorable, and I'd rather not expose UUIDs to users. Other software like Docker uses this for similar use cases, and it seems to work out pretty well.

> What do the comments do?

Comments are intended as user provided names.

> Also, toggling comments on the devices screen has no visible effect.

It should be including devices with comments, if they aren't already included via another toggle. Seems to work from a quick check.

> Platform and address will only be of some use when you are trying to identify a rogue device that got into the network.

That's partly why they're there, plus you can't assume that someone has done the pre-setup and so is trying to guess which is which if e.g. you can see that ePT is using an iPad and the box has a Laptop. Also when auth is added it'll be extra indication that you're selecting the right device.

Once we've logs we might be able to cut down what's here, based on what is/isn't useful in practice. I aimed for things a power user could find useful to track down a particular piece of hardware from the rather limited set of information available from Java/Javascript. For example I would have liked hardware manufacturer via the MAC address, but that's not sanely accessible cross-platform so we get to trust the (easier to change) user agent instead.

> Or, as you would presumably prefer, be removed altogether.

This screen isn't used to collect stats during a game, screens for use by a Game Data Manager have a different set of tradeoffs. Unknown devices are interesting precisely because they are unknown.

> Logging in as a known operator still caused a write.

Hmm, I can't reproduce. What was written to?

frank-weinberg:
> then the debate should be about 

what our criteria are for including features. You're constantly requesting features to be removed from PRs (or even from the established codebase) and discarding any use cases for them because you consider them to be too niche. Yet here you present something that a very small subset of the users will even understand let alone use and suddenly the fact that it is niche is completely irrelevant. That's not acceptable. If you're not willing to apply this standard to your own contributions, you should drop it altogether.

> this is as simple as I can get it while providing reasonable security

This is providing zero security. Logging is not security.

> The idea is that the name is an id which is both distinct and memorable, and I'd rather not expose UUIDs to users

An increasing number is guaranteed to be distinct (whereas your pseudo-UIDs are just likely to be) and a 1 to 3-digit number is not harder to memorize than two random words. They are also not exposing UUIDs.

> Comments are intended as user provided names.

So why do you not use them as names but instead hide them away from the user?

> screens for use by a Game Data Manager have a different set of tradeoffs

Because those are a different breed of human that deals with information differently? You were presenting "each UI element is a step towards TMI and has to be thoroughly justified" as a general principle. And now that it's in your way it's only applicable selectively?

> What was written to?

It didn't say.

brian-brazil:
> Yet here you present something that a very small subset of the users will even understand let alone use and suddenly the fact that it is niche is completely irrelevant. 

I never claimed it wasn't niche, and I am happy to discuss whether we should have auth features at all. If we have authorisation we need to first have some way to identify&authenticate clients, so that we can then authorise them. This screen represents my thoughts on how to do that for this particular use case.

Unlike the eLT case however it is not related to time-sensitive data entry while under pressure, plus  it'll likely be on a laptop rather than a tablet. The UX concerns and other tradeoffs are thus different and much more relaxed than the NSO-facing screens. This is more like the Rulesets page UX wise.

The other thing with this feature is that it's completely optional, the various electronic NSO positions don't need to know how this works or that auth features even exist. Realistically only the handful of "professional" game data managers might be forced to care, if they weren't interested already. Similarly with logging, the logs will be there if you need them but otherwise they're as relevant to a user as the fact that we keep 6 autosaves around.

> This is providing zero security. Logging is not security.

As indicated in the PR description, the next one will add auth. This PR is big enough as-is.

It's also untrue that this is not security, being tamper-evident is a valid security technique. Here it may well be enough to discourage most bad actors. This is also not event logging, that'll be another PR yet again.

> So why do you not use them as names but instead hide them away from the user?

I am using them effectively as names, but they're only relevant on the Devices screen so that's the only place I made them visible.

> An increasing number is guaranteed to be distinct (whereas your pseudo-UIDs are just likely to be) and a 1 to 3-digit number is not harder to memorize than two random words. 

More likely to be miscommunicated though, and a long running scoreboard install will get into bigger numbers (hitting 10k isn't out of the question over a single season). I can make it fully unique if it's really a concern. It's also a bit of fun.

> It didn't say.

If you catch it again let me know, as there's no logging yet you'll need to use your browser's debugging to poke at the WS connection.

frank-weinberg:
>  This is more like the Rulesets page UX wise.

You have opposed the addition of rules and settings using the same argument. You also opposed the addition of an extra file category in the Up/Download screen. This isn't about this PR compared to #391. This is about you setting up rules and restrictions on my PRs and features and then breaking all of them in your own PR. When thos goes in, I won't accept any of these arguments as valid to block a PR or remove a feature going forward. Also including:

> the next one will add

You repeatedly disallowed planned future features as a reason to add code/UI elements now.

> names, but they're only relevant on the Devices screen 

There is also a name displayed on the start screen. Using the user provided name here (if there is one) would remove any worries about them possibly being hard to remember. I don't see any benefit at all to having the two of them in parallel.

> Here it may well be enough to discourage most bad actors.

Because most bad actors will be in the small subset of people that are even aware of this feature?

> hitting 10k isn't out of the question over a single season

How? That would require 200 devices per week. And different devices at that so they aren't remembered. That's completely unrealistic.

brian-brazil:
In general I want to try to keep every screen focused on its task/role, and to fight feature creep.

This PR is feature creep, so once again the thing to debate is do we want auth features at all.

> You have opposed the addition of rules and settings using the same argument.

No, I argued that they shouldn't be a part of the operator screen. I didn't say we should get rid of rules and settings entirely.

> You also opposed the addition of an extra file category in the Up/Download screen. 

That was feature creep, and there was more to it than that.

> This is about you setting up rules and restrictions on my PRs and features and then breaking all of them in your own PR. 

I believe I have been consistent, it only appears inconsistent if all the nuances and details are ignored.

My argument in both #407 and for rules/settings on the operator screen relates to not having unnecessary UI elements on NSO facing screens. If I was proposing to add this information to an existing screen then you'd have a point.

> When thos goes in, I won't accept any of these arguments as valid to block a PR or remove a feature going forward.

Either of us doing whatever we want to the codebase is not a viable way to operate, no matter what the claimed justification is. Good faith debate and working to find consensus is key.

> You repeatedly disallowed planned future features as a reason to add code/UI elements now.

This PR makes sense on its own, even if there weren't going to be further PRs building on it in the immediate future. I believe that the UI/UX should always be sane, there's more lenience on the backend.

This PR is already over 1k lines with 32 commits. The etiquette I'm used to is that many moderately sized PRs are preferable to monster PRs. If you're okay with reviewing even larger PRs then I can do larger PRs.

> There is also a name displayed on the start screen. Using the user provided name here (if there is one) would remove any worries about them possibly being hard to remember. I don't see any benefit at all to having the two of them in parallel.

That requires that the data manager set it all up in advance though, which they should only need to do for the largest of events. There many also be duplicate comments, who cares which ePLT tablet is which? Put another way, even with user provided names you still need to show an ID.

> How? That would require 200 devices per week. And different devices at that so they aren't remembered. That's completely unrealistic.

The same device can appear multiple times, which can happen if the bouts are at least 2 weeks apart. Or if cookies are being wiped more regularly or rejected. Events with hundreds/thousands of attendees are a thing. Auto-incrementing ids are not a panacea, and we use UUIDs internally anyway.

frank-weinberg:
> I argued that they shouldn't be a part of the operator screen.

I'm not talking about that one I'm talking about individual rules/settings that you argued to leave out as being too niche to be worth the extra row/button/dropdown.

> This PR makes sense on its own

...once we agree that niche features have a place in CRG.

> That requires that the data manager set it all up in advance though

Why? If it isn't set up in advance, the situation is no worse than what the PR provides now. And once it is set up, the situation improved.

> Events with hundreds/thousands of attendees are a thing.

And you want all of them to access the scoreboard? Then your Pseudo-UIDs won't be collision free either.

> Auto-incrementing ids are not a panacea

No but in this case they make the purpose of the auto-generated name way more obvious and for the number of devices we are realistically facing at least as easy to remember as your two random words.

brian-brazil:
> I'm talking about individual rules/settings that you argued to leave out as being too niche to be worth the extra row/button/dropdown.

I'm not seeing a problem here. It's natural that we reevaluate features over time, and that things get both added and removed as the software evolves.

> ...once we agree that niche features have a place in CRG.

I meant that this PR is self-contained and makes sense standalone, that is that it doesn't depend on future work.

> And you want all of them to access the scoreboard? Then your Pseudo-UIDs won't be collision free either.

Probably not wise for that big an event, though the WiFi will fall over before the scoreboard does (quick numbers I did previously indicated that even the old laptops we use in DRD can handle around a hundred connections, which is the typical limit of an enterprise-class WiFi AP).

The IPs would also help distinguish, though they're there more for network debugging. A future PR will probably hide the IPs by default, once we have logs. Collisions aren't the end of the world here.

However since you're so worried about it I've adjusted the code to work up to about 9k devices without collision, and to fallback to UUID at that point.

frank-weinberg:
> However since you're so worried about it

I'm not worried about collisions. I'm saying that these names are potentially confusing (as they may appear to bear a meaning) while they do not provide any benefit over autoincrement numbers in realistic scenarios. And I'm criticising that you continue to show a generated name to users after there is a user provided name.

Also your new code is horribly inefficient when the number of devices gets large.

brian-brazil:
> as they may appear to bear a meaning

I tried to avoid any words that may cause an incorrect meaning to be inferred, they're designed to be meaningless but memorable.

> And I'm criticising that you continue to show a generated name to users after there is a user provided name.

I'd added that, I had intended to keep it to the Devices page.

> while they do not provide any benefit over autoincrement numbers in realistic scenarios.

I think we're going to have to disagree on that point.

> Also your new code is horribly inefficient when the number of devices gets large.

I had done the math, and I'm not worried about it. Having a spike of more than 100 new devices a second seems unlikely. Once there's logging, the GC can also be made more aggressive for uninteresting devices too. If someone actually manages to hit it I'll make it more complicated and efficient at that point, but for the data sizes we're dealing with it'd be a premature optimisation.

PR #409, opened 2019-12-08
Fix buttons in Up/Donwload tab

frank-weinberg:
At some point during the lifetime of the screen the templates for new
items would break, causing items added later to not have proper download
and remove buttons. (Stopping and restarting the backend seems to
reliably cause this, but there are definitley other ways.) By adding the
elements that get lost after cloning, the buttons will reliably be as intended.

Also remove some dead code.


brian-brazil:
I recall coming across something weird here, but didn't get around to debugging.

PR #411, opened 2019-12-09
Add indicators for unserved/serving.

### LINKED ISSUE #410: Check screen usability for color blind users
Last weekend I had a PLT who had a very hard time noticin the light blue highlight on the shaded rows of the PLT screen. This caused problems as the source of skaters not being reset to the bench was unclear.

brian-brazil:
The current colours don't work for those who are red/green
colourblind, so add a symbol which should work for everyone.

Fixes #410

### DIFF
--- a/html/components/plt-input.css
+++ b/html/components/plt-input.css
@@ -56,6 +56,12 @@ body:not(.AllowSelect) .PLT {
 
 .PLT.Team .Unserved { background-color: #0FF; }
 .PLT.Team .Serving { background-color: #70ff70; }
+.PLT.Team .Penalty .Sitting, .PLT.Team .Penalty .Box { position: relative;}
+.PLT.Team .Penalty .Sitting::before, .PLT.Team .Penalty .Box::before {
+    font-weight: normal; position: absolute; left: .2em; top: .15em;
+ }
+.PLT.Team .Penalty .Sitting.Unserved::before, .PLT.Team .Penalty .Box.Unserved::before { content: "!"; }
+.PLT.Team .Penalty .Sitting.inBox::before, .PLT.Team .Penalty .Box.Serving::before { content: "\203E" /* overline */; }
 .PLT.Team .OnTrack:not(.Advance) { background-color: #70ff70; }
 .PLT.Team tr[role='Pivot'] .OnTrack:not(.Advance) { background-color: #00ee00; }
 .PLT.Team tr[role='Jammer'] .OnTrack:not(.Advance) { background-color: #00cc00; }


frank-weinberg:
I don't like the new placement. The symbols are almost hiding in the border now.
Also there is more room in the Penalty field than the Jam field.
My suggestion: Top: 0, left: .15em, size at 100%,  underscore instead of en-dash - that works better in terms of size and transitions nicely into the period.

brian-brazil:
> I don't like the new placement. The symbols are almost hiding in the border now.

IWB would always have the dot in the bottom right, and that means then having it in the 2nd row so it'll like up on the ePLT.

I think we'd also want to keep the dot unobtrusive if you're doing ePT only.

> underscore instead of en-dash

Problem with the underscore is that if we're in on the bottom it merges with the border. An endash is about the same width as an underscore. a hyphen would be too difficult to see/distinguish from a full stop. It's all the same with a little css.

frank-weinberg:
>IWB would always have the dot in the bottom right

No. "underneath the penalty code" as per Standard procedures (though I've seen it on all sides of the letter). And while I don't think underneath is reasonable for our display, it's associated with the code, not the jam, so having it in the code field is appropriate.

> I think we'd also want to keep the dot unobtrusive

But we should keep it visible. And tucked into the corner as it currently is, it isn't really visible.

brian-brazil:
> "underneath the penalty code"

Under and to the right is underneath, though IWB is much rarer today.

> it's associated with the code, not the jam

The information we have here associated with the penalty, neither the code or jam.

> And tucked into the corner as it currently is, it isn't really visible.

It's going to be in some corner.

frank-weinberg:
>Under and to the right is underneath

OED disagrees: https://www.lexico.com/en/definition/underneath

> It's going to be in some corner.

If the character is at the top, the actual dot will bewell away from the corner. and if additionally spacing to the side is slightly incresed, the characters will actually be separated from the border on smaller screens.

brian-brazil:
I realised overnight that we shouldn't have the dot at all. It's a holdover from paperwork, that isn't needed electronically (just like for SBO we don't show a dot for trips once we have a score). It'd also be confusing on the eIWB views in the common case where Assume Penalties Served is enabled as there's only ePT.

So we can do top-left then, as it's not quite a IWB-y. Adjusted accordingly, with an overline rather than an underscore.

frank-weinberg:
I'd still put the left margin at .2em to clearly separate the ! from the border. It's supposed to support a highlight after all. (And if we use the .1em from the top for both symbols, the ! will be horizontally  aligned with the penalty code while still fitting atop the 'Box'.)

brian-brazil:
Changed, .15em from top looks a little better to me (and happens to align with the code on my screen, .2 is slightly below).

> It's supposed to support a highlight after all.

There is no highlight on this. I removed them as it looked weird.

PR #412, opened 2019-12-09
Update Jammer name on jam start

frank-weinberg:
This fixes a problem where Jammer names would sometimes not be
displayed.

### DIFF
--- a/html/views/overlay/index.html
+++ b/html/views/overlay/index.html
@@ -32,24 +32,25 @@
 					</div>
 					<div class="JammerWrapper">
 						<div class="JammerBox ShowInJam Clock ColourTeam1">
-							<div class="JammerName VerticalCenter Box" sbDisplay="DisplayLead,StarPass,Position(Jammer),Position(Pivot),Skater" sbModify="jammer"></div>
+							<div class="JammerName VerticalCenter Box" sbDisplay="DisplayLead,StarPass,Position(Jammer).Name,Position(Pivot).Name,/ScoreBoard.InJam" sbModify="jammer"></div>
 						</div>
 					</div>
-					</div>
-					<div class="Team2" sbContext="Team(2)">
-						<div class="Team Row Wrapper barBackgroundBottom">
-							<div class="Indicator ColourTeam2 Box">&nbsp;</div>
-							<div class="Name Box" sbDisplay="AlternateName(overlay),Name"></div>
-							<div class="Score Box" sbDisplay="Score"></div>
-							<div class="Mini Box"><div class="JamScore" sbDisplay="JamScore"></div><div class="LeadJammer Lead">&#9733;</div></div>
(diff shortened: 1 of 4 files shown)


html/views/overlay/index.html
```diff
@@ -32,24 +32,25 @@
 					</div>
 					<div class="JammerWrapper">
 						<div class="JammerBox ShowInJam Clock ColourTeam1">
-							<div class="JammerName VerticalCenter Box" sbDisplay="DisplayLead,StarPass,Position(Jammer),Position(Pivot),Skater" sbModify="jammer"></div>
+							<div class="JammerName VerticalCenter Box" sbDisplay="DisplayLead,StarPass,Position(Jammer).Name,Position(Pivot).Name,/ScoreBoard.InJam" sbModify="jammer"></div>
```
brian-brazil, line 35: Why the /  ?

frank-weinberg: To indicate that this should not be prefixed with the sbContext

brian-brazil: Ah, github had hidden that diff.  A . might be more obvious, and it's how DNS does it.

frank-weinberg: Using a . to indicate that this will not take the prefix which ends in a . feels unintuitive to me.
(I did consider prefixing everything that takes the prefix with a . but figured that would be an unnecessary breaking change).

brian-brazil: At this point I'm wondering if we should keep using sbDisplay/sbModify here. At some point doing things via WS.Register is going to be clearer.

frank-weinberg: Given that the entire screen is done that way, I'd keep it.

brian-brazil: It's stretching the feature more and more though, and not all of that screen is done that way. It's also only this screen and the overlay hiding extensive business logic behind sbModify rather than its own associated .js file. It's one thing to do simple number conversion, but I think we're well beyond that at this point.

frank-weinberg: sbModify is just calling functions from the associated js files. (That common functions have been factored into common.js is independent of sbModify.)

brian-brazil: I'm not liking the scope creep here, I was already find this way of doing things with sbModify to be a bit inconsistent and unintuitive and this change pushes it further in this direction.

brian-brazil: The other thing is that we already have CSS stuff to cover what should be shown when certain clocks are running, so we could rely on that rather than adding a new way to get it.

brian-brazil: Ugh, the CSS doesn't quite cover it currently. I had a reworking in mind for `clockRunner` to make this all saner (make Lineup the default clock) but it was too high risk coming up to 4.0.0.

I can clean this all up later.

PR #413, opened 2019-12-14
Fix handling of fieldings around Undo.

### LINKED ISSUE #397: Un-stop jam doesn't reset fieldings and flags
Start a jam with a jammer selected, mark lead, stop jam, unstop jam.

You'd expect the jammer to still be fielded, and Call to be unmarked - however that doesn't happen. This seems to be due to ScoreBoard.Team not getting updated, which I suspect is related to updateTeamJams using Flags.INTERNAL.

frank-weinberg:
Fixes #397

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
@@ -570,6 +570,8 @@ public class ScoreBoardImpl extends ScoreBoardEventProviderImpl implements Score
         ScoreBoardClock.getInstance().rewindTo(snapshot.getSnapshotTime());
         if (getCurrentPeriod() != snapshot.getCurrentPeriod()) {
             if (getCurrentPeriod().getAll(Period.NChild.JAM).size() > 0) {
+                // We're undoing a period advancement. Move the upcoming Jam
+                // (and any associated Fieldings) back out before deleting the period.
                 Jam movedJam = (Jam) getCurrentPeriod().getFirst(Period.NChild.JAM);
                 getCurrentPeriod().remove(Period.NChild.JAM, movedJam);
                 movedJam.setParent(this);
--- a/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
@@ -568,22 +568,17 @@ public class ScoreBoardImpl extends ScoreBoardEventProviderImpl implements Score
     }
     protected void restoreSnapshot() {
         ScoreBoardClock.getInstance().rewindTo(snapshot.getSnapshotTime());
-        if (getCurrentPeriod() != snapshot.getCurrentPeriod() &&
-                getCurrentPeriod().getAll(Period.NChild.JAM).size() > 0) {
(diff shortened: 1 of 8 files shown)


src/com/carolinarollergirls/scoreboard/core/impl/TeamImpl.java
```diff
@@ -298,10 +288,13 @@ public TeamSnapshot snapshot() {
     public void restoreSnapshot(TeamSnapshot s) {
         synchronized (coreLock) {
             if (s.getId() != getId()) {	return; }
-            for (ValueWithId skater : getAll(Child.SKATER)) {
-                ((Skater)skater).restoreSnapshot(s.getSkaterSnapshot(skater.getId()));
-            }
+            boolean ended = s.getFieldingAdvancePending();
```
brian-brazil: You only use this once, no need for a variable.

src/com/carolinarollergirls/scoreboard/core/impl/ScoreBoardImpl.java
```diff
@@ -568,22 +568,17 @@ protected void createSnapshot(String type) {
...
     protected void restoreSnapshot() {
         ScoreBoardClock.getInstance().rewindTo(snapshot.getSnapshotTime());
-        if (getCurrentPeriod() != snapshot.getCurrentPeriod() &&
-                getCurrentPeriod().getAll(Period.NChild.JAM).size() > 0) {
-            Jam movedJam = (Jam) getCurrentPeriod().getFirst(Period.NChild.JAM);
-            getCurrentPeriod().remove(Period.NChild.JAM, movedJam);
-            movedJam.setParent(this);
-            set(Value.UPCOMING_JAM, movedJam);
+        if (getCurrentPeriod() != snapshot.getCurrentPeriod()) {
+            if (getCurrentPeriod().getAll(Period.NChild.JAM).size() > 0) {
```
brian-brazil, line 572: This is already here, but could do with a quick comment about what it's doing.

src/com/carolinarollergirls/scoreboard/core/impl/TeamImpl.java
```diff
@@ -298,10 +288,13 @@ public TeamSnapshot snapshot() {
...
+            set(Value.FIELDING_ADVANCE_PENDING, ended);
             updateTeamJams();
+            if (scoreBoard.isInJam()) {
+                set(Value.CALLOFF, false);
```
brian-brazil, line 294: This doesn't feel right, shouldn't we restore rather than set?

E.g. start jam -> lead -> unstart jam -> start jam incorrectly has lead set on the new jam.

frank-weinberg: Call and Inj: If the Jam is ended via the Calloff or Injury button, the field will be set before the snapshot is taken. And there are no circumstances where setting either before the Jam has ended is correct.

Lead and Lost: If a Jam has gone on long enough that one of those has been marked, there's a significant chance that actually the undo is in error. And if the operator cancels the undo (i.e replace Start Jam with Start Jam), it'd be surprising if the marks were dropped, especially as scoring trips or SP aren't dropped either. (And I'd be vehemently opposed to removing trips and thus possibly points on a clock action.)

I have addressed the other remarks.

PR #417, opened 2019-12-15
Don't mark Call when stopping via Inj button

### LINKED ISSUE #416: If you press Inj while there is a lead jammer, Call is also set

frank-weinberg:
Fixes #416


brian-brazil:
Can you add a unittest?

frank-weinberg:
done

PR #421, opened 2020-01-10
Fix the media select dropdowns

### LINKED ISSUE #419: [BUG] v 4.0.1 no more responds when uploading an image
Hi all,

found that CRG do no more answer when dropping an image in team logo.

I'm pretty sure it's due to the ' character in the image name.

The logo png is named "Track'ass Roller Derby de Tours.png"
If i removed it from the html/images/team path it all rework (after restarting the java apps).
Uploading it via the upload button and selecting path do the same.
The logo comes from a version 3.9.5 teams logo path and was working.
Removing the ' characters from the filename remove the issue.

Have a nice debug christmas holliday :)

frank-weinberg:
Closes #419  following @IWishICouldUseComputers suggestion

Also fixes a bug were deleted/renamed files were not removed and
changes of internal name lead to duplication.

### DIFF
--- a/html/components/teams-tab.js
+++ b/html/components/teams-tab.js
@@ -66,22 +66,11 @@ function createTeamsTab(tab) {
 		WS.Set(getPrefix() + ".Name", teamName.val());
 	});
 	var waitingOnUpload = "";
-	var logoSelect = $("<select>").append($("<option value=''>No Logo</option>"))
-		.appendTo(controlTable.find("td:eq(1)"));
-	logoSelect.change(function() {
-		WS.Set(getPrefix() + ".Logo", logoSelect.val() == "" ? "": "/images/teamlogo/" + logoSelect.val())
-	});
+	var logoSelect = mediaSelect(getPrefix() + '.Logo', 'images', 'teamlogo', 'Logo').appendTo(controlTable.find("td:eq(1)"));
 	WS.Register("ScoreBoard.Media.Format(images).Type(teamlogo).File(*).Name", function(k, v) {
-		var val = logoSelect.val();	// Record this before we potentially remove and re-add it.
-		if (v != null) {
-			if (waitingOnUpload == k.File) {
-				val = k.File;
-				waitingOnUpload = "";
-			}
-			logoSelect.children("[value='"+k.File+"']").remove();
-			var option = $("<option>").attr("name", v).attr("value", k.File).text(v);
-			_windowFunctions.appendAlphaSortedByAttr(logoSelect, option, "name", 1);
-			logoSelect.val(val);
+		if (v != null && waitingOnUpload == k.File) {
+			waitingOnUpload = '';
+			logoSelect.val(escape('/images/teamlogo/' + k.File));
 			logoSelect.change();
 		}
 	});
--- a/html/json/WSutils.js
+++ b/html/json/WSutils.js
@@ -17,9 +17,9 @@ function toggleButton(key, trueText, falseText) {
 function mediaSelect(key, format, type, humanName) {
 	var select = $("<select>").append($("<option value=''>No " + humanName + "</option>"));
 	WS.Register("ScoreBoard.Media.Format(" + format + ").Type(" + type + ").File(*).Name", function(k, v) {
-		select.children("[value='"+k.File+"']").remove();
+		select.children("[value='" + escape("/" + format + "/" + type + "/" + k.File) + "']").remove();
 		if (v != null) {
-			var option = $("<option>").attr("name", v).val("/" + format + "/" + type + "/" + k.File).text(v);
+			var option = $("<option>").attr("name", v).val(escape("/" + format + "/" + type + "/" + k.File)).text(v);
 			_windowFunctions.appendAlphaSortedByAttr(select, option, "name", 1);
 			select.val(WS.state[key]);
 		}
--- a/html/settings/teams/index.html
+++ b/html/settings/teams/index.html
@@ -4,6 +4,7 @@
 		<title>Teams Editor</title>
 		<script type="text/javascript" src="/external/jquery/jquery.js"></script>
 		<script type="text/javascript" src="/json/core.js"></script>
+		<script type="text/javascript" src="/json/WSutils.js"></script>
 		<script type="text/javascript" src="/javascript/utils.js"></script>
 		<script type="text/javascript" src="/components/teams-tab.js"></script>
 


brian-brazil:
Is there anywhere else we need to do this?

Having to do explicit escaping like this is also a smell, we should prefer using an API that doesn't require it as is the practice with e.g. SQL.

frank-weinberg:
> Is there anywhere else we need to do this?

I did not search the code. Off the top of my head Logos are the only file type where we generate dynamic HTML based on filenames in other places, so maybe just copy the offending file into the logo ordner and see if anything freezes. (I won't be able to do this before sunday.)

> we should prefer using an API that doesn't require it

Well, I'm still assuming that this is a bug in the jQuery version we are using (especially as not all filenames with apostrophes cause problems), so we could look into updating. But that's a significantly bigger change given that we are 2 major releases behind what is currently supported upstream.

brian-brazil:
> Well, I'm still assuming that this is a bug in the jQuery version we are using

I can't see how this is is a jQuery bug, we're the ones hand-constructing invalid selectors. This is just like SQL injection.

So the question is, is there anywhere else we need escaping due to having selectors with user-provided input?

frank-weinberg:
> we're the ones hand-constructing invalid selectors

Single quotes are allowed in attribute values, so in which way is the selector invalid?

> So the question is, is there anywhere else we need escaping due to having selectors with user-provided input?

If you want to audit the code in that regard, go ahead. I don't have time to do that at the moment.

brian-brazil:
> Single quotes are allowed in attribute values, so in which way is the selector invalid?

Consider how you would parse `[value='foo'bar']` given the unmatched quotes.

frank-weinberg:
I tried changing to `[value="foo'bar"]` when I was looking into the problem (as that is the preferred syntax anyways) - same result. Plus, as was reported, some file names with single quotes did not cause the problem. And either way I'd expect an error message or nothing being matched, not an infinite loop.

IWishICouldUseComputers:
Hi @frank-weinberg I have tested this branch and found that this change resolves the issue for "fullscreen" and "sponsor_banner" folders but the error still happen when introducing a file with a long file and and quote mark in it to the **teamlogo** folder. I hope to have a chance to look at where this is happening later today.

@brian-brazil
> Having to do explicit escaping like this is also a smell, we should prefer using an API that doesn't require it as is the practice with e.g. SQL.

I would normally look to create a new issue to cover the work that is significantly larger than the scope of the original issue, on the basis that it may not make the priorities list? Do you want me to create one?

Regarding the security I would have thought that the JS should have no security implications, on the basis that it is client side? Have I miss understood? Is there server side JS as well as the Java API? - Perhaps I can catch up with you on this subject after DRD practice on Sunday.

frank-weinberg:
> I hope to have a chance to look at where this is happening later today.

There are team logo dropdowns in the teams tab and the team/time tab. Most likely it's one or both of these. Best course of fixing would likely be to convert them to also use the media select utility function.

IWishICouldUseComputers:
Spot on.
line 81 of teams-tab.js added escape: 
`logoSelect.children(escape("[value='"+k.File+"']")).remove();`
Resolves the immediate issue and testing I can see that the logo selected (with a troublesome file name) is presented on the main Team/Time screen and I can set it back to "No Logo".

> Best course of fixing would likely be to convert them to also use the media select utility function

I will take a look.

brian-brazil:
> Regarding the security I would have thought that the JS should have no security implications, on the basis that it is client side?

XSS is XSS. Though we allow anyone who can get to the scoreboard to upload arbitrary HTML :(

IWishICouldUseComputers:
So, I have had a go at updating the logo interaction on the teams page, but I got distracted by the way it is already being done on the Team/Time page, and think that the way the UX works is quite nice and so was wondering if we could simply do the same, ie move the Team/Time logo interaction into a function in WSutils.js and then  use the same code for both pages.

The down side to this is that I would advocate for removing the upload functionality from the teams tab, on the basis that it already exists on the Up/Download tab, although with a fresh head I'm sure it would be fairly easy to keep it if we think it would have a detrimental impact on the SBO.

The upside is a minor reduction in code we need to maintain and improved consistency in the way logos are assigned between the two tabs.

Existing:
before

Proposed:
- Removes the upload option.
- Selected logo is shown once the selection has been made (as per Team / Time page).
after

@frank-weinberg - if you want to get a fix put together on Sunday don't let my learning using this hold you up, I am still very much finding my feet and appreciate that you will want to get a fix published, I have a busy next few days so might not get back to it properly until Wednesday.

This is where I got to with the above described approach, I wanted to get it working before trying to move to a common function. (once my github account is unlocked I will stop posting code as attachments =)
[teams-tab.js.txt](https://github.com/rollerderby/scoreboard/files/4048768/teams-tab.js.txt)

brian-brazil:
> The down side to this is that I would advocate for removing the upload functionality from the teams tab, on the basis that it already exists on the Up/Download tab

This was a quality of life thing I added, I don't think we should be quick to remove it. How to upload images has always been a common user support question.

> I am still very much finding my feet and appreciate that you will want to get a fix published

I don't think we're in a hurry to do another release. Compared to all the bugs we've had in published releases over the years, this is a pretty minor one.

IWishICouldUseComputers:
> This was a quality of life thing I added, I don't think we should be quick to remove it. How to upload images has always been a common user support question.

Makes sense.

> I don't think we're in a hurry to do another release. Compared to all the bugs we've had in published releases over the years, this is a pretty minor one.

Grand I will aim to take a look on Wed.

frank-weinberg:
I have migrated the selector on the teams tab and verified that the selector on the Team/Time tab already uses the utility function. That should cover all media selectors.

engineerlondon:
I will give this a test today, looking at the change it makes sense and I’m glad you did it because I think it would have taken me a while.

html/components/teams-tab.js
```diff
@@ -66,22 +66,11 @@ function createTeamsTab(tab) {
...
-				val = k.File;
-				waitingOnUpload = "";
-			}
-			logoSelect.children("[value='"+k.File+"']").remove();
-			var option = $("<option>").attr("name", v).attr("value", k.File).text(v);
-			_windowFunctions.appendAlphaSortedByAttr(logoSelect, option, "name", 1);
-			logoSelect.val(val);
+		if (v != null && waitingOnUpload == k.File) {
+			waitingOnUpload = '';
+			logoSelect.val(escape('/images/teamlogo/' + k.File));
```
brian-brazil, line 73: This feels wrong, why is an encdoded/escaped value being put in a variable? I'd expect any escaping to be done right before usage, not in an entirely different file as that's pretty fragile.

html/json/WSutils.js
```diff
@@ -17,9 +17,9 @@ function toggleButton(key, trueText, falseText) {
 function mediaSelect(key, format, type, humanName) {
 	var select = $("<select>").append($("<option value=''>No " + humanName + "</option>"));
 	WS.Register("ScoreBoard.Media.Format(" + format + ").Type(" + type + ").File(*).Name", function(k, v) {
-		select.children("[value='"+k.File+"']").remove();
+		select.children("[value='" + escape("/" + format + "/" + type + "/" + k.File) + "']").remove();
```
brian-brazil, line 20: escape() is for URLs, I doubt that's that appropriate for a CSS selector. jQuery has a escapeSelector() function, however escaping should be a last resort as it's prone to mistakes like these. The filter function with a function would be best here, given that jQuery lacks that sort of stuff that has become standard for dealing with SQL injections.

brian-brazil:
Also, this is not a complete fix as there's other places we're using user provided file names.

Security-related fixes like these need to be systematic - not ad hoc.

engineerlondon:
> this is not a complete fix as there's other places we're using user provided file names.

I suggest a new issue that lists the places or describes the issue that need fixing should be created. Allowing some up front discussion on the best solution.

This resolves the problem raised in the issue of the UI crashing. And in my opinion fixing that and then iterating will be easier than going straight from broken to systemic change? Unless the fix should be applied elsewhere?

Perhaps creating the issue and discussing the best resolution will clarify the right way forward for this specific issue as well.

brian-brazil:
> Unless the fix should be applied elsewhere?

Exactly. The bare minimum when dealing with an issue like this is to check for other places that are affected, and fixing them too.

frank-weinberg:
> I suggest a new issue that lists the places or describes the issue that need fixing should be created.
> This resolves the problem raised in the issue of the UI crashing.

Fully agree. This PR is about fixing a UI crash. It does achieve this. Trying to get a full audit of how user provided strings are handled into it is massive scope creep and unwarranted as the potential issues that such an audit may find are neither introduced nor worsened by this PR.

html/json/WSutils.js
frank-weinberg: How about you provide a better solution then instead of just unconstructively complaining how our fixes of a bug you introduced are insufficient?

engineerlondon:
On this branch the manual (ie adding via drag and drop on file system) of Custom Screens page copes with the type of file name that was proven to cause an issue:
image

Videos also seems happy:
image

And it resolves the raised issue:
image

> escapeSelector vs escape: "The main difference is that $.escapeSelector() can be reliably used in all of jQuery's supported browsers" - https://api.jquery.com/jQuery.escapeSelector

escapeSelector also resolves this issue as can be expected and looks like it will deliver slightly better cross browser support.

frank-weinberg:
`escapeSelector` was added in jQuery 3.0. We are on 1.6.

brian-brazil:
> This PR is about fixing a UI crash. It does achieve this. 

This is an injection bug, and which are usually security issues (think SQL injection). There's a set of standard practices for security issues, and handling them whack-a-mole as proof-of-concepts are provided is one of the worst responses you can have to a security issue.

In addition the fix proposed here is possibly incorrect, due to using the wrong escaping. It's also fragile, and thus likely to regress in future.


This class of issues (css selector injection) should be fixed comprehensively using best software engineering practices, not just the bare minimum to solve one specific bug report.

> Trying to get a full audit of how user provided strings are handled into it is massive scope creep

I never proposed that, I'm only talking about user input that feeds into CSS selectors here.

frank-weinberg:
> This is an injection bug, and which are usually security issues

So far nobody has provided any way this would open an attack vector. Especially not one that isn't open via intentional features anyways.

> This class of issues (css selector injection) should be fixed comprehensively using best software engineering practices, not just the bare minimum to solve one specific bug report.

So you are proposing we first define the requirements for at least the affected parts of the scoreboard and then proceed by deriving the various levels of specification along with a set of tests that will afterwards verify that the implementation follows the specification? And during these several months the bug remains open because apparently a quick fix would make the thorough work impossible?

engineerlondon:
Unfortunately the change seems to be stopping the Teams tab logo from persisting, to re-produce:
1. Copy the flag image "scoreboard\html\images\teamlogo\American Flag.jpg" to the folder: "scoreboard\html\images\teamlogo"
Login as a user, navigate to "http://localhost:8000" and select "Scoreboard Operator Panel" then provide a user eg: "Test".
2. Click "Teams" Tab, Click the drop down select "Black", Click the drop down (should read "No Logo"), select "American Flag".
3. Click the drop down select "White".
4. Click the drop down select "Black"
Expect: Team Black Logo to be set to "American Flag", Result: "No Logo"

frank-weinberg:
I won't spend any more time on this when it will be denied based on @brian-brazil 's gut feeling anyways.

brian-brazil:
> So far nobody has provided any way this would open an attack vector. 

I never said this was a security issue, if you can upload a file you can do plenty of other bad stuff. However this sort of problem usually indicates a security issue, and thus the practices for properly dealing with injection come from the security domain (they're good software engineering practices in any case).

> So you are proposing we first define the requirements for at least the affected parts of the scoreboard and then proceed by deriving the various levels of specification along with a set of tests that will afterwards verify that the implementation follows the specification?

I never proposed anything like that, nor would that typically be part of a solution. What I'm asking for is simple, find all the places where we might have CSS selector injection. That shouldn't take more than a few minutes to do for someone experienced with the codebase.

> gut feeling

This is not a gut feeling, this is attempting to avoid practices which are known to be problematic.

@engineerlondon if you'd like to take this over I'm happy to take you through it.

Issue #420, opened 2019-12-31
Custom penalties not being picked up


brian-brazil:
When you define a new ruleset pointing to a custom penalty file and then use it, the standard wftda 2018 penalties are being used still.

PR #422, opened 2020-01-14
Fix rounding of clocks

frank-weinberg:
The WS display would use the same rounding function as the XML display
but without the $sb objects, causing it to always round up instead of
opposite to the clock direction. Change it to use the established
behaviour.

This fixees the bug of lineup and period occasionally adding to 31 with a red lineup clock.

### DIFF
--- a/html/components/team-time-tab.js
+++ b/html/components/team-time-tab.js
@@ -296,7 +296,7 @@ function createPeriodEndTimeoutDialog(td) {
 	WS.Register("ScoreBoard.Clock(Period).Running");
 	var checkTimeFunction = function(v) {
 		if (!applying) return;
-		var currentSecs = Number(_timeConversions.msToSeconds(v));
+		var currentSecs = Number(_timeConversions.msToSeconds(v, isTrue(WS.state['ScoreBoard.Clock(Period).Direction'])));
 		var targetSecs = Number(waitDiv.find("span.TargetSeconds").text());
 		if (currentSecs > targetSecs) {
 			return;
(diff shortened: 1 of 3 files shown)


html/components/team-time-tab.js
```diff
@@ -1145,10 +1145,11 @@ function createTimeoutDialog() {
 		}
 		if (v != null && row.length == 0) {
 			var jam = Number(WS.state[prefix+'.PrecedingJamNumber']);
-			var dur = isTrue(WS.state[prefix+'.Running']) ? 'Running' : _timeConversions.msToMinSec(WS.state[prefix+'.Duration']);
+			var dur = isTrue(WS.state[prefix+'.Running']) ? 'Running' :	_timeConversions.msToMinSec(WS.state[prefix+'.Duration'], true);
```
brian-brazil, line 1148: Is it right to round these up? Timeouts (usually) use a count up clock, so this can display a different value to what was on the scoreboard.

brian-brazil (commented):
My plan for this was to move it all into the backend, both to avoid duplication and to allow us to unittest all of this. That'd give us confidence that the various corner cases were correctly handled.

frank-weinberg:
> My plan for this was to move it all into the backend

That makes sense longer term, but for now I'd like to get the immediate bug fixed and a 4.0.2 out of the door ASAP given that this has game impact.

html/components/team-time-tab.js
frank-weinberg: The original behaviour was to always round up when we are not dealing with a value straight from a clock. I decided to keep that for now.

Given that the behaviour for clocks is to round opposite to the running direction it would probably make sense to round any duration down, but I'd consider that a separate change from this bugfix.

brian-brazil:
If you're doing a release, I've one tiny bugfix to get in too.

frank-weinberg:
If #421 gets done tomorrow, we can include that one as well. And I'd have a look if I can fix #420 tonight.

brian-brazil:
I can grab 420 if you're looking at 421. We mainly need to check if there's any other ids that are user provided rather than being uuids or numbers.

PR #423, opened 2020-01-14
Remove unused overlay video elements.

brian-brazil:
Some streaming software renders these, covering most of the screen.

### DIFF
--- a/html/views/overlay/index.js
+++ b/html/views/overlay/index.js
@@ -101,9 +101,7 @@ function initialize() {
 	});
 
 	WS.Register([ 'ScoreBoard.Settings.Setting(Overlay.Interactive.BackgroundColor)' ], function(k,v) {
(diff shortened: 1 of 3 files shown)


html/views/overlay/index.js
```diff
@@ -102,7 +102,7 @@ function initialize() {
 
 	WS.Register([ 'ScoreBoard.Settings.Setting(Overlay.Interactive.BackgroundColor)' ], function(k,v) {
 		if (k == 'ScoreBoard.Settings.Setting(Overlay.Interactive.BackgroundColor)') {
```
frank-weinberg: Isn't this ìf` superfluous? We're registering on only this key. (And if it isn't, a comment might be warranted.)

brian-brazil: Fixed

PR #426, opened 2020-01-18
Always show a jam score if there is one.

brian-brazil:


frank-weinberg:
But points during initial never happen. So what does this PR even fix? ;)

brian-brazil:
Overtime jams :)

JeneralPain:
Except that’s during a triggered OTJ not during normal jams

brian-brazil:
Overtime jams aren't that special, and the scoreboard view doesn't even know that they happen.

PR #430, opened 2020-01-30
Close trip editor after inserting a trip

frank-weinberg:
This makes the button's behaviour consistent with the other buttons in
the row. It is also the common use case.


brian-brazil:
One thing I've been considering here is arrows to move you back/forward in terms of which trip you're looking at. Particularly in the rare case you've ended up with 10+ trips, but also to make it easier to get to the initial trip.

frank-weinberg:
I have been thinking about this as well. But the way the dialogs are set up, this is relatively complicated to do, so I left it out for the time being. (Same for the dialog on the LT sheet.)

PR #431, opened 2020-02-01
Fix overlay points-per-jam header and roster logo size

MarbleDice:
This commit addresses two issues:

1. The header on the points-per-jam panel of the broadcast overlay is misaligned

This is because the `NoLogo` class was assigned to an h1 element, instead of the div that contains an h1 element. The `NoLogo` class is assigned 4 other times throughout the overlay, and always to a div element that contains an h1.

ppj0

ppj1

2. The team logos on the roster panels of the broadcast overlay are too large

The height of logos on the roster and penalties panels are controlled through separate CSS styles. I adjusted the size of the roster logo to match the gap on the penalties page.

tr0

tr1


brian-brazil:
Thanks for the detailed report and fix!

PR #432, opened 2020-02-01
Hide non-in-game skaters on the overlay penalties panel

MarbleDice:
Skaters who are not-in-game are not removed from the broadcast overlay's penalties panel. This limits the usefulness of role assignment since, unlike the roster panel, the penalties panel does not handle more than ~17 skaters. Skaters who are not-in-game also cannot receive penalties, so there is no need to display them.

This PR makes a few changes to accommodate this:

1. When a skater's role is updated, the skater's row (in the penalties panel only) will be hidden (or revealed, if they have been added back into the game).
2. Rows which are not visible are sorted to the end, so as not to interfere with the alternating row color.
3. The border-radius is applied to the container rather than the first and last skater rows, so the corner appearance does not change with row order/visibility.

roster

penalties

### DIFF
--- a/html/views/overlay/index.js
+++ b/html/views/overlay/index.js
@@ -28,7 +28,8 @@ function initialize() {
 	WS.Register( [
 			'ScoreBoard.Team(*).Skater(*).Name',
 			'ScoreBoard.Team(*).Skater(*).Number',
-			'ScoreBoard.Team(*).Skater(*).Flags' ], function(k,v) {
+			'ScoreBoard.Team(*).Skater(*).Flags',
+			'ScoreBoard.Team(*).Skater(*).Role'], function(k,v) {
 				var me = '.RosterTeam' + k.Team + ' .Team' + k.Team + ' .Skater[data-skaterId=' + k.Skater + ']';
 				var mb = '.PenaltyTeam' + k.Team + ' .Team' + k.Team + ' .Skater[data-skaterId=' + k.Skater + ']';
 				if (v == null) {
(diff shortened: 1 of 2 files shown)


html/views/overlay/index.js
```diff
@@ -38,18 +39,27 @@ function initialize() {
 				}
 				ensureSkaterExists(k.Skater, k.Team);
 
-				if(k.field == 'Number')	{ updateSkaterNumber(me,mb,v); }
-				if(k.field == 'Flags') {
+				if (k.field == 'Flags') {
 					$('.'+k.field, me).attr('data-flag', v);
 					$(mb).attr('data-flag', v);
-				} else if (v == "") {
-					c = $('.'+k.field, me).html('&nbsp;');
```
MarbleDice, line 46: I think these assignments to undeclared variable c are extraneous, but if there's a one-character global variable floating around triggering side effects then I could have introduced a bug here?

html/views/overlay/index.js
```diff
@@ -178,12 +188,13 @@ function ensureSkaterExists(skaterId, team) {
 	}
 }
 
-function updateSkaterNumber(me,mb,v) {
-	sv = v;
-	if(v == '' || v == '-' || v == null) { sv = 'ZZZB'; v = '-'; }
```
MarbleDice, line 183: Again, method parameter v here is assigned to but not referenced later so I believe this is extraneous.

html/views/overlay/index.js
brian-brazil: There shouldn't be. This code used to be all cryptic variables names, and most likely I missed this back when I was untangling it.

html/views/overlay/index.js
```diff
@@ -178,12 +188,13 @@ function ensureSkaterExists(skaterId, team) {
...
-	$('.Number',me).parent().attr('data-sort', sv);
-	$('.Number',mb).parent().attr('data-sort', sv);
-	$(me).parent().sortDivs(); $(mb).parent().sortDivs();
+function updateSort(sel) {
+	var skaterRow = $(sel);
+	var n = $('.Number', sel).text();
+	var sortValue = (skaterRow.hasClass('NoShow') ? '1' : '0')
```
brian-brazil: This is a bit hard to read, I'd suggest using an If and adding a comment.

MarbleDice: A comment documenting the compound sort is definitely a good call. I pushed changes and retested.

brian-brazil:
Thanks!

PR #434, opened 2020-02-06
Fix login dialog not being visible if you open the operator screen in…

brian-brazil:
… P2.

### DIFF
--- a/html/javascript/utils.js
+++ b/html/javascript/utils.js
@@ -85,6 +85,9 @@ _crgUtils = {
 				dialog.dialog("destroy");
 			}
 		});
+		// If we're in period 2, the dialog can end up off screen once all the
+		// SK rows finish loading. This ensures the dialog stays up top.
+		dialog.dialog("widget").css({"top": "5em", "position": "fixed"});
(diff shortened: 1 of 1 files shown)


html/javascript/utils.js
```diff
@@ -84,8 +84,11 @@ _crgUtils = {
 			close: function() { if (callback('default'))
 				dialog.dialog("destroy");
 			}
-		});
-	},
+    });
+    // If we're in period 2, the dialog can end up off screen once all the
+    // SK rows finish loading. This ensures the dialog stays up top.
+    dialog.dialog("widget").css({"top": "5em", "position": "fixed"});
+  },
```
frank-weinberg: It looks like indentation is inconsistent with the rest of the file.

brian-brazil: Ah, I did the replacement the wrong way around. Fixed.

frank-weinberg: The rows with brackets are still indented by spaces

brian-brazil: That should do it...

PR #436, opened 2020-02-08
Add Starpass Information to the overlay-view

reflexdd:
Starpass will be shown on the overlay with the same behaviours as on the scorebard-view. This works with Jammer names entered or not. Color of the SP is linked to the teamcolor.

### DIFF
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -9,7 +9,7 @@ function jammer(k, v) {
 	var inJam = isTrue(WS.state["ScoreBoard.InJam"]);
 
 	if (jammerName == null || jammerName == "") {
-		jammerName = leadJammer ? "" : "";
+		jammerName = leadJammer ? "Lead" : "";
 	if (pivotName == null)
 		pivotName = "";
 	}
@@ -61,8 +61,6 @@ function smallDescriptionUpdate(k, v) {
 	var lcn = WS.state["ScoreBoard.Clock(Lineup).Name"];
 	var tcn = WS.state["ScoreBoard.Clock(Timeout).Name"];
 	var ret = '';
-
-	// console.log(lc,tc,to,or,lcn,tcn);
 
 	$.each(["1", "2"], function (idx, id) {
 		tto = WS.state["ScoreBoard.Team(" + id + ").Timeouts"];
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -20,7 +20,7 @@ function jammer(k, v) {
 	};
 	$(".Team" + id + " .Lead").toggleClass("HasLead", (leadJammer && !starPass));
 	$(".Team" + id).toggleClass("HasJammerName", (jn != ""));
-	$(".Team" + id).toggleClass("starPass", (starPass != false));
+	$(".Team" + id).toggleClass("starPass", starPass);
 	return jn
 }
 
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -20,7 +20,7 @@ function jammer(k, v) {
 	};
 	$(".Team" + id + " .Lead").toggleClass("HasLead", (leadJammer && !starPass));
 	$(".Team" + id).toggleClass("HasJammerName", (jn != ""));
-	$(".Team" + id).toggleClass("IsStarPass", (starPass != false));
+	$(".Team" + id).toggleClass("starPass", (starPass != false));
 	return jn
 }
 
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -20,7 +20,7 @@ function jammer(k, v) {
 	};
 	$(".Team" + id + " .Lead").toggleClass("HasLead", (leadJammer && !starPass));
 	$(".Team" + id).toggleClass("HasJammerName", (jn != ""));
-	$(".Team" + id).toggleClass("starPass", starPass);
+	$(".Team" + id + " .Lead").toggleClass("HasStarPass", starPass);
 	return jn
 }
 
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -9,9 +9,8 @@ function jammer(k, v, ov) {
 	var inJam = isTrue(WS.state["ScoreBoard.InJam"]);
 
 	if (jammerName == null || jammerName == "") {
-		if(!ov)	jammerName = leadJammer ? "Lead" : "";
-	if (pivotName == null)
-		pivotName = "";
+		if (!ov) {jammerName = leadJammer ? "Lead" : "";}
+		if (pivotName == null) {pivotName = "";}
 	}
 
 	var jn = !starPass ? jammerName : pivotName;
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -9,7 +9,7 @@ function jammer(k, v, ov) {
 	var inJam = isTrue(WS.state["ScoreBoard.InJam"]);
 
 	if (jammerName == null || jammerName == "") {
-		if (!ov) {jammerName = leadJammer ? "Lead" : "";}
+		jammerName = (leadJammer && !ov) ? "Lead" : "";
 		if (pivotName == null) {pivotName = "";}
 	}
 
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -20,6 +20,7 @@ function jammer(k, v) {
 	};
 	$(".Team" + id + " .Lead").toggleClass("HasLead", (leadJammer && !starPass));
 	$(".Team" + id).toggleClass("HasJammerName", (jn != ""));
+	$(".Team" + id).toggleClass("IsStarPass", (starPass != false));
 	return jn
 }
 
@@ -132,7 +133,7 @@ function toClockInitialNumber(k, v) {
 			if (name != null && number != null)
 				ret = name.substring(0, 1) + number;
 
-			if (name == 'Period' && WS.state['ScoreBoard.Rulesets.CurrentRule(Period.Number)'] == 1) 
+			if (name == 'Period' && WS.state['ScoreBoard.Rulesets.CurrentRule(Period.Number)'] == 1)
 				ret = 'Game';
 		}
 	});
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -9,7 +9,7 @@ function jammer(k, v) {
 	var inJam = isTrue(WS.state["ScoreBoard.InJam"]);
 
 	if (jammerName == null || jammerName == "") {
-		jammerName = leadJammer ? "Lead" : "";
+		jammerName = leadJammer ? "" : "";
 	if (pivotName == null)
 		pivotName = "";
 	}
@@ -62,7 +62,7 @@ function smallDescriptionUpdate(k, v) {
 	var tcn = WS.state["ScoreBoard.Clock(Timeout).Name"];
 	var ret = '';
 
-	console.log(lc,tc,to,or,lcn,tcn);
+	// console.log(lc,tc,to,or,lcn,tcn);
 
 	$.each(["1", "2"], function (idx, id) {
 		tto = WS.state["ScoreBoard.Team(" + id + ").Timeouts"];
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -1,5 +1,5 @@
 
-function jammer(k, v) {
+function jammer(k, v, ov) {
 	id = getTeamId(k);
 	var prefix = "ScoreBoard.Team(" + id + ").";
 	var jammerName = WS.state[prefix + "Position(Jammer).Name"];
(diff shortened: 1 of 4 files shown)


reflexdd:
SP-Scoreboard-Overlay-Status-0
SP-Scoreboard-Overlay-Status-1
SP-Scoreboard-Overlay-Status-4
SP-Scoreboard-Overlay-Status-2
SP-Scoreboard-Overlay-Status-3

brian-brazil:
This looks pretty good.

I could see that being a little confusing when there's both a SP and a skater name. Could the separation be clearer without getting over complicated?

reflexdd:
I will see if i can make the separation a bit clearer.
Meanwhile I was playing a bit around with a re-styling of the overlay and found the solution quite nice having SP and Lead status on the left side.
SP-Scoreboard-Overlay-Status_V2-1

brian-brazil:
That would work if that element is a little wider as you have there (and gives us more space to make the TTO/OR indicators more readable in future). The square corners don't look the best though.

reflexdd:
Wich element would you make wider? The SP/Lead Box or the Team-Name Section?
We can try make the left boxes a little wider. We then have to think about what to show if there is no color assigned. Maybe a small grey separator to indicate that there could be a color assigned.  I will see if i find a good solution to give TO/OR more space.

reflexdd:
I made a little sketch on how I can imagine to have TO/OR more visible
scribble-scoreboard

brian-brazil:
> Which element would you make wider? The SP/Lead Box or the Team-Name Section?

The SP/Lead box. The current checked in version is a bit narrow for what you're proposing.

> I will see if i find a good solution to give TO/OR more space.

Let's not have too much scope creep here. We can discuss this in a future PR.

reflexdd:
Ok coming back to the original change. I would see two possibilities to make it more distinguishable. 
The first would be to add a grey border to the SP Box when skaternames are set. That's also the one I would probably prefer.
Screenshot_2020-02-15-Screenshot-3

The other one would be to use the same silver background for these boxes.
Screenshot_2020-02-15-Screenshot-1
Screenshot_2020-02-15-Screenshot-2

brian-brazil:
I think putting it in the left box like you suggested before looks best.

reflexdd:
Sure. Would you leave the Lead information where it is now or put it also in that box, like I suggested. I think having it on the left side makes it clearer to read.

brian-brazil:
Like you had in https://github.com/rollerderby/scoreboard/pull/436#issuecomment-584421068, but keeping the rounded corners.

reflexdd:
Changes are now made to display Lead and Star Pass in the left Box.
Screenshot_2020-02-16-Screenshot-1
Screenshot_2020-02-16-Screenshot-2
Screenshot_2020-02-16-Screenshot-3

html/views/common.js
```diff
@@ -20,6 +20,7 @@ function jammer(k, v) {
 	};
 	$(".Team" + id + " .Lead").toggleClass("HasLead", (leadJammer && !starPass));
 	$(".Team" + id).toggleClass("HasJammerName", (jn != ""));
+	$(".Team" + id).toggleClass("IsStarPass", (starPass != false));
```
brian-brazil: Why not just `starPass` ?

html/views/common.js
```diff
@@ -9,7 +9,7 @@ function jammer(k, v) {
 	var inJam = isTrue(WS.state["ScoreBoard.InJam"]);
 
 	if (jammerName == null || jammerName == "") {
-		jammerName = leadJammer ? "Lead" : "";
+		jammerName = leadJammer ? "" : "";
```
brian-brazil: Will this break the main scoreboard when there's a lead jammer we don't know the name of?

html/views/common.js
```diff
@@ -61,7 +62,7 @@ function smallDescriptionUpdate(k, v) {
 	var tcn = WS.state["ScoreBoard.Clock(Timeout).Name"];
 	var ret = '';
 
-	console.log(lc,tc,to,or,lcn,tcn);
+	// console.log(lc,tc,to,or,lcn,tcn);
```
brian-brazil: You can delete this.

html/views/overlay/index.js
```diff
@@ -81,18 +81,32 @@ function initialize() {
...
 			for(var r=0; r<style.rules.length ; r++ ) {
 				var dd = style.rules[r];
-				if(dd.selectorText == rd && dd.style[0] == ns) style.deleteRule(r);
+				if(dd.selectorText == rd && dd.style[0] == ns) {
+					style.deleteRule(r);
+				}
 			}
-			if(v != null) style.addRule(rd, ns + ': ' + v);
+
+			for(var r=0; r<style.rules.length ; r++ ) {
```
brian-brazil: This is getting messy.  Can we switch this to working like the PLT does?

html/views/common.js
reflexdd: You are right. I will put it back

html/views/overlay/index.js
reflexdd: I replaced that now

html/views/common.js
reflexdd: True, that was outdated as it resultetd from a previous version, wich I had.

html/views/common.js
reflexdd: But it would be nice to not have the "Lead" name on the overlay.

reflexdd: I'm not that expert in JS. Would there be a posibility to check if the function is called from the overlay wihtout duplicating the whole function and call that one out of the overlay?

brian-brazil: This isn't a JS thing, it's CRG specific. This particular function is also kinda hairy, and has been the source of a string of bugs.

So I'd like to keep the logic together, so we don't inadvertently break the overlay when changing the scoreboard display.

I'd suggest a small wrapper function to pass in a bool. Though at this stage we probably should be looking at using WS.Register rather than continuing to abuse sbModify.

html/views/common.js
```diff
@@ -20,6 +20,7 @@ function jammer(k, v) {
 	};
 	$(".Team" + id + " .Lead").toggleClass("HasLead", (leadJammer && !starPass));
 	$(".Team" + id).toggleClass("HasJammerName", (jn != ""));
+	$(".Team" + id).toggleClass("starPass", (starPass != false));
```
brian-brazil: Did you forget to push?

brian-brazil:
From a quick test this looks good and nothing seems to have been broken. The lead when there's no jammer name just needs resolving.

html/views/common.js
reflexdd: No, I changed
`$(".Team" + id).toggleClass("IsStarPass ", (starPass != false));` to
`$(".Team" + id).toggleClass("starPass", (starPass != false));`
and removed the redundant class IstStarPass in the css of the html.

brian-brazil: You can still simplify to `starPass`

reflexdd: Ok, got your point.

html/views/common.js
reflexdd: I moved this to the overlay index.js and used WS.Register instead of sbModify to display Lead,Starpass and Skaternames

html/views/common.js
```diff
@@ -20,6 +20,7 @@ function jammer(k, v) {
 	};
 	$(".Team" + id + " .Lead").toggleClass("HasLead", (leadJammer && !starPass));
 	$(".Team" + id).toggleClass("HasJammerName", (jn != ""));
+	$(".Team" + id).toggleClass("starPass", starPass);
```
brian-brazil: HasStarPass for consistency

html/views/overlay/index.css
```diff
@@ -45,20 +45,22 @@ div 		{ white-space: nowrap; display: block; padding: 0; margin: 0;  overflow: h
...
-.Team .Name 		{ width: 64%; padding-left: 1%;
-                  border-left: 2px solid #666;  }
-.Team .Score		{ width: 18%; text-align: right; padding-right: .1em ; font-size: 90%; }
-.Team .Mini		{ width: 8%; font-size: 45%; line-height: 99%; padding-top: 2px; }
-
-.Team .Mini .LeadJammer,
+.Teambox .Indicator		{ width: 4%; height: 100%; font-size: 45%;  float: left; text-align: center;}
+.Teambox .Indicator .Lead		{ opacity: 0; height: 100%; }
+.Teambox .Indicator .HasLead	{ opacity: 1; }
+.TeamBox .starPass .Lead{ display: none; }
```
brian-brazil: Isn't starPass also under Indicator?

Whitespace is also off here

html/views/overlay/index.css
```diff
@@ -45,20 +45,22 @@ div 		{ white-space: nowrap; display: block; padding: 0; margin: 0;  overflow: h
...
+.Team .Name { width: 67%; padding-left: 2%;  border-left: 2px solid #666;  }
+.Team .Score { width: 18%; text-align: right; padding-right: .1em ; font-size: 90%; }
+.Team .Mini	{ width: 8%; font-size: 45%; line-height: 99%; padding-top: 12px; }
 .Team .Mini .JamScore	{ width: 100%; text-align: center; }
 
 .Team .Lead		{ opacity: .2; }
 .Team .HasLead		{ opacity: 1; color: red; }
 
-.Team .DotTimeouts	{ width: 3%; }
+.Team .DotTimeouts	{ width: 3%; padding-top: 2px}
```
brian-brazil: semicolon

html/views/overlay/index.js
```diff
@@ -25,6 +25,46 @@ function initialize() {
 			'ScoreBoard.Period(*).Jam(*).TeamJam(*).Lead',
 			'ScoreBoard.Period(*).Jam(*).TeamJam(*).Lost'], function(k,v) { jamData(k,v); } );
 
+	WS.Register( ['ScoreBoard.Team(*).DisplayLead'], function(k,v) {
```
brian-brazil: You need to register StarPass too, otherwise you can miss updates to starpass that don't affect DisplayLead

In general you must register everything you access.


This is also effectively re-writing jammerName, so that it's now implemented in two places which I'd like to avoid. I'd suggest hacking around what's there for now, and refactoring/reworking all that later.

reflexdd: That's my impression too. it's getting messy having almost the same funcionality in two places.

html/views/overlay/index.css
reflexdd: > Isn't starPass also under Indicator?

It's under Indicator, but the  .starpass Class is added when a Starpass is happening and will then hide the Lead div.

reflexdd: > Whitespace is also off here

Wich whitespace do you mean here? The css property or the whitespace in the whole stylsheet?

brian-brazil: > It's under Indicator, but the .starpass Class is added when a Starpass is happening and will then hide the Lead div.

My point is let's keep this consistent, and mention Indicator for both.

> Wich whitespace do you mean here?

`.Lead{`.

reflexdd: Ok, i think i see what you mean. I will change it to 
`.TeamBox .Indicator .HasStarPass  { display: none; }`

html/views/overlay/index.js
```diff
@@ -191,7 +151,26 @@ function initialize() {
 			setTimeout(function() { $('body').removeClass('preload'); }, 1000);
 }
 
-
+function jammer_ov(k, v) {
```
brian-brazil, line 154: This is still duplicating, try to share as much code as you sanely can with the existing function.

I very much look forward to cleaning all of this up, but that's a bigger effort.

reflexdd: I tryed now to use the orginal jammer function by calling it from the jammer_ov function and handing over a bool. That seems to work without breaking the original function.

html/views/common.js
```diff
@@ -9,7 +9,7 @@ function jammer(k, v) {
 	var inJam = isTrue(WS.state["ScoreBoard.InJam"]);
 
 	if (jammerName == null || jammerName == "") {
-		jammerName = leadJammer ? "Lead" : "";
+		if(!ov)	jammerName = leadJammer ? "Lead" : "";
```
brian-brazil: Please don't use if without braces, it's fragile.

html/views/overlay/index.js
```diff
@@ -165,7 +151,10 @@ function initialize() {
...
+	var jn=	jammer(k,v,true);
```
brian-brazil: `return jammer(k,v,true);`

brian-brazil:
Looking good, just some small style stuff.

html/views/common.js
frank-weinberg: I think `jammerName = leadJammer && !ov ? "Lead" : ""; `  would be better.

brian-brazil: If you're doing that, put parens around the condition for clarity.

brian-brazil: Did you forget to push? I'm not seeing the change Speedy requested.

reflexdd: I missed the comment. Will push it now.

brian-brazil:
Thanks!

PR #442, opened 2020-02-12
Fix SK sheet for SP with points on initial

### LINKED ISSUE #440: Star Pass + Initial Trip Points
When trying to record initial trip points (say, during an overtime jam 😄 ) after a star pass, the initial points seem to bounce around between the original jammer and the new jammer.

Replication Steps:
1. Start an overtime jam (or for now, a normal jam), with jammers and pivots assigned
2. Click the "Star Pass" button
3. Click the "Score +1" button
a. Points will be recorded on the original jammer
4. Click the "Trip +1" button (or wait for the trip to advance automatically)
a. Points will now be recorded on the new jammer
5. Click the "Trip -1" button
a. Points bounce back to the original jammer

It appears impossible to record a the following scenarios
1. Original Jammer scores points on Trip 1, New Jammer scores points on trips 2 through n
2. New Jammer scores points on Trip 1 but never makes it to trip 2.

Low Priority, but the bouncing between jammers suggests that there is something subtle going on with the trip assignments.

frank-weinberg:
The case of points in initial and SP in trip 2 is not explicitly covered
by the Statsbook manual. Closest to the explicit cases would have been to
display '4+' but that looks as if there is a number missing. Thus I went with
'4+SP' instead.

Closes #440

### DIFF
--- a/html/components/sk-sheet.js
+++ b/html/components/sk-sheet.js
@@ -108,8 +108,8 @@ function prepareSkSheetTable(element, teamId, mode) {
 				}
 				break;
 
-			case 'Calloff': case 'Injury': case 'NoInitial': case 'StarPass':
-			case 'ScoringTrip(1).AfterSP': case 'Fielding(Pivot).SkaterNumber':
+			case 'Calloff': case 'Injury': case 'StarPass':
+			case 'Fielding(Pivot).SkaterNumber':
 				var row = jamRow;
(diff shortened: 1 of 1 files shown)


brian-brazil:
SP is a tad wider than NI so can end up wrapping on narrower screens, but it's not as if this is going to be common.

PR #439, opened 2020-02-10
When points are added, advance to the first scoring trip.

AdamSmasherDerby:
Submitted based on the overwhelming support for this modification on the user group. (I count several dozen responses in favor, and only three including Bhrian and Speedy opposed) When points are entered for a skater on an initial trip, a scoring trip is added and the points are transferred to that scoring trip.

Does not break manual entry of initial trip points if desired, and does not apply during overtime.

Resolves issue #428.

PR #443, opened 2020-02-16
Make it easier to understand&handle there being points on initial.

brian-brazil:
Viewing trip 2 now also includes the initial's score, and a button
to insert before trip 1. This is as it's not obvious that the NI column
is the initial trip button.

Screenshot_2020-02-16_09-29-47

### DIFF
--- a/html/components/sk-sheet.css
+++ b/html/components/sk-sheet.css
@@ -42,5 +42,6 @@ body:not(.AllowSelect) .SK {
 #TripEditor .buttons td { text-align: center; }
 #TripEditor .ui-button { font-size: 75%; }
 #TripEditor .ui-button.checked { background: #3f3; }
+#TripEditor .Hide { display: none; }
 
 #skaterSelector .Hide { display: none; }
--- a/html/components/sk-sheet.js
+++ b/html/components/sk-sheet.js
@@ -264,11 +264,16 @@ var tripEditor;
 
 function setupTripEditor(p, j, teamId, t) {
 	var prefix = 'ScoreBoard.Period('+p+').Jam('+j+').TeamJam('+teamId+').ScoringTrip('+t+').';
+	var initialPrefix = 'ScoreBoard.Period('+p+').Jam('+j+').TeamJam('+teamId+').ScoringTrip(1).';
 
 	tripEditor.dialog('option', 'title', 'Period ' + p + ' Jam ' + j + ' Trip ' + (t==1?'Initial':t));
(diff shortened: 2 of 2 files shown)


riffreff:
I think, that is a really good short-term solution for the current issue. It makes the "4+4-issue" way more comprehensible and resolvable for users. 
I would like to see this implemented soon. And then over the next months we will see, if users still can't learn to use the Trip +1 button. And then consider if it still needs a re-visit of the concept.

AdamSmasherDerby:
I support this change, but only IN ADDITION to automatically assigning all points to a scoring trip.  That way, in the overtime and other ruleset cases where a user would want to assign points to the first trip, it is obvious how to do so.

However, I can not support this as an alternative to getting #439 implemented, because points on an initial trip are wrong.  We should not be making wrong data.

brian-brazil:
> That way, in the overtime and other ruleset cases where a user would want to assign points to the first trip, it is obvious how to do so.

I don't think this is what you meant to say.

> And then over the next months we will see, if users still can't learn to use the Trip +1 button. And then consider if it still needs a re-visit of the concept.

I think a wait and see approach is a good idea, this is all pretty new still. Changing the behaviour also risks confusion this early on.

In the longer term if this ends up becoming rare we could remove/trim this to simplify things, and rely on the trip navigation (i.e. buttons for view prev/next trip) that'll be also be present by then. Plus the knowledge that'll have accumulated within the userbase.

frank-weinberg:
We should definitely have an explanatory text and quick fix (resp. undo) button. Either standalone as suggested here or as a part of #439. 
But overall I still favor #439 and as to me it seems that that PR will be ready within the next week I dont think it's worth it to add this now just to replace it after a few days.

brian-brazil:
> to me it seems that that PR will be ready within the next week 

That doesn't seem likely to me, even presuming that we agree it should be merged. It's already been a week and the more complex interactions have yet to be touched on.

brian-brazil:
After some thought I've tweaked the wording to indicate what you could have done to not get into this situation. Keeping it succinct, correct, understandable and non-accusatory is surprisingly tricky.

AdamSmasherDerby:
Nice job with the text - definitely agree that "concise" and "understandable" are mutually opposing goals here. Recommend merge.

brian-brazil:
Yeah, the trick was finding a way not to have to mention overtime jams or non-lead jammers.

frank-weinberg:
Superseded by #439

PR #445, opened 2020-02-16
Add FO when penalty limit is reached.

frank-weinberg:
Also make FO jam number follow corresponding penalty


frank-weinberg:
I realized that it would make sense to add removing the FO when a penalty is removed. Gonna add that tonight. So please don't merge yet.

brian-brazil:
One thing to think about aswell is when a penalty is added, but in the middle rather than being the last penalty. Not sure what the right thing to do there is, probably whatever is simplest in the code.

frank-weinberg:
It should already handle this case, but I I think I'll add it to the unit test as it's not obvious that it does.

frank-weinberg:
All done. And fixed one of those Heisenbugs on the side.

brian-brazil:
Looks okay, can you add a test for expulsions not being affected? Do the new tests already cover the bug you discovered?

How did this present as a Heisenbug? It looks like a typical NPE to me.

frank-weinberg:
> Do the new tests already cover the bug you discovered?

Yes. That's how I found it.

> How did this present as a Heisenbug?

When the next link in a deleted penalty was missing, the following penalties would not be renumbered, leaving a gap in the list of penalties.

brian-brazil:
> When the next link in a deleted penalty was missing, the following penalties would not be renumbered, leaving a gap in the list of penalties.

That's not a Heisenbug, a Heisenbug is a bug that disappears/changes as you try to debug it.

PR #446, opened 2020-02-17
Fix rounding of countdown clocks on scoreboard view and overlay

### LINKED ISSUE #444: Period clock changes missing from scoreboard display
I ran current HEAD at a tournament this weekend and about half the time when the period clock was adjusted during a timeout the last change did appear on the operator screen but not on the main scoreboard. (Both on the same machine as the backend.)

I intend to look into this myself in the coming days but wanted to flag this for other dev users and as a blocker for the 4.0.2 release.

frank-weinberg:
This change should probably be done somewhere in WS.js, but I couldn't figure out where and wanted to fix the issue at hand.

closes #444

### DIFF
--- a/html/views/common.js
+++ b/html/views/common.js
@@ -140,6 +140,7 @@ function toClockInitialNumber(k, v) {
 }
 
 function toTime(k, v) {
+	k = WS._enrichProp(k);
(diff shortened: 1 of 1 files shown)


brian-brazil:
It'll all get cleaned up when I move it to the backend anyway

PR #447, opened 2020-02-18
Don't allow initial trip to be deleted if its the only trip.

brian-brazil:
When an attempt is made to delete the initial trip, set its score
to 0 in this case instead. This is so deleting a trip always
removes its score.


brian-brazil:
This both fixes an NPE if you delete the initial trip, as well as providing consistent Trip -1 semantics.

That unlink fixes Numbered numbers but remove doesn't is weird. Those should probably be unified at some point.

frank-weinberg:
The Trip holds more data than the score, so removing it and adding a new one would be cleaner. And this should probably done from `unlink()` instead of all calling sites. (Just check that `neighborsRemoved` is false so it doesn't trigger when the whole Jam is deleted.)

brian-brazil:
The other information is the star pass, and that doesn't un-happen due to the trip count being incorrect.

frank-weinberg:
There is also the timing data, we have talked about adding annotations, and who knows what the future brings. With this PR as it is we would have to remember to update two unrelated pleces every time we add a new value to scoring trips. At least you should add a function to ScoringTrip that clears all data that should be cleared on a blocked deletion attempt.

brian-brazil:
I'm not a fan of the duplication either, my first attempt included removing it but got very yak-shavy. Let me see what I can do.

brian-brazil:
That should be a bit better now. Thinking on it the only other thing to clear would be the end time, so I've done that.

frank-weinberg:
Annotations are now also a thing to clear.

PR #450, opened 2020-02-18
Improve handling of trip advancement timer

frank-weinberg:
- Cancel the timer when the trip score is set back to 0.
- When a timer is running Trip -1 cancels the timer without removing a trip.


brian-brazil:
> Cancel the timer when the trip score is set back to 0.

Hmm, can't this only be done from the SK sheet? What's your thinking here?

> When a timer is running Trip -1 cancels the timer without removing a trip.

I'm not sure this makes sense. That advanced trip doesn't exist yet, so I'd expect it to work as it does now with the cancellation and removal of the trip.

This needs unittests. Also the frontend indicator is completely de-coupled from the backend with logic that's already a bit different, so that'd need to be fixed too.

frank-weinberg:
> Hmm, can't this only be done from the SK sheet?

There's a 'Score -1' button. I would assume that this is the most commonly used way to undo an erroneous score entry.

> I'm not sure this makes sense.

1. An experienced SBO might realize that they entered points before Jam end but the Jammer never actually left the EZ. Currently they have to wait for the auto-advance to happen before pressing Trip -1 to fix the paperwork. With this change they can press it right away.

2. We recommend pressing Trip -1 to undo the trip advancement when a Jammer gets reabsorbed after getting lead. An SBO might do the same when the Jammer gets reabsorbed after a completed scoring trip. (I wouldn't recommend doing this but I can see it happening.) As the Jammer cannot change the trip score and it will not be signalled again, this change will help prevent a wrong score in this scenario.

On the other hand it is rather unlikely that someone wants to remove a scoring trip that they just entered points for. (This would require the points entry and the previous starting of the trip to be in error. The only case where this is somewhat likely is when the trip was only started as a reaction to entering the points.)

> This needs unittests.

Do you have a suggestion how to do them without adding 4s per test?

> Also the frontend indicator is completely de-coupled from the backend 

What indicator? We do not indicate that a trip advancement is pending.

brian-brazil:
> There's a 'Score -1' button.

Ah yes.

> An experienced SBO 

Okay, those make sense.

> Do you have a suggestion how to do them without adding 4s per test?

There's ways of doing it, but here we can cheat a bit. We don't actually need it do run, so we can call `cancel` and check the result.

> What indicator? We do not indicate that a trip advancement is pending.

The trip score shows up to the inside of the total score in green fading to black acts as one currently (including changing to 0 for about a second when the trip advanced, and then disappearing - which is confusing).

If we're doing magic on the backend we should make this consistent and clearer.

frank-weinberg:
I added unit tests. For the trip score indicator I agree that it can be improved but I don't have a clear vision how it should behave and I think that it is beyond the scope of this PR.

src/com/carolinarollergirls/scoreboard/core/Team.java
```diff
@@ -51,6 +51,9 @@
     public int getScore();
 
     public ScoringTrip getCurrentTrip();
+    
+    //returns true if an advancement was pending
+    public boolean cancelTripAdvancement();
```
brian-brazil: This is an internal API for testing, no need to make it part of the public interface. The tests can already access it if they downcast to a TeamImpl.

frank-weinberg: I'd say a case could be made for having this as part of the public API. But that's not a hill I'm even getting wounded on. Removed.

Issue #448, opened 2020-02-18
Entering Abitrary Scoring Trips on SK Sheet


official-sounding:
In the current scoreboard, you can click on any scoring trip and enter a score in that trip, even if there are not scores in previous trips.  Whereas on the ePT sheet, if I click in an empty box in the row for a given skater, I will enter the "first available" penalty for that skater, rather than in the box that I actually clicked on.   

As a visual aid, I am talking about clicking somewhere in the highlighted section
![image](https://user-images.githubusercontent.com/1295669/74754460-3f7c2180-5237-11ea-90f3-fcff6db2925d.png)

Fortunately, the underlying data model + front-end code seems to do the right thing when I enter points in an arbitrary trip.  Currently, CRG-to-Statsbook doesn't pick up the trips, however.

I think that the current behavior is something of a foot-gun/overly surprising, especially given the mismatch with how the ePT page operates.  Additionally, There isn't necessarily a use case for the current behavior, so I think it would be worth discussing alternative behaviors:

1. Open on the next empty scoring trip (in the case of the above screenshot, open on "Trip 3").  Essentially align the behaviors between ePT and eSK
2. Open on the last entered scoring trip (in the case of the above screenshot, open on "Trip 2").
3. A somewhat hybrid approach, open on the current trip if the jam is active and the score is 0 (eg if there is only a "." in the box), otherwise open on the next trip.
4. Don't open a pop-up if it's not a current scoring trip or the next scoring trip.
5. Something else entirely.

brian-brazil:
1 would make the most sense to me. I think that's safe even if the backend has advanced the trip but the frontend doesn't know about it just yet.

frank-weinberg:
I'd be fine with 1 as well.

Issue #441, opened 2020-02-12
Cannot edit points on trips 11+


official-sounding:
(I am guessing this is a known issue, but want to log it since I was playing at the margins to generate test cases for Smasher's tool)

When a jammer has 10 or more trips, the scoreboard does not allow for editing of points recorded on trips 11 or beyond.  If I catch the issue during lineups, I can fix the issues using the trip/score buttons, but in subsequent jams trips and points beyond 11 are "locked in".

It's obviously an incredibly niche issue that will (never?) show up in normal game play, but the inability to edit the score on trips 11+ can create a situation where the score is irreparably wrong -- A scoreboard operator would have to remove points from a previous scoring trip to make the scoreboard display correctly.

frank-weinberg:
I remember @brian-brazil recently mentioning the idea of adding "Previous Trip" and "Next Trip" buttons to the edit box. That would cover this issue.

brian-brazil:
Yeah. It should be rare enough that that's sufficient.

PR #451, opened 2020-02-19
Support annotations on scoring trips

frank-weinberg:
and update the UI of the trip editor
Screenshot_20200219_083323
Screenshot_20200219_083610

(The example annotation is not part of this PR.)

### DIFF
--- a/html/components/sk-sheet.css
+++ b/html/components/sk-sheet.css
@@ -42,9 +42,8 @@ body:not(.AllowSelect) .SK {
 #TripEditor input[type="number"] { width: 50px; }
 #TripEditor table { width: 100%; }
 #TripEditor td { padding: 0.25em; text-align: center; width: 33%; }
+#TripEditor td.close { text-align: right; }
 #TripEditor .header { font-weight: bold; }
-#TripEditor .Annotation.read>td { text-align: left; }
-#TripEditor .close>td { text-align: right; }
 #TripEditor .ui-button { font-size: 75%; }
 #TripEditor .ui-button.checked { background: #3f3; }
 #TripEditor .Hide { display: none; }
--- a/html/components/sk-sheet.css
+++ b/html/components/sk-sheet.css
@@ -41,16 +41,10 @@ body:not(.AllowSelect) .SK {
 
 #TripEditor input[type="number"] { width: 50px; }
 #TripEditor table { width: 100%; }
-#TripEditor td { padding: 0.25em; text-align: center; width: 33%; background: #ccc; }
+#TripEditor td { padding: 0.25em; text-align: center; width: 33%; }
 #TripEditor .header { font-weight: bold; }
-#TripEditor .Annotation.read>td { text-align: left; padding: 0.5em; }
-#TripEditor .Annotation.head>td { border-top: 2px solid white; }
-#TripEditor .Annotation.head>td:first-of-type { border-radius: 12px 0px 0px 0px; }
-#TripEditor .Annotation.head>td:last-of-type { border-radius: 0px 12px 0px 0px; }
-#TripEditor .Annotation.edit>td { border-radius: 0px 0px 12px 12px; border-bottom: 2px solid white; }
-#TripEditor .Annotation.read>td { border-radius: 0px 0px 12px 12px; border-bottom: 2px solid white; }
-#TripEditor tr:not(.Annotation)>td:first-of-type { border-radius: 12px 0px 0px 12px; }
-#TripEditor tr:not(.Annotation)>td:last-of-type { border-radius: 0px 12px 12px 0px; }
+#TripEditor .Annotation.read>td { text-align: left; }
+#TripEditor .close>td { text-align: right; }
 #TripEditor .ui-button { font-size: 75%; }
 #TripEditor .ui-button.checked { background: #3f3; }
 #TripEditor .Hide { display: none; }
--- a/html/components/sk-sheet.css
+++ b/html/components/sk-sheet.css
@@ -36,11 +36,23 @@ body:not(.AllowSelect) .SK {
 .SK.Period .Trip, .Period .JamTotal, .Period .GameTotal { width: 9%; }
 .SK.Period .GameTotal, .SK.Period .JamTotal,.SK.Period .Lost, .SK.Period .Trip2 { border-left-width: 2px; }
 .SK.Period .GameTotal { background-color: #d3ecb6; }
+.SK.Period .hasAnnotation { position: relative; }
+.SK.Period .hasAnnotation::after { position: absolute; right: 2px; top: 2px; height: 5px; width: 5px; background-color: red; content: ""; }
 
 #TripEditor input[type="number"] { width: 50px; }
-#TripEditor td { padding-bottom: 0.5em; }
-#TripEditor .buttons td { text-align: center; }
+#TripEditor table { width: 100%; }
+#TripEditor td { padding: 0.25em; text-align: center; width: 33%; background: #ccc; }
+#TripEditor .header { font-weight: bold; }
+#TripEditor .Annotation.read>td { text-align: left; padding: 0.5em; }
+#TripEditor .Annotation.head>td { border-top: 2px solid white; }
+#TripEditor .Annotation.head>td:first-of-type { border-radius: 12px 0px 0px 0px; }
+#TripEditor .Annotation.head>td:last-of-type { border-radius: 0px 12px 0px 0px; }
+#TripEditor .Annotation.edit>td { border-radius: 0px 0px 12px 12px; border-bottom: 2px solid white; }
+#TripEditor .Annotation.read>td { border-radius: 0px 0px 12px 12px; border-bottom: 2px solid white; }
+#TripEditor tr:not(.Annotation)>td:first-of-type { border-radius: 12px 0px 0px 12px; }
+#TripEditor tr:not(.Annotation)>td:last-of-type { border-radius: 0px 12px 12px 0px; }
 #TripEditor .ui-button { font-size: 75%; }
 #TripEditor .ui-button.checked { background: #3f3; }
+#TripEditor .Hide { display: none; }
 
 #skaterSelector .Hide { display: none; }
--- a/html/components/sk-sheet.css
+++ b/html/components/sk-sheet.css
@@ -47,5 +47,6 @@ body:not(.AllowSelect) .SK {
 #TripEditor .ui-button { font-size: 75%; }
 #TripEditor .ui-button.checked { background: #3f3; }
 #TripEditor .Hide { display: none; }
+#TripEditor .Invisible { visibility: hidden; }
 
 #skaterSelector .Hide { display: none; }
--- a/html/components/sk-sheet.js
+++ b/html/components/sk-sheet.js
@@ -282,17 +282,23 @@ function prepareSkSheetTable(element, teamId, mode) {
 var tripEditor;
 
 function setupTripEditor(p, j, teamId, t) {
+	while (t > 1 && WS.state['ScoreBoard.Period('+p+').Jam('+j+').TeamJam('+teamId+').ScoringTrip('+(t-1)+').Score'] === undefined) { t--; }
+	if (t < 1) { t = 1; }
+	
 	var prefix = 'ScoreBoard.Period('+p+').Jam('+j+').TeamJam('+teamId+').ScoringTrip('+t+').';
 
-	tripEditor.dialog('option', 'title', 'Period ' + p + ' Jam ' + j + ' Trip ' + (t==1?'Initial':t));
+	tripEditor.dialog('option', 'title', 'Period ' + p + ' Jam ' + j + ' Trip ' + (t===1?'Initial':t));
 	tripEditor.find('#score').val(WS.state[prefix+'Score']);
 	tripEditor.find('#afterSP').toggleClass('checked', isTrue(WS.state[prefix+'AfterSP']));
 	var annotation = WS.state[prefix+'Annotation'] || '';
-	tripEditor.find('textarea#annotation').val(annotation);
-	tripEditor.find('span#annotation').text(annotation);
-	tripEditor.find('.edit.Annotation').addClass('Hide');
-	tripEditor.find('.read.Annotation').removeClass('Hide');
+	tripEditor.find('#annotation').val(annotation);
+	tripEditor.find('#prev').toggleClass('Hide', t === 1);
+	tripEditor.find('#next').toggleClass('Hide', WS.state[prefix+'Score'] === undefined);
 	tripEditor.data('prefix', prefix);
+	tripEditor.data('team', teamId);
+	tripEditor.data('period', p);
+	tripEditor.data('jam', j);
+	tripEditor.data('trip', t);
 	tripEditor.dialog('open');
 }
 
@@ -329,36 +335,39 @@ function prepareTripEditor() {
 									WS.Set(tripEditor.data('prefix')+'AfterSP', check);
 								}))))
 				.append($('<tr class="buttons">')
-						.append($('<td>').append($('<button>').attr('id','remove').text('Remove').button().click(function() {
+						.append($('<td>').append($('<button>').attr('id','remove').text('Remove Trip').button().click(function() {
 							WS.Set(tripEditor.data('prefix')+'Remove', true);
 							tripEditor.dialog('close');
 						})))
-						.append($('<td>').append($('<button>').attr('id','insert_before').text('Insert Before').button().click(function() {
+						.append($('<td>').append($('<button>').attr('id','insert_before').text('Insert Trip').button().click(function() {
 							WS.Set(tripEditor.data('prefix')+'InsertBefore', true);
 							tripEditor.dialog('close');
+							setupTripEditor(tripEditor.data('period'), tripEditor.data('jam'), tripEditor.data('team'), tripEditor.data('trip'));
+							tripEditor.find('#score').val(0); // the update of the popup may run before the WS is updated
 						}))))
-				.append($('<tr>').append($('<td>').attr('colspan', '3').append($('<hr>'))))
-				.append($('<tr>').addClass('head Annotation')
+				.append($('<tr>').append($('<td>').attr('colspan', '2').append($('<hr>'))))
+				.append($('<tr>').addClass('Annotation')
 						.append($('<td>').addClass('header').text('Notes: '))
-						.append($('<td>').append($('<button>').addClass('read Annotation').text('Edit').button().click(function() {
-									tripEditor.find('.edit.Annotation').removeClass('Hide');
-									tripEditor.find('.read.Annotation').addClass('Hide');
-									tripEditor.find('textarea#annotation').focus();
-								}))))
-				.append($('<tr>').addClass('read Annotation')
-						.append($('<td>').attr('colspan', '3').append($('<span>').attr('id', 'annotation'))))
-				.append($('<tr>').addClass('edit Annotation')
-						.append($('<td>').attr('colspan', '3')
+						.append($('<td>').append($('<button>').text('Clear Notes').button().click(function() {
+							tripEditor.find('#annotation').val('');
+						}))))
+				.append($('<tr>').addClass('Annotation')
+						.append($('<td>').attr('colspan', '2')
 								.append($('<textarea>').attr('cols', '25').attr('rows', '4').attr('id', 'annotation').change(function() {
 									WS.Set(tripEditor.data('prefix')+'Annotation', $(this).val());
-									tripEditor.find('span#annotation').text($(this).val());
-									tripEditor.find('#clearAnn').toggleClass('Hide', $(this).val() === '');
-								}).focusout(function(){
-									tripEditor.find('.edit.Annotation').addClass('Hide');
-									tripEditor.find('.read.Annotation').removeClass('Hide');
 								}))))
-				.append($('<tr class="buttons close">')
-						.append($('<td colspan="2">').append($('<hr>')).append($('<button>').attr('id','submit').text('Close').button().click(function() {
+				.append($('<tr>').append($('<td>').attr('colspan', '2').append($('<hr>'))))
+				.append($('<tr class="buttons nav">')
+						.append($('<td>')
+								.append($('<button>').text('⬅ Prev').attr('id', 'prev').button().click(function() {
+									tripEditor.dialog('close');
+									setupTripEditor(tripEditor.data('period'), tripEditor.data('jam'), tripEditor.data('team'), tripEditor.data('trip') - 1);
+								}))
+								.append($('<button>').text('Next ➡').attr('id', 'next').button().click(function() {
+									tripEditor.dialog('close');
+									setupTripEditor(tripEditor.data('period'), tripEditor.data('jam'), tripEditor.data('team'), tripEditor.data('trip') + 1);
+								})))
+						.append($('<td>').addClass('close').append($('<button>').attr('id','close').text('Close').button().click(function() {
 							tripEditor.dialog('close');
 						}))))
 		);
--- a/html/components/sk-sheet.js
+++ b/html/components/sk-sheet.js
@@ -36,7 +36,8 @@ function prepareSkSheetTable(element, teamId, mode) {
 				'ScoreBoard.Period(*).Jam(*).TeamJam(' + teamId + ').Fielding(Pivot).SkaterNumber',
 				'ScoreBoard.Period(*).Jam(*).TeamJam(' + teamId + ').ScoringTrip(*).AfterSP',
 				'ScoreBoard.Period(*).Jam(*).TeamJam(' + teamId + ').ScoringTrip(*).Current',
-				'ScoreBoard.Period(*).Jam(*).TeamJam(' + teamId + ').ScoringTrip(*).Score'
+				'ScoreBoard.Period(*).Jam(*).TeamJam(' + teamId + ').ScoringTrip(*).Score',
+				'ScoreBoard.Period(*).Jam(*).TeamJam(' + teamId + ').ScoringTrip(*).Annotation'
 		], handleUpdate);
 	}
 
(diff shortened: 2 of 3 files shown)


brian-brazil:
I don't think we should have annotations on scoring trips, that's far too granular for something that's pretty rare. I think they should be per-jam.

brian-brazil:
Also this wouldn't be resilient to a trip being deleted.

frank-weinberg:
> I think they should be per-jam.

- There is a popup for editing per-trip data but none for editing per-jam data yet. So doing it per Jam would add more complexity.
- With per-trip annotations you can associate an annotaion with a jam by adding it to any trip of that jam. With per-jam annotations there is no way to associate an annotation with a specific trip.
- On the lineup side we are doing annotations per fielding which is the same level of granularity. So doing it per trip is more consistent.

So overall we get more options and a more consistent UX with less complexity.

(It also takes less extra complexity to implement #443 based on per trip annotations vs. per-jam annotations.)

> Also this wouldn't be resilient to a trip being deleted.

Why should it? That's like complaining that removing a trip removes the points from the trip.

brian-brazil:
> With per-trip annotations you can associate an annotaion with a jam by adding it to any trip of that jam. With per-jam annotations there is no way to associate an annotation with a specific trip.

In what situation might you have enough per-trip annotations that that would matter? The only ones I've seen relate to box trips/penalties, and the scoreboard already tracks that. Plus it'd be be risky to regularly open dialogs during a jam.

I'm expecting that this is something that'd be used less than once a game by the operator, and they can always say Trip 2 or whatever in the annotation. I'm expecting notes like jam stopped due to a child running on track or a non-lead jammer called it off, and other such very rare events.

In terms of the Trip+1 discussion, per-jam also makes more sense. I don't think suggesting the user  delete the is trip a good idea - we could inadvertently lose timing data if they meant to clear but actually hit remove because that's what the annotation says. Plus it would move around, plus it'd look weird to have an indicator on the NI column.

I think that either a per-jam or per-team+jam annotation is the way to go here.

> On the lineup side we are doing annotations per fielding which is the same level of granularity. So doing it per trip is more consistent.

We've sturctured-ish data there though of the no penalty/penalty over turned which relate to a specific skater. We don't have an equivalent here.

> (It also takes less extra complexity to implement #443 based on per trip annotations vs. per-jam annotations.)

I don't think it makes a difference code wise, and actually gets messier in user terms.

> Why should it? That's like complaining that removing a trip removes the points from the trip.

That may be logical, but I don't think it'd be the expected behaviour. If I "delete" a jam on paper by shifting the scores, any ad-hoc annotations don't go away.

frank-weinberg:
> In what situation might you have enough per-trip annotations that that would matter?

Jammer goes through the pack without helmet cover. Adding a "Helmet cover off" annotation to the 0 makes perfect sense. (And yes, one is enough for per-trip to be better in the scenario. even when mentioning the trip in the per-jam annotation we lose the direct association of "there is an annotation on this 0")

Adding the boxes part of dots and boxes would also be an option with per-trip annotations.

> [jam-end scenarios]

Putting these annotations on the last scoring trip doesn't feel significantly less natural to me than placing them on the jam number (or wherever you would place a per-jam annotation in the UI). I don't think that's enough to justify the added complexity of another poup. Esecially not given that there are other scenarios that would be worse off with per-jam.

> we could inadvertently lose timing data

If we have the trip +1 annotation on an initial trip the timing data on that trip is already worthless.

> Plus it would move around

In the above scenario the annotation moving with the 0 is the desired behaviour. If it's about a jam end scenario annotated to the last trip it moving with the last trip also comes naturally.
Plus: Annotations are presumably rare, deleting trips also is. Having both on the same jam should thus be so rare that this becomes a non-issue.

> plus it'd look weird to have an indicator on the NI column.

On the other hand that could help getting users to pick up on the fact that clicking there allows them to edit the initial trip.

>>    (It also takes less extra complexity to implement #443 based on per trip annotations vs. per-jam annotations.)

> I don't think it makes a difference code wise, and actually gets messier in user terms.

For starters the per-trip annotations allow us to refer to the existing remove or insert button in the same popup. With per-jam you'd have to either add a special case button for this scenario or describe a multi click sequence that starts with hiding the annotation. The former is extra code complexity for a special case and the latter is a way worse UX.

brian-brazil:
> Jammer goes through the pack without helmet cover. 

I've seen that literally once, I don't see the need to primarily design for something that happens less than once in a hundred games. Per-jam annotations would be sufficient for that.

> Adding the boxes part of dots and boxes would also be an option with per-trip annotations.

That's a lot of complexity, and I don't think makes sense given we already have eLT/ePLT to track that. The Box button we have on the operator screen is already concerning me a bit, now that we have proper box trip tracking.

> If we have the trip +1 annotation on an initial trip the timing data on that trip is already worthless.

What about the initial trip they just deleted? We're back to 4+4.

> Putting these annotations on the last scoring trip doesn't feel significantly less natural to me than placing them on the jam number

What if the last trip is a 0, which then gets deleted as they didn't actually complete trip?

Putting per-jam information on trips risks accidental deletion.

> On the other hand that could help getting users to pick up on the fact that clicking there allows them to edit the initial trip.

True, but it's going to be extremely rare outside overtime jams as things are currently proposed. The navigation will also help with that.

> With per-jam you'd have to either add a special case button for this scenario or describe a multi click sequence that starts with hiding the annotation.

The vast majority of the time this comes up hitting clear is the right thing to do, so no need to put a lot of effort into optimising the special case. The only realistic case where you'd need to do the additional two clicks to delete a trip is when you were in overtime jam but forgot to go into overtime mode, which is going to be much less than once per hundred games.

frank-weinberg:
> I've seen that literally once

The per-jam scenarios you listed aren't exactly more common.

> What about the initial trip they just deleted?

That's the one with the broken timing data. And if 4+4 was corrrect, after deleting the wrong initial timing data will be correct.

> What if the last trip is a 0, which then gets deleted

So the JR signals that the Jammer was out of the EZ, then an Annotation is made on that 0 and after that the JR changes their mind on the Jammer being out and bothers to skate to the score table to get a change to the paperwork made? I think that is such an unlikely scenario that we can live with the operator having to copy the annotation over to another trip. We're easily in the same ballpark as an unmarked overtime jame here.

It seems to me that overall for either version you can find occasional circumstances where it is slightly worse than the other one and rare ones where it is notably worse. And the actual frequencies are easily within the (relatively wide) margin of error with which we can measure them (i.e. based on anecdata). So it's a tie in that regard.

Which brings me back to the fact that per-trip annotations can use the existing trip-popup whereas per-jam ones would need a new popup. So per-trip annotations win in terms of code-complexity.

And even if scoring trips and fieldings are different things, having multiple annotations per jam on the LT sheet is more consistent with having multiple annotations per jam here than with having only one here.

brian-brazil:
> So it's a tie in that regard.

Pretty much. The rarity of all these events (even when combined) also indicates that per-jam is sufficient to record them.

> That's the one with the broken timing data. And if 4+4 was corrrect, after deleting the wrong initial timing data will be correct.

The vast majority of the time 4+4 is incorrect, so we should aim for an interface that makes clearing the annotation easy rather than risking the operator incorrectly changing it to 4+4.

This will be the common case, and what I think we should be optimising for. All things considered I think a new dialog is a good tradeoff here, and also could in future also allow us to surface things like how long a jam took in more obvious ways than it is now.

I note also that all of the weird bugs in the Trip+1 PR as it stands come from it being per-scoringtrip rather than per-teamjam.

frank-weinberg:
IF we do a per-jam popup with info like duration, this would also be the place where users expect a button to delete the Jam. Now a click on the wrong button would not just delete an initial trip with bogus timing data and no points but a whole Jam. That's a higher risk than what's proposed here.

And reducing the risk of a mis-click is a matter of phrasing the annotation which is beyond the scope of this PR.

brian-brazil:
> IF we do a per-jam popup with info like duration, this would also be the place where users expect a button to delete the Jam. 

Let's not make that that easy :)

> And reducing the risk of a mis-click is a matter of phrasing the annotation which is beyond the scope of this PR.

Considering the primary use case of this is going to be the trip advance stuff, I think considering it in this PR is appropriate. Not everyone using the scoreboard is going to have perfect English, and during games aren't exactly stress-free scenarios.

frank-weinberg:
> Let's not make that that easy

It should be easy as it is a relatively common scenario that a jam is whistled dead early on and then declared not a jam or that a SBO forgets about undo after erroneusly starting a jam. Without deleting the surplus jam the jam number is off which a lot of people don't like.

> Considering the primary use case of this is going to be the trip advance stuff, I think considering it in this PR is appropriate.

If we are going to discuss this in this PR, we might as well include those changes here as well and close #439 and #443.

brian-brazil:
> Without deleting the surplus jam the jam number is off which a lot of people don't like.

One thought that occurs to me is a confirmation dialog that only pops up if the jam lasted more than N seconds or had a score, but that's getting sidetracked.

> If we are going to discuss this in this PR

We don't have to combine everything into one PR, however we can consider how this design will impact on the goals of the other PRs.

frank-weinberg:
Ok. This PR is suggesting adding annotations to scoring trips. You seem to be saying that per se its complexity to usefulness ratio is acceptable. Of the open PRs it would allow simplifying #443 and it would make it easier to implement changes requested on #439. 

You are however saying that it should not be merged and instead a hypothetical new PR that adds annotations to Jams should be merged. Said PR would have approximately the same usefulness and impact on other open PRs while requiring additional complexity in the form of a new popup.

To get this additional complexity argued away, you then introduce another hypothetical PR that adds this new popup for the purpose of displaying a set of not clearly specified data. Due to the lack of specification the usefulness to complexity ratio of that PR remains vague.

And the only objective justification for prefering the hypothetical PRs that you give is that if #439 is modified based on this PR and the annotation is worded as the sample in the screenshot here it might mislead users into accidentally deleting a scoring trip. This can be solved with a lot less effort by simply wording the annotation differently.

Beyond that it's a matter of taste. Each version is preferred by 50% of maintainers. But one is already implemented while the other isn't. And the person who puts in the work should be the one who gets to break the tie on matters of taste.

brian-brazil:
> Due to the lack of specification the usefulness to complexity ratio of that PR remains vague.

That's a future hypothetical, and a weak argument at best. Something to consider though as to how else we might end up using these. It might also be a place to add OS offsets (if we ever decide to support them), which are per-jam rather than per trip.

Given the rarity of use cases all around, we probably shouldn't end up with both types of annotation.

> This can be solved with a lot less effort by simply wording the annotation differently.

I disagree there. When working with stressed and potentially non-native speakers, the more words there are the more likely they'll misunderstand and take a hasty unintended action just to try to make it go away. With the particular example here it's emphasising the word 'Remove' so I'd expect quite a few users to press the Remove button due to that alone - which isn't what we want. Do you have wording which you think would avoid this?

In addition this proposal is going to be confusing when annotations move around, as in general per-trip annotations will be harder for users to understand than per-jam.

frank-weinberg:
> Do you have wording which you think would avoid this?

Just leave the second sentence out of the screenshot example. As it refers to a quite rare situation we can get away with not having a howto for said situation in the UI.

> annotations move around

Which is one of the rare scenarios we agreed cancel out in the end. I'm not going to reenter that debate.

brian-brazil:
> Just leave the second sentence out of the screenshot example. As it refers to a quite rare situation we can get away with not having a howto for said situation in the UI.

I'm fine with not mentioning how to go back to 4+4 for those reasons. That's not my argument.

We're talking once a game that an annotation would appear in an automated fashion. The user wants it to go away. What will happen when they try to make it go away?

Even with that change to the text, I think Remove would be clicked as often as Clear as this UI is not intuitive.

If the thesis is that any situation which results in 4+4 is completely unacceptable, then we should not be removing it on plausible one operator error it only to make it easy to reoccur due to another plausible operator error. Getting that right would be very hard with per-trip, but we get it for free with per-jam. I can also see users hitting Edit when they want to change points for a trip.


The edit/lock is also unnecessary, this should be a simple textarea like we have on LT.

> Which is one of the rare scenarios we agreed cancel out in the end. I'm not going to reenter that debate.

It's not just about the rare scenarios that'd cause an annotation to be manually entered in the first place. The question is that once an annotation exists, how should it behave? I don't think it moving around is what a user would expect, it seems like being complicated for the sake of being complicated from a design standpoint.

frank-weinberg:
> I think Remove would be clicked as often as Clear

I think your believe is fueled by you being primed with the previous text and a general dislike of per-trip-annotations. The annotation-box is visually separated from the other elements. We could further help by making the labels "Remove Trip" and "Clear Notes" (and probably "Insert Trip Before" for consistency).

> The edit/lock is also unnecessary, this should be a simple textarea like we have on LT.

The sample annotation is too big to be completely visible in the input field. Plus this way it is a lot easier to read.
What I could see is going to the input by clicking on the text (or when it is empty) and back to lock when the focus moves out of the input like we do with team names on the operator interface. This would also move the clear button right next of the "Notes" header strengthening their association. This would be rather obscure but presumably editing an existing annotation is not a common use case anyways.
(And I thought about making the LT annotations behave like these but decided to make that a separate PR, which I will now wait with until we have settled on a design for this one.)

> I don't think it moving around is what a user would expect

I don't think it is any more surprising than the other content of the cell (points) moving around.

brian-brazil:
> I think your believe is fueled by you being primed with the previous text and a general dislike of per-trip-annotations.

No, I'm trying to look at it as fresh user would who has no idea how any of this works.

> The annotation-box is visually separated from the other elements. 

It's not, it's placed right in the middle of them.

> We could further help by making the labels "Remove Trip" and "Clear Notes" (and probably "Insert Trip Before" for consistency).

More words don't necessarily help. The issue is that both Remove and Clear can be taken to mean delete the note - and it's not fully clear visually which buttons relate to the note, which to the trip, and which to the score.

> The sample annotation is too big to be completely visible in the input field. Plus this way it is a lot easier to read.

That's one reason I suggested a textarea rather than just an input.

> This would be rather obscure but presumably editing an existing annotation is not a common use case anyways.

It's not, but that doesn't mean we need to go out of our way to make it hard. However selecting all the text and pressing backspace to clear it is something that I'd expect to be common.

> I don't think it is any more surprising than the other content of the cell (points) moving around.

Probably not, but this will likely be the first time a user encounters it - which they probably wouldn't ever had to really worry about otherwise. Thus it is a concern.


As I see it per-trip annotations are something that might have sane use maybe once a year across all users (how often do two really weird things happen in a single jam, but on different trips? Taking the 1/100 number for incidences in a game, that'd imply no more than one in 10k games (not jams) which is about how many are in a year in total). Given that per-teamjam effectively handles all the same use cases, and is simpler to understand for users, per-trip appears to be overengineering.

frank-weinberg:
> I'm trying to look at it as fresh user would

Me as well. So apparently one of us is failing.

> That's one reason I suggested a textarea rather than just an input.

You wrote "like we have on LT". LT has a single line input.
And text in a textarea is still harder to read.
And having a multiline input all the time will make it >50% of the dialog even when empty. That's too much IMO.

> Probably not, but this will likely be the first time a user encounters it

How so? With annotations being rare, it's way more likely that none of the moved trips will be annotated. Why should that be different for the first insertion/deletion by a new user.

> As I see it per-trip annotations are something that might have sane use maybe once a year

As I see it per-trip annotations are the best place for the auto-completed initial trip annotation. It refers to a specific trip and deleting that trip makes it superfluous meaning it makes sense to delete the annotation with the trip.

> per-trip appears to be overengineering

How is using the approach that's most straightforward to code overengineering?

reflexdd:
I don't know if thats the right place but what I found irritating when adding points was, that I was searching for a submit button. That the close-Button is having this function is'nt quite clear. Can't we label it with submit instead of close?

frank-weinberg:
Things are submitted as they are entered. Naming the close button submit might lead users to assume that closing via the X in the top right cancels their changes.

reflexdd:
> Naming the close button submit might lead users to assume that closing via the X in the top right cancels their changes.

That's what I would expect the buttons to work. Closing via X = Dismiss and submit = submit

frank-weinberg:
That's how it initially worked both here and on the LT paperwork. But that didn't work for box trips as there is some non-local checks involved. Having them change immediately and everything else not was less intuitive than applying all changes immediately.

frank-weinberg:
I have now redesigned this to essentially match the design from #443 but with the close button moved to the bottom right.

Screenshot_20200220_191539
Screenshot_20200220_192914

AdamSmasherDerby:
I think per trip annotations are the way to go, and will merge this into #439.

Before I do that, however, I have a few interface suggestions that make more sense to hash out in THIS discussion before carrying them over. 

Specifically, my suggestion would be to save a step by having the dialog popup always contain an editable box under "Notes."  It could either contain notes that are auto-populated, such as when the trip is automatically incremented, or allow for manual entry. No need for an "edit" button, just a "clear notes" button to clear the box if desired.

This also makes the meaning of the annotation symbol in the SK sheet straightforward: symbol = note present, no symbol = no note present.

I'd also suggest changing the "Remove" button text to "Remove Trip", as I feel the additional clarity there is valuable.

Here's an issue that I will need help addressing, however - having an arbitrary text box that users can type in raises question of data sanitization before passing to the backend for storage.  I assume there's a Java equivalent of[ "Robert); DROP TABLE Students;" ](https://xkcd.com/327/)that we'd like to avoid. If that's the case, what's the best way to do that?

frank-weinberg:
> always contain an editable box under "Notes." 

I still think this is too prominent given its anticipated frequency of usage but alas.

> changing the "Remove" button text to "Remove Trip"

This led down a bit of a rabbit hole: I felt that the Insert button should then be "Insert Trip" for consistency and because this is probably more intuitive for a user who thinks in terms of boxes on the paperwork and not a linked list of scoring trips (i.e. most users) - insert a trip in the box I clicked on, not the one before that.
And most of the time the user will want to enter points/SP/notes for the newly added trip, so we should navigate there. Which requires the same infrastructure as the navigation suggested in #441. And as we are redesigning the popup anyways, now is a good time to fit that in as well.
And then I addressed #448 as well as it is a small related fix.

Screenshot_20200221_075605

frank-weinberg:
> Bobby Tables

All we are doing with the user input here is store it as a string and paste it back into the textarea. It's never parsed in any way. Bobby Tables scenarios happen, when user data is used as (part of) a string that's parsed and executed, e.g. a database query. So we are safe.

AdamSmasherDerby:
OK - when I get home this evening I'll merge this into #439 and update it accordingly.

brian-brazil:
> As I see it per-trip annotations are the best place for the auto-completed initial trip annotation. It refers to a specific trip and deleting that trip makes it superfluous meaning it makes sense to delete the annotation with the trip.

That's a fair argument, but keep in mind that users will effectively never need to correctly delete that trip so it's a very weak argument. They will however clear the note reasonably often, so that plus the various jam-ended-weirdly cases are what we should mainly be considering.

> How is using the approach that's most straightforward to code overengineering?

We're adding a new concept of per-trip annotations with their associated complexity which all users will in principle have to learn about - but will never need in practice. That's over engineering to me, which is about more than just numbers of lines of code. We shouldn't be making things more complex for users just to save a handful of lines of code.

> Which requires the same infrastructure as the navigation suggested in #441. And as we are redesigning the popup anyways, now is a good time to fit that in as well. And then I addressed #448 as well as it is a small related fix.

I suspect that these issues might be fairly subtle to get right, so bundling it all into one PR may delay us unnecessarily as we work through those.

Visually the interface looks better now. I'd put navigation&close above the notes, as you'll likely want them more and they're more "core".

> I think per trip annotations are the way to go, 

What's your reasoning?

> So we are safe.

Even if we aren't, that'd be a bug and treated as such. There's lot of places users can enter arbitrary text that we need to protect against bugs and XSS.

> OK - when I get home this evening I'll merge this into #439 and update it accordingly.

For the sake of your sanity, I'd suggest waiting until this is merged. Also watch for the case where there's already a note.

brian-brazil:
> I suspect that these issues might be fairly subtle to get right,

Yip. This doesn't handle when a trip is auto-advanced after the dialog is opened, and the buttons are not in consistent places if you're on the first/last trip. It's also weird that you can to go to the Next trip when there is only an initial trip.

frank-weinberg:
>  This doesn't handle when a trip is auto-advanced after the dialog is opened

That's not handled correctly in current HEAD either. The popup doesn't receive WS updates. (And doing so would be sufficiently complex to warrant a separate PR. Which should then look at the LT popup as well.)

> It's also weird that you can to go to the Next trip when there is only an initial trip.

You can navigate to the same set of trips that you can open with a click. Changing that would mean you can open that trip by clicking on it but when you navigate out of it you can't navigate back. I'd consider that a bug.

AdamSmasherDerby:
> > OK - when I get home this evening I'll merge this into #439 and update it accordingly.
> 
> For the sake of your sanity, I'd suggest waiting until this is merged. 

Does that mean you believe this is close to being merged?  "Wait until this is merged" I can accept as long as t !=> never.

>Also watch for the case where there's already a note.

Good point, and one I'd had in mind.  Having the annotation infrastructure in place really does simplify #439, since all it has to do is append a note to the annotation field when it auto-increments trips.

frank-weinberg:
> the buttons are not in consistent places if you're on the first/last trip

That's fixed now.

brian-brazil:
> That's not handled correctly in current HEAD either. The popup doesn't receive WS updates. (And doing so would be sufficiently complex to warrant a separate PR. 

But it only matters as of this PR, as up to now we could get away with assuming that the user on a given screen is the only one mutating that screen. This change makes that no longer true. Thus it needs to be handled in the same PR that adds prev/next, otherwise we're introducing a bug. Accordingly I suggest splitting prev/next to another PR.

> Does that mean you believe this is close to being merged? "Wait until this is merged"

It was actually separate advice, as you'd likely just be making work for yourself.

The open question at this point is the best modelling of the annotations, and how to handle all the prev/next stuff if it remains in this PR. Everything after that should be fairly standard.

> What's your reasoning?

I note a lack of an answer here.

frank-weinberg:
> This change makes that no longer true.

How the flying hell does the addition of these buttons mean that suddenly multiple people will be entering data on the SBO screen?

brian-brazil:
> How the flying hell does the addition of these buttons mean that suddenly multiple people will be entering data on the SBO screen?

It doesn't. However we have both the user and the scoreboard itself performing mutations at the same time, and this change would need to take that into account.

frank-weinberg:
1. Take current HEAD.
2. Start a Jam
3. Enter a Star Pass for Team 1
4. Enter Points for Team 1
5. Before the trip advances click on trip 2 for the current Jam of Team 1
6. In the popup note how neither the SP indicator gets active nor do the blank points turn to 0 as the trip advances.

And you want to tell me that a quality of life navigation button also not showing up in this situation makes this so much worse that you'd rather not have those buttons at all?

AdamSmasherDerby:
> > What's your reasoning?
> 
> I note a lack of an answer here.

I agree with Speedy's arguments.  I didn't see a need to restate them.

brian-brazil:
> And you want to tell me that a quality of life navigation button also not showing up in this situation makes this so much worse that you'd rather not have those buttons at all?

Yes, I'd rather not add an incomplete minor feature if it's not going to work as expected for a user.

frank-weinberg:
This minor feature fixes the rather major bug that there are circumstances where the score can not be properly adjusted in the interface. This fix should not be held back by the fact that it has a minor issue that can be easily worked around and only occurs in unusual circumstances. (The number of SBOs that edit the SK sheet during a Jam will be small. The instances where they want to edit a trip that has only just started will be about as common as points during an initial trip.) Especially given that it's not obvious how to properly update the popup in the first place - adding WS handlers directly would require unregister capability we don't have. And even if that is solved, changing the score field while the user is repeatedly clicking the spinners will influence the resulting value. That'd be a way bigger problem than what we have now.

brian-brazil:
> Merged #451 into dev.

I've previously clarified with you the etiquette in relation to code reviews and merging thereof. I am disappointed that you chose to merge two PRs which not only have not completed basic code review, but which in addition also contain contentious issues that were still under active discussion and not yet  resolved to the satisfaction of all maintainers.

This is not healthy behaviour, and it is not fun for me to participate in a project where the basic ground rules are thus violated. If you do not wish to have a co-maintainer you could have said so, rather than crossing the line in this fashion.

frank-weinberg:
> I've previously clarified with you the etiquette in relation to code reviews and merging thereof.

It is my understanding of our agreement that PRs from third parties only need approval from one of us. And seeing how we routinely merge PRs that the other has not commented on, this aligns with practice.

>  you chose to merge two PRs

I only chose to merge #439. And I think I have made it rather clear that I would do so once there are no specific issues left with the code submitted. The merge of this one was an indirect consequence through Smasher having merged it into his branch.

> which not only have not completed basic code review

I have reviewed #439. And from the thread here it would seem hat the code taken over from this one has been reviewed by you and there were no problems with the code left.

> but which in addition also contain contentious issues

And the solution to those can't be to simply stall development forever. On #439 the only item you brought up that was not explicitly resolved was the judgement of "is it worth it". And as all arguments on that issue have been repeated ad nauseam already, what's the point of continuing to debate it?

And the same with the question of per-trip vs. per-jam annotations. All points have been made and it seems clear that between us there again would be a disagreement based on personal preference. Smasher merging this version into his branch is as good a tie breaker as any. (And for transparency: Before his first comment here, he sent me a PM asking how to best handle the interaction of the two PRs from a technical POV. In my answer I explicitly stated that I would accept his tie-breaking vote on the "where to annotate" question if it didn't go my way.)

Finally on the nav buttons I stand by my last comment: An issue that has no workaround is objectively worse than another facet on an issue that has one and does not come up significantly more often. And seeing how there are still conceptual problems to solve before a fix is possible, I don't think we should keep the community up in arms over #439 while we work on those. But if it helps to restore peace, I can commit to working on this item with highest priority. I do have an idea how dangling WS handlers can be avoided (in the existing handler check if the update is for the currently edited trip and if so update the popup). But I have no idea what to do about updates interfering with repeated spinner clicks.

> it is not fun for me to participate in a project where the basic ground rules are thus violated.

Do you want to imply that what happened for the last two weeks was fun for you? Because for me it definitely wasn't.

brian-brazil:
Whatever self-justification you have come up with is not relevant. Instead of trying to find a path forwards that was amenable to all, you instead merged changes which you know your co-maintainer had outstanding objections to. Doing whatever you want like this is simply unacceptable, and I am not interested in contributing to a project where that sort of thing happens.

frank-weinberg:
This essentially amounts to you requesting veto rights (for either of us) based on personal preference. And based on the experiences in the last months that is not an environment in which I am interested in contributing.

PR #515, opened 2022-06-11
Preserve aspect ratio in full-screen scoreboard displays

valhallasw:
See https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit for details.

Videos already scale correctly without `object-fit` (as `` and `<video>` tags are seemingly handled inconsistently), but specifying it explicitly doesn't hurt.

### DIFF
--- a/html/components/settings-tab.js
+++ b/html/components/settings-tab.js
@@ -256,16 +256,38 @@ function createScoreBoardViewPreviewRows(table, type) {
   var imageViewSelect = $('<label>Image View: </label>')
     .add(mediaSelect('ScoreBoard.Settings.Setting(ScoreBoard.' + type + '_Image)', 'images', 'fullscreen', 'Image'))
     .attr('ApplyPreview', 'Image');
+
+  var imageScaleSelect = $('<label>Image Scaling: </label>').add(
+      WSControl(
+        'ScoreBoard.Settings.Setting(ScoreBoard.' + type + '_ImageScaling)',
+        $('<select>')
+          .attr('ApplyPreview', 'ImageScaling')
+          .append('<option value="contain">Scale to fit</option>')
+          .append('<option value="cover">Scale to fill</option>')
+          .append('<option value="fill">Stretch</option>')));
+
   var videoViewSelect = $('<label>Video View: </label>')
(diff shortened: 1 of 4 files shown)


frank-weinberg:
I'm reluctant to just merge this as there may be people out there who rely on the prior behavior. This should probably be configurable. (But I'm leaning towards the new behavior as default.)

If you're not comfortable adding the configuration yourself, I can do that when I find time (maybe next weekend).

valhallasw:
I can take a look. Can you confirm whether the following rough sketch makes sense?

- Config entries `ScoreBoard.View_ImageScaling` and `ScoreBoard.Preview_ImageScaling` should be added to https://github.com/rollerderby/scoreboard/blob/dev/src/com/carolinarollergirls/scoreboard/core/admin/SettingsImpl.java
  - Do I need to modify any http api endpoints? Given the fairly generic setup I assume this is not necessary.
- https://github.com/rollerderby/scoreboard/blob/dev/html/views/standard/index.js should get an additional `WS.Register('ScoreBoard.Settings.Setting(ScoreBoard.' + view + '_ImageScaling)'` listener to apply a CSS class to the div
  - When upgrading, the config entries will not be set -- I assume the WS listener will have to handle a `null` value in this case? Or does the listener just not get called at all?
- The UI button/checkbox goes in https://github.com/rollerderby/scoreboard/blob/dev/html/components/settings-tab.js ? Need to dig into this code, it has been a while since I did extensive JS dev :-)

frank-weinberg:
That sounds right.

- Your assumption about no (explicit) api changes being needed is correct.
- Adding the entries to SettingsImpl.java gives them a default value that will be set whenever there is no value in the auto save. So they should not ever be null. But if they are not set on startup, the listener will not be called.
- settings-tab.js is the correct place. And for UI consistency it should be a button not a checkbox. Basically do the same as swapTamsButton and hideLogosButton.

valhallasw:
As screenshots explain more than a thousand words:
image

frank-weinberg:
Nice.
That's way more configurable than I anticipated. And the refactorings are also welcome improvements.

PR #559, opened 2022-12-11
Fix for issue #556

### LINKED ISSUE #556: Order issue of jams in overlay points per jam
Looks like jam order is not correctly displayed in overlay point per jam view.
Using last version V 5.0.9, jam order in overlay looks like it's order like player number :
jam 1, jam10, jam11, jam2, jam3.
Instead of jam 1 to 9 then 10.

Also is it possible to display directly the jam number IE on the separation line.

Thank's :)

aoanla:
Fixed jam ordering in pts per jam overlay by coercing "data-sort" to zero-padded version so it works with string sort.

### DIFF
--- a/html/views/overlay/index.js
+++ b/html/views/overlay/index.js
@@ -387,7 +387,7 @@ function jamData(k, v) {
 
   if ($(me).length === 0) {
     pointsPerJamColumnWidths();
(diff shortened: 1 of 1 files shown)


aoanla:
#556

PR #573, opened 2023-02-14
init roster view

### LINKED ISSUE #569: Announcer view
As an announcer for bouts, I would like a read-only view so that I can properly announce team rosters and also as a reference for in-game announcements.

atkinsonm:
Resolves #569 

Very basic initial "roster" view. Basically a copy of the team tab in the operator panel with inputs disabled and other buttons removed.

image


atkinsonm:
Things I'd like to change if I get more time and sharpen my webdev skills:

* Remove the tab view
* Ensure it is mobile-friendly (I haven't actually tried it on mobile yet)
* Add team logos if present
* Change the inputs/selects to normal text and format as appropriate

But functionally speaking this is a good first version IMO.

frank-weinberg:
At least removing the tab and turning inputs into text should be straightforward. Adding logos should also be relatively easy. I'll look into those later this week.

frank-weinberg:
This is how it looks now:

Screenshot_20230215_221026

atkinsonm:
Looks great! Thanks for taking it to the finish line

PR #581, opened 2023-03-18
Add Download Operators button

### LINKED ISSUE #580: [FEATURE REQUEST] - Include button to Download Operator settings only
As per comments made in the FB Group, there is a functionality request to include a Download Operators button

bullseye555:
Adding Download Operators button to _Download ScoreBoard JSON_ table in sb_data

For #580

### DIFF
--- a/html/components/data-tab.js
+++ b/html/components/data-tab.js
@@ -128,6 +128,8 @@ function createDataTab(tab) {
     rulesets.find('tr.Content.Selected').each(function () {
       paths = paths + ',ScoreBoard.Rulesets.Ruleset(' + $(this).attr('id') + ')';
     });
+	
+	
     var d = new Date();
     var name = $.datepicker.formatDate('yy-mm-dd_', d);
     name += _timeConversions.twoDigit(d.getHours());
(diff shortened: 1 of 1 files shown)


frank-weinberg (changes requested):
- Don't combine independent changes into a single PR
- Your indentation is broken - the file is indented using spaces, some of your code uses tabs. You can very obviously see the problem in line 97, but other parts are affected as well.
- The way you are doing single deletes breaks things:
  - Using toggleClass means that if you click Delete then reject in the dialog and click Delete again, the element will be unmarked and not deleted. Use addClass instead.
  - If the deletion is rejected in the dialog, you do not reset the class. So if you click a different Delete button later on, the item you rejected to delete earlier will be deleted at that point.
  - The updateSeletedUrlSelectedDelete function changes the set of elements that "Download Selected" would fetch to the items that you are deleting (and it does nothing else).

bullseye555:
Thanks for the feedback @frank-weinberg 

It's my first time working in JS and noted in the separate pushes for different changes (when I was working with Git at my last job they got us to combine everything...)

I'll split out the changes and push through just the Download Operators to this PR and then will fix up the Single Delete tomorrow/during the week and do a new PR for that

bullseye555:
@frank-weinberg Single Delete changes removed - this is now purely only for Operator Download. Added line indenting converted to spaces instead of tab

frank-weinberg (approved):

frank-weinberg:
@bullseye555 Given the amount of requests we have around this screen, I'm looking int overhauling it in a way that does away with the folding sections. It's likely that this will cover the single game download as well.

bullseye555:
Thanks @frank-weinberg - I just fixed the single delete, too... 

I'll push it anyway, so then that FR can be completed - and included in the next release, if you haven't made the changes to that screen

Issue #577, opened 2023-03-10
[Feature Request] Deleting single entries


reflexdd:
At the moment we have to select a element (Game/Team etc.) and then click on Show "Delete Scoreboard" and "Delete Selected Elements"

This is not really intuitiv. I would expect to be able to have e delete button right near to the item.
It took me a couple of minutes to figure out, that it has to be done in an extra section and lately i was told the same by some league members who tried to manage some gamedata.

My suggestion would be: place a delete button next to the Edit/View Button of each element.

Thank you 
REFlex

PR #582, opened 2023-03-18
Add single entry delete

bullseye555:
Add delete button for each entry per #577


frank-weinberg:
Closing this as it is superseded by #583 (but it would have been mergable if that hadn't been the case).

bullseye555:
Thanks @frank-weinberg 

Hoping to begin helping out with some more of the minor requests

PR #597, opened 2023-03-26
add migration functionality

### LINKED ISSUE #574: [FEATURE REQUEST] - Migration functionality
I have a powershell script that I've written that will move all the images, AutoSave files, statsbooks, custom penalties/rulessets, etc from a previous version of CRG (eg 5.0.9) and place it in a destination folder (which will be the new version, eg 5.0.10)

But while the logo's etc auto "import", I'm finding that even though CRG is loading the most recent autosave from the previous version, nothing that needs importing (eg from a JSON) is being carried over.

What I'd like is an option on the _Data / Teams / Rulesets_ page where you can select the CRG Parent folder (EG: **D:\CRG\CRG5.0.9** - where scoreboard-Windows.exe is) and have it import the following:

1. Folder: html\images
2. Folder: html\custom (where it's not part of the base "custom" set)
3. Files: *.xlsx (ideally maintaining folder trees) [assumes user has a blank statsbook loaded somewhere, and has exported previous games as a Statsbook]
  -- This also, ideally, includes pre-populating the Blank Statsbook field with the same value as that in the source version
4. Files: *.json (excluding any core CRG JSONs, again ideally maintaining folder trees) [assumes user has exported previous games or other data as a JSON]
5. Data: Import _Games_, _Teams_, _Rulesets_ etc {essentially, perform the _Download All Data_ on the source version, and then upload that into the new scoreboard

If this was easier to have both scoreboards running, and to perform the migration, you connect the "new" scoreboard at (for example) port 8001 to the "old" scoreboard at (for example) 8000 that would be ok too - open to suggestions on how best to complete something like this [the idea is that it saves users time in have to "reconfigure" everything each time there's a version change]

frank-weinberg:
When starting up without an autosave directory (including first startup after unzipping) CRG will look for autosaves in all sibling folders of the current folder and import user data from the one with the most recent autosave.

There is a new command line switch `--import=<path>` or `-i=<path>` to control importing behavior:
- If `<path>` is empty, importing is skipped even if there is no autosave.
- If `<path>` is nonempty, CRG will attempt to import from the given path, even if an autosave is present.
  - Be aware that doing this will overwrite your current autosave.

Closes #574

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/Main.java
+++ b/src/com/carolinarollergirls/scoreboard/Main.java
@@ -5,6 +5,18 @@ import java.awt.Font;
 import java.awt.Toolkit;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.CopyOption;
+import java.nio.file.DirectoryStream;
+import java.nio.file.FileAlreadyExistsException;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.swing.BoxLayout;
 import javax.swing.JFrame;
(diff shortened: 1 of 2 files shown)


spikenheimer:
this update is awesome.  unless you are thinking you're setting up a clean instance to do testing on and you think you're going insane because somehow all your previous settings / files are showing up magically.

aka you're a sorcerer frank.  kudos!

PR #613, opened 2023-05-07
added LiberationSans font files, added font-face rules, fixed font-family

webdingens:
Currently views require the fonts listed in the CSS font-family rule to be installed on the machine. By providing the referenced font files via @font-face rule the views should look the same on every machine and not use the fallback the browser thinks is okay to use.
Tested on
Win11: Firefox current, Chrome current.
MacOS BigSur: Firefox
Fixes the bug on Mac Firefox where the score is larger than the box.


frank-weinberg:
While we're at it, wouldn't it make sense to set the font for all screens not just those 3?

webdingens:
Created a styles folder like the javascript folder. jquery-ui sets Arial as a font for some elements. Without that the styles could probably be injected like the `index.css`.

webdingens:
Edit: moved the styles import to core.js and removed the !important, which was unnecessary.

PR #653, opened 2023-07-18
Don't make both teams the other team on SK sheet view

### LINKED ISSUE #652: Score sheet in scoreboard doesn't show which team a timeout was charged to
Recently at a tournament I was looking back over the game data to fill out the parts of the statsbook that didn't export conveniently from the scoreboard. (JRDA statsbooks require marking how many timeouts and official reviews were used by each team in each period.)

The "Sheets" section of the scoreboard would be a great place to get that information from, but I found a bug where the Timeout line for both teams' scores sheets just says "Team TO Other team".

image

frank-weinberg:


atkinsonm (approved):
LGTM 👍🏻

PR #659, opened 2023-08-21
feat(views/roster): :sparkles: display roles on roster view

### LINKED ISSUE #654: Feature: highlight players on track in roster view
As an announcer, I would like the roster view to highlight players on the track so I can focus my attention on players in play in the current jam.

Bonus: highlight players currently in the penalty box.

### LINKED ISSUE #657: Filter non-skaters from roster view
In the roster view, we currently show all team members. It would be nice if we could filter out "not skating" members to make fit easier for users.

atkinsonm:
## Description

Show roles and color code on roster view same as on lineup tracking

## Why

Closes #654 

## Testing

Tested locally in a test game with lineup tracking

image

When lineup tracking is disabled, players currently appear in "bench" by default

image

## Notes

* I added the roles column to the roster view because the shades of green used on the lineup tracking are close enough that they could be confused with one another without context. From an accessibility standpoint, we should probably keep the names in addition to colors and maybe also change to more contrasting colors.
* `NotInGame` is just the value of `role` but itself isn't very friendly to read.
* `Bench` when lineup tracking is disabled should probably be replaced with "not tracking" or similar to avoid confusion.
* I haven't gotten penalty box highlighting/displaying on the roster view yet. Ideally I'd like to see an indicator of when a player is in the box and a separate color. The name or code of the penalty being served would be a nice to have. This point may saved for a follow-up issue.

### DIFF
--- a/html/components/roster-tab.css
+++ b/html/components/roster-tab.css
@@ -26,6 +26,7 @@
 #Teams table.Skaters th.Title { font-size: 150%; } 
 #Teams table.Skaters th#skaterCount { font-size: 100%; font-weight: normal; } 
 
+#Teams table.Skaters tr[flag='ALT'] { display:none; }
 #Teams table.Skaters tr[role='Pivot'] { background-color: #00ee00 !important; } 
 #Teams table.Skaters tr[role='Pivot']>td { background: #00ee00 !important; }
 #Teams table.Skaters tr[role='Jammer'] { background-color: #00cc00 !important; } 
--- a/html/components/roster-tab.css
+++ b/html/components/roster-tab.css
@@ -25,3 +25,10 @@
 
 #Teams table.Skaters th.Title { font-size: 150%; } 
 #Teams table.Skaters th#skaterCount { font-size: 100%; font-weight: normal; } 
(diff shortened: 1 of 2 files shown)


frank-weinberg:
NotInGame rows would be removed anyway when #657 is done.
For Bench I'd leave the field empty.
There is also `Ineligible` for skaters that have fouled out, been expelled or have to sit for 3 jams after causing a call-off. That feels like it might be useful info if we can find a term to display that people intuitively understand.

frank-weinberg:
How about merging role and flags columns like this:
grafik

atkinsonm:
> How about merging role and flags columns like this:
> grafik
> 

I like that. When I return from RollerCon I can make those changes

frank-weinberg:
I already have them ready. (That's how I created the screenshot.) I'll see if I can push them to this branch. Else I'll open a new PR.

atkinsonm:
> I already have them ready. (That's how I created the screenshot.) I'll see if I can push them to this branch. Else I'll open a new PR.

Thanks! Code looks great 👍

PR #681, opened 2023-12-28
feat: remove all skaters button

### LINKED ISSUE #670: Feature Request: Clear Team of Skaters
as i sit here updating rosters for home teams with a slew of updates, i find it tedious to manually hit remove / tab over to yes for each skater in the roster (i'm just copypasta-ing the whole team back in from a spreadsheet the captains sent) i realize it would be great to be able to just remove all skaters at once.  for sure it probably needs at least 1 maybe 2 confirmation screens but it would be nice.

i'm sure its less trivial to do this than the UI implies but just a thought for a low priority feature enhancement.

atkinsonm:
Adds a button to the teams tab that allows an operator to remove all skaters from a team with one click. Closes #670 

It accomplishes this by looping through the skater IDs and setting all skaters to `null`.

## Demos

Remove All buttons per team on the teams tab (this is from in-game but it also shows on the data screen)

image

On click dialog. Choosing no closes the dialog with no changes.

image

After confirming, all skaters are removed for the selected team only

image

### DIFF
--- a/html/components/teams-tab.js
+++ b/html/components/teams-tab.js
@@ -272,11 +272,20 @@ function createEditTeamTable(element, teamPrefix, isGameTeam) {
     .append('<col class="Button">')
     .append('<thead/><tbody/>')
     .children('thead')
-    .append('<tr><th></th><th class="Title">Skaters</th><th></th><th id="skaterCount"></th><th></th></tr>')
+    .append('<tr><th></th><th class="Title">Skaters</th><th></th><th id="skaterCount"></th><th id="removeAllButton"></th><th></th></tr>')
     .append('<tr><th>Number</th><th>Name</th><th>Pronouns</th><th>Flags</th><th>Add</th>')
     .append('<tr class="AddSkater"><th/><th/><th/><th/><th/></tr>')
     .append('<tr><th colspan="5"><hr/></th></tr>')
     .end();
+
+  $('<button>')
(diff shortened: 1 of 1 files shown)


frank-weinberg:
I'm wondering if a RemoveAllSkaters call in the backend would be better than doing the iteration in the frontend. Feels more in line with how we do such things elsewhere in the scoreboard.

But I have just done a complete rewrite of the frontend over the last few weeks. So this will unfortunately need to be redone anyways. (It was actually on my roadmap of things to add for 2025.* already.) 

So we could add this version to 2023.* if we decide it doesn't break the feature freeze criteria:
- No features that have non-negligible risk of introducing new bugs - I think this ok in that regard
- Nothing that could change the SBO workflow - depends on if we assume setting up teams is part of the SBO workflow.

atkinsonm:
I'm completely open to a better way and agree that backend is probably better than frontend here; this is just how I could figure out making it work.

Agree with your assessments on non-breaking criteria.

spikenheimer:
@atkinsonm sorry, let my email clog up over the holidays,  this looks great, can’t wait to use it!
@frank-weinberg my dev brain would agree this fits best as a call to a backend thing probably so just my agreement w y’all’s above assessments…

frank-weinberg:
This will be a backend call in 2025.*. Implemented it yesterday.

PR #682, opened 2023-12-28
init feature request template

atkinsonm:
Adds an issue template for feature requests. Now that we have a bug report template, the create issue screen looks like this:

image

The "open a blank issue" link can be hard to find. Adding a feature request template will make it more obvious how we expect folks to submit feature requests, and also give us the opportunity to request certain bits of info from them.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: "[FEATURE REQUEST] One-liner description of feature request"
+labels: enhancement
+assignees: ''
+
+---
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: "[FEATURE REQUEST] One-liner description of feature request"
+labels: enhancement
```
frank-weinberg: We have a "Feature Request" label. That should be used (and the title tag is unnecessary)

atkinsonm: Will fix. Title-wise it seemed like at least some issues used that convention, but I'm all for reducing redundancy between the title and labels.

PR #695, opened 2024-05-09
Fix start mid-game functionality

### LINKED ISSUE #693: [BUG] Start mid-game functionality does not set some values correctly
**Describe the bug**
In the current release version, attempting to start a game mid-game does not currently handle some of the input values:
* "Period" - the game will be started in period 1, even if 2 is selected here.
* "Timeouts Taken" and "Reviews Taken" - no timeouts or official reviews will be created, even if a non-zero value is selected for these fields.
* "Prior Jam" - although the label says "prior jam", the scoreboard will actually be started with this value as the _next_ jam.

**To Reproduce**
1. Click the 'Start New Game' button in the tab bar, then click "Start Mid-Game" in the modal to show the form.
2. Set Period to 2.
3. Set "Prior Jam" to 10.
4. Set 1 timeout taken for each team.
5. Start the game

image

**Expected behavior**
I would expect this to start the game in period 2, jam 11, with each team having 2 timeouts remaining. Instead, it starts the game in period 1, jam 10, and each time has all three timeouts available.

image

**Version:**
v2023.5

**NOTE:** This behavior was working at one point. Running a git bisect, it looks like something in #630 broke some of the start midgame functionality.

frank-weinberg:
Closes #693

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/game/GameImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/game/GameImpl.java
@@ -864,6 +864,7 @@ public class GameImpl extends ScoreBoardEventProviderImpl<Game> implements Game 
     }
 
     protected boolean quickClockControl(Button button) {
+        if (quickClockAlwaysAllowed) { return false; }
         long currentTime = ScoreBoardClock.getInstance().getCurrentTime();
         long lastTime = lastButtonTime;
         boolean differentButton = button != lastButton;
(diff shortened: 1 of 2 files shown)


atkinsonm (approved):

goldenapples:
I'm seeing the period number and timeouts taken issues are fixed. The jam number is still off by one, but in the opposite direction - if I try to start game with "previous jam" as 10, the game will be actually started after jam 11:

image

frank-weinberg:
Good catch, thank you. I'll fix that before merging.

PR #767, opened 2024-09-15
re-add compatibility for iOS 9.3

### LINKED ISSUE #731: [BUG] Older browsers not working
**Describe the bug**
We are using older iPads, they have been functioning fine for ePLT with v2023. However, with the frontend update of v2025 it's not working anymore. It looks like the Option Editor is not function as it should. It just stays how you see it on the screenshot. I cannot press a button and the Option Editor stays at the front. 

**To Reproduce**
Open any input method and the option editor is completely on the screen.

**Expected behavior**
Being able to use ePLT ;) 

**Screenshots**
Sorry for the photo of the screen, I had a hard time creating a screenshot with the tablet :) 
IMG20240706164048-3

**Version:**
2025beta2

**If this issue occurred on a smartphone, tablet, or other device, please complete the following information:**
- iPad 3rd Generation
- OS 9.3.6
- Safari 9
Unfortunately I can't upgrade Safari or install a newer browser. :(  

**Additional context**
I completely understand if older browsers aren't supported anymore.
It would be a shame as old tablets are still fine for such a thing and I don't want to create eWaste.

frank-weinberg:
closes #731 

Support is not perfect: iOS does not support `display: grid` and some calculations with CSS variables. But as far as I can tell without access to an affected device, the relevant screens (JT, single team PLT and PBT) degrade in a way that keeps them usable.

### DIFF
--- a/html/components/game-advance/index.js
+++ b/html/components/game-advance/index.js
@@ -1,7 +1,5 @@
-'use strict';
-
 function advanceGame() {
-  let url = new URL(window.location);
+  var url = new URL(window.location);
   url.searchParams.set('game', WS.state['ScoreBoard.CurrentGame.Game']);
   window.location.replace(url);
 }
--- a/html/components/igrf-tab/index.js
+++ b/html/components/igrf-tab/index.js
@@ -1,5 +1,3 @@
-'use strict';
-
 $('table.Officials')
   .clone(true)
   .attr('officialType', 'Ref')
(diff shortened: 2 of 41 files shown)


sandlerj:
Hi! New to this project, but as a derby player, user of this application, and software engineer who wants to join on here, I wanted to ask about the PR - 
If rolling back some js functionality to support older browsers, would there ever be an intent to stop supporting old browsers to that modern JS features can be used? Needing to stick to ES5 indefinitely seems like it could be difficult to maintain going forward.

Issue #405, opened 2019-11-29
Tooltips on the buttons that explain what they do


voidptr:
Since the start/stop jam and TO, etc, buttons now change titles, it can be hard to remember what does what when setting up shortcuts. To ease the transition, I'd like to see a mouse-over tooltip over the buttons that describes what they do. 

Similar tooltips on all the other buttons would be neat too.

brian-brazil:
What they do changes over time, and we don't use tooltips elsewhere. There's also a possibility that the tooltip would get in the way during actual gameplay.

frank-weinberg:
I can see how tooltips might be useful for new users. But doing them in a way that doesn't interfere with operation could be tricky and in order to keep the interface consistent they should probably be done for at least all buttons on a tab if not across the whole UI which makes this a rather large task. (Especially figuring out good texts may easily be more work than implementing the tooltips.)

charlie-chaperon:
I agree that it could be useful for new score board operators. Maybe you could add a button in the top line "turn tooltips on/off" which is deactivated by default? That way you can also use it to quick check / remember how a (rarely used) button works.

gdunn:
It's easy enough to add a title attribute to where we want tool tips. The hard part will be enabling/disabling it. Would need to feed in an option (or options) along with the gameid
![Screenshot 2022-07-24 174413](https://user-images.githubusercontent.com/1120941/180637563-6f531af9-bc63-4317-a177-fe2983fdbaff.png)

PR #825, opened 2025-02-15
Add tooltips to buttons with dynamic visibility support

tanzoniteblack:
I was taking a look at https://github.com/rollerderby/scoreboard/issues/405 and thought that a decent middle ground on solving this is instead of adding a new button to hold a variable that will dictate whether to show tooltips, we could instead use hovering while holding a modifier key on the keyboard to control that display logic.

This PR adds integrated tooltips across score control buttons to provide contextual guidance to users, but only when the `Control` key is being held & the user is actively hovering over an element with the class `tooltip` that has a child element with the class `tooltiptext`. I couldn't add the `tooltiptext` span directly as a child of the `button` or the button's display rules interfere with getting the tooltip to show up on top of the other elements on the page.

Notes: 

Because of the way browsers detect key presses, the user will have had to interact with the page in some format before the control + hover functionality will work. This could be pressing a button or even just clicking anywhere on the page.

I am in zero way tied to _where_ this javascript code should live nor what the styling should be, so please feel free to adjust that or make suggestions on what changes I should make!

And finally, I only added a few tooltip texts because I'm still learning how to use this program and am the prime audience for _needing_ these help bubbles!

image

### DIFF
--- a/html/components/sk-input/index.html
+++ b/html/components/sk-input/index.html
@@ -29,14 +29,17 @@
   <div class="sbGroup sbStack">
     <div class="ScoreRow">
       <div class="sbStack">
-        <button
-          class="sbKeyControl sbImportant"
-          sbAttr="id:-: 'Team[Team]ScoreDown'"
-          sbProp="disabled: TripScore: == 0"
-          sbSet="TripScore: -1: change"
-        >
-          Trip Pts -1
-        </button>
+        <span class="tooltip">
+          <button
+            class="sbKeyControl sbImportant"
+            sbAttr="id:-: 'Team[Team]ScoreDown'"
+            sbProp="disabled: TripScore: == 0"
+            sbSet="TripScore: -1: change"
+          >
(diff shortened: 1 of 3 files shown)


frank-weinberg:
I see a number of problems with this approach. But they may all be solvable.

- The people most likely to benefit from the tooltips are the ones least likely to know that they can activate them with a control key. But as I'm typing this comment, I had the thought that maybe we can add a button that toggles permanent tooltips and itself has a tooltip informing about the control key for temporary tooltips.
- Aside from that issue the main challenge is finding good texts for the tooltips. Though I'm not opposed to adding the infrastructure first and letting the texts come over time.
- As for the implementation, this should probably be done as an sbTooltip attribute on the button/element that is then processed in WS.js (like the other sb* attributes) to generate the actual html/css.

PS: Sorry for the late response - I have been super busy with non-CRG stuff the last few months.

PR #844, opened 2025-06-03
Migrate from Ant to Gradle

cimnine:
I've started toying with the Scoreboard. Originally, I only wanted to implement [mDNS](http://www.multicastdns.org/)-based [DNS-SD](http://www.dns-sd.org/). (Because when streaming, I hate to figure out the IP of the server each time. Having `http://scoreboard.local:8000` would be much nicer. See separate PR https://github.com/cimnine/scoreboard/pull/1.)

But then I couldn't get the [Ant](https://ant.apache.org/) setup running quickly enough. When I had it running, it wasn't working particularly well with my preferred IDE, IntelliJ (e.g. no debugger). And when I added the jar of the library for the mdns/dns-sd stuff, it had further dependencies – of course. I almost stopped there entirely.

Instead, I've started to migrate the project to [Gradle](https://gradle.org/) – with quite some success.

With the gradle-based build, I see (and have already experienced) a few benefits:

- Works out of the box with various IDEs such as IntelliJ and Eclipse
- The gradle wrapper allows getting started quicker, like without having to install ant (or even gradle) (and it even selects/fetches the correct JDK if not present, when I understood that correctly).
- The subsequent builds are much faster.
- There is dependency management (which will make updating dependencies easier, some are years old and have known vulnerabilities)
- The startup scripts for Unix/Linux and Windows are managed/generated by Gradle.
- Personally, I believe not many people are familiar anymore with ant.

Are you open to such a fundamental change to this project?

Working:

- [x] Compile
- [ ] Tests
- [x] Run
- [x] UberJar

Need work:

- [x] Bundle html, config, logs, etc. in the release artefact
- [ ] Test on Windows
- [ ] Git-based versions
- [ ] More…?
- [ ]


frank-weinberg:
I am not particularly attached to ant, so in principle I am open to such a change. Some considerations:

- I don't like the idea of outside factors affecting my software. Even more so, if they cause work on my end. So any `:latest` dependency is a no go for me.
  - In addition to a new version triggering a need for change, with dependencies not bundled that need could also arise from dependencies no longer being available on the remote. How big is this risk with the proposed setup?
- CRG is very intentionally on a Java 8 requirement. While I would be ok with having a higher requirement for development, I'd still like to see the unit tests run with Java 8 in the CI. The PR looks like you've moved to all Java 21.
- I'd prefer for the project to be tied to a specific platform as little as possible. Using GitHub Actions is a move in  the wrong direction in that regard.

cimnine:
Thank you for your thoughts. I'll continue to toy with this PR for my personal enjoyment for now. If I get to a point where I consider this change ready and worthwhile, I'll let you know.

frank-weinberg:
Closing this as not ready for now. Feel free to reopen or create a new PR when you have something ready for review.

PR #859, opened 2025-06-30
Selecting Role for a Stored Official for a game doesn't enable 'Add' Button

### LINKED ISSUE #858: [BUG] Selecting Role for a Stored Official for a game doesn't enable 'Add' Button
**Describe the bug**
I'm hoping im just being dumb here but I've been testing some of the features I haven't yet got round to in 2025.6 and was able to successfully store an official, however when I select that official on both NSOs and SOs, the Add button doesn't enable once I have slected a role. 

I was able to add a stored official by typing anything in the boxes and then backspacing and also by using chrome dev tools 😋 however the dropdowns don't remove the disable class and tag on the 'Add' button

**To Reproduce**
Steps to reproduce the behavior (Assumes Stored Official Available):
1. Go to 'Games / Teams / Rulesets'
2. Click on 'Edit' on any game (default Black vs White used)

On either NSOs or SOs:

3. Click the dropdown for a role and select one
4. Click the dropdown for a Stored Official and select one
5. See 'Add

Steps 3 and 4 are interchangable

**Expected behavior**
'Add' Button becomes enabled

**Screenshots**

Image

**Version:**
2025.6

 - OS and version: MacOS 12.7.4, Chrome, 137.0.7151.120 (Official Build) (x86_64)
 - OS and version: Windows 10, Chrome, 137.0.7151.104 (Official Build) (x86_64)
 - OS and version: ipadOS 18.3.1, Chrome 138.0.7204.56, (FYI the HNSO layout on my ipad pro is a little... wide...)

queekusme:
Fixes #858

I haven't mad any changes to the 'Role' select or the igrfUpdateAddButton as I wanted to confirm exactly what the intended behaviour was first.

At the moment (no changes) the function checks the Name field and Prepared Official select for changes, if changes are found then the Add button will enable.

Although no bug occurs, logically it feels more correct to have either of those two (as current) **AND** a Role entered before the 'Add' button is enabled.

This would require some changes to the JavaScript file which I am happy to make however wanted to confirm if this was desired functionality before making the change.

I fly out to the RDWC tomorrow (1st) so will review any comments when I'm back

### DIFF
--- a/html/components/igrf-tab/index.html
+++ b/html/components/igrf-tab/index.html
@@ -191,7 +191,7 @@
       </tr>
       <tr class="AddPreparedOfficial sbSubHeader">
         <th colspan="2">
-          <select class="Prepared">
+          <select class="Prepared" sbOn="change: igrfUpdateAddButton">
             <option value="">Select Stored Official</option>
             <option
               sbContext="/ScoreBoard"


frank-weinberg:
We currently allow to unset the role after an official has been entered and not having a role assigned doesn't break anything but might be useful in some edge case. So I'd not force the role to be set.
Also happy to talk about this in Innsbruck, if we have an opportunity.

queekusme:
This is why I wanted to check first, logically it makes sense but as you
say its possible to remove the role after the fact

If I interpret you correctly, name is the most important part, so not
allowing ‘Add’ until it’s entered (as current) is correct.

And yeh, having a quick chat in Innsbruck would be great, I’ll be posting
on my mastodon (https://mastodon.scot/@queekusme) so if you have that feel
free to message or you can use that to find me… I’m volunteering on the
AV/Camera crews though so will be around and about a bit

Issue #877, opened 2025-09-22
VS Code and ant are not playing nice together.


craig-spannring:
The "Language Support for Java™ by Red Hat" extension for VS Code 
compiles the code while you are typing to help provide IntelliSense.  
The downside of this, it will produce .class files even if there are fatal 
compile errors.
    
In cases where VS Code has encountered a compile-time error and 
created a bogus .class file, the Ant build system notes that the .class 
file is newer than the .java file and will skip rebuilding it and report 
that everything is fine.  (Obviously everything is not fine.)

craig-spannring:
A solution for this is in pull request #878.

PR #878, opened 2025-09-22
Make ant vscode play together

craig-spannring:
The interaction between VS Code's build system and ant's build as defined in build.xml 
is less than ideal.  For details see issue #877.

This change solves the issue by using separate outputPath directories 
for Ant and VS Code.  I've kept the same path for VS Code (i.e. "bin") 
and told Ant to place the output in "build-ant".

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -82,7 +82,15 @@
 			<isfalse value="${isRelease}"/>
 		</condition>
 		<mkdir dir="${version.dest.dir}"/>
+		<mkdir dir="${version.vscode.dest.dir}"/>
 		<propertyfile file="${version.dest.dir}/${version.release.file}">
+			<entry key="release" value="${version}"/>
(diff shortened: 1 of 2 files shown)


craig-spannring:
The current version makes VS Code and ant play together about as nicely as they can.

build.xml
```diff
@@ -97,13 +117,13 @@
 		<mkdir dir="html/images/sponsor_banner" />
 		<javac srcdir="${src.dir}"
 			destdir="${dest.dir}"
-			deprecation="${compile.deprecation}"
+                        deprecation="${compile.deprecation}"
```
frank-weinberg: Here and in the changes below the indentation is off.

craig-spannring: Leading spaces have now been converted to tabs.

Issue #883, opened 2025-10-03
[BUG] Trailing whitespace in login name can lead to non-intuitive behavior


craig-spannring:
**Describe the bug**
When entering the "Scoreboard Operator Panel" the system will ask for a login name. 
The system differentiates between "foo" and "foo ".  If a user enters "foo" the first time 
they use the scoreboard and creates new key controls, but then later on tries to login 
using the name "foo ", the keyboard controls they expected will be missing.

**To Reproduce**
1. Enter the "Scoreboard Operator Panel"
2. When the "Operator login" dialog pops up enter your name, with no trailing whitespace and hit enter. 
3. Click the "Edit Key Controls" button and assign keys to some of the buttons. 
4. Click the "Edit Key Controls" button to exit the key editing mode. 
5. Logout of the Scoreboard Operator Panel. 
6. In the "Operator Login" dialog enter your name, but add a space at the end. 
7. You will not see any of your key controls
8. It also seems to add unnamed "ghost" operator entries in the Data Management screen.  These ghost entries can't be selected, deleted, or downloaded. 

**Expected behavior**
The operator will expect to see the keys mapped to the buttons they assigned them too.

**Version:**
v2025.5

**If this issue occurred on a computer, please complete the following information:**
 - MacOS 14.7.1 (Sonoma) and Linux Mint 22.1
 - Safari 17.6 and Chromium 139.0.7258.138

**Additional context**
I first noticed this after a tournament where we had some pretty experienced SBOs.

craig-spannring:
Should be fixed in pull request #884.

frank-weinberg:
I'm not sure I agree with the assumption here. To me "foo" and "foo " are different strings and I would expect them to be different logins.
And as far as I am aware this is also how logins work on most services.

JeneralPain:
You are right that they would be treated separately, however, most do not accept spaces to avoid “oddities with how some forms, databases and SSO apps” hands the space.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

frank-weinberg:
Then maybe we should do the same. 
And either way this is more a feature request than a bug. And one that could cause people who use a login that starts/ends with a space to lose access to their settings. So one that's better moved to the next major release

JeneralPain:
[...]

PR #884, opened 2025-10-03
Delete leading and trailing spaces on enter login name.

craig-spannring:
The name passed into html/nso/sbo/index.js:_login() is now stripped of all leading and trailing spaces before operating on that name.  This way we'll never use names that differ only in trailing whitespace.

This should fix bug #883


frank-weinberg:
superseded by #907

Issue #929, opened 2026-04-04
[BUG] All operator buttons (Start Jam / Stop / Timeout) become unresponsive after Undo


a1ly404:
## Describe the bug

After pressing the **Undo** button during or between jams, all operator control
buttons — Start Jam, Stop Jam/TO, and Timeout — become permanently unresponsive
for the rest of the session. The scoreboard UI continues to load and display
correctly, but no jam can be started and no timeout can be called. The only
recovery is a full application restart.

The root cause is that `restoreSnapshot()` in `GameImpl.java` sets a guard flag
`restoreRunning = true` at the start of the restore operation but only clears it
back to `false` on the normal happy-path exit. If any exception is thrown during
the restore — for example due to the box trip snapshot reordering introduced in
commit `ebb90645` — the flag is never cleared and permanently blocks the three
button handlers:

```java
// startJam() — GameImpl.java
if (!isInJam() && !restoreRunning && !isOfficialScore() && ...) { ... }

// stopJamTO() — GameImpl.java
if ((...) && !restoreRunning && !isOfficialScore() && ...) { ... }

// timeout() — GameImpl.java
if (!restoreRunning && !isOfficialScore() && ...) { ... }
```

## To Reproduce

Steps to reproduce the behaviour:

1. Start the scoreboard (v2025.8 or v2025.9).
2. Start a new game.
3. Start the period clock (click **Start**).
4. Start a jam (click **Start**).
5. While the jam is running, press **Undo**.
6. Observe: the scoreboard reverts state, as expected.
7. Now attempt to press **Start Jam** again.
8. **Nothing happens** — the button fires but the jam does not start.
9. Attempt **Timeout** — also has no effect.
10. Attempt **Stop** — also has no effect.
11. All three buttons remain permanently unresponsive until the application is
    fully restarted.

Alternatively: the issue can be triggered more reliably by performing an Undo
when a skater has an active box trip (penalty box entry recorded), which causes
the box trip snapshot restoration path introduced in `ebb90645` to throw.

## Expected behaviour

After an Undo, the scoreboard should return to the previous state and all
operator buttons should remain fully functional. `restoreRunning` must be
guaranteed to return to `false` regardless of whether the restore succeeds or
fails.

## Root Cause Analysis

Two commits introduced between v2025.7 and v2025.8 interact to cause this:

**Commit `58904866`** — "protect against overwriting snapshots during a restore"

Introduced the `restoreRunning` boolean flag and added it as a guard to
[...]

PR #930, opened 2026-04-04
fix: for All operator buttons (Start Jam / Stop / Timeout) become unresponsive after Undo

a1ly404:
Created this bug in the scoreboard repo.  [here](https://github.com/rollerderby/scoreboard/issues/929)

fix: wrap restoreSnapshot in try/finally to guarantee restoreRunning is cleared

If restoreSnapshot throws after setting restoreRunning=true, the flag
was never cleared, permanently blocking startJam, stopJamTO, and timeout
for the rest of the session. All three check restoreRunning as a guard,
requiring a full app restart to recover.

try/finally guarantees the flag is reset regardless of success or failure.

Affects v2025.8 and v2025.9. Not present in v2025.7.
Root cause: commit 58904866, widened by ebb90645.

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/game/GameImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/game/GameImpl.java
@@ -953,27 +953,30 @@ public final class GameImpl extends ScoreBoardEventProviderImpl<Game> implements
     }
     protected void restoreSnapshot() {
         restoreRunning = true;
-        ScoreBoardClock.getInstance().rewindTo(snapshot.getSnapshotTime());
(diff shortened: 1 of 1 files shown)


a1ly404:
@frank-weinberg thank you so much for everything you do for the rollerderby communitee! We ran into this problem during one of our tournaments. We reverted back to v2025.7 to fix. This was also causing an ad hoc game and any other games that we imported to have the scoreboard buttons to have no effect. We first noticed it when the time to derby clock stopped counting down.

a1ly404:
I will close this until I can find the actual root cause

PR #936, opened 2026-04-06
replace `white` → `#f9f9f9` for luma key alpha

yetzt:
i'm using a hardware video switcher. in order to be able to use luma key instead of chroma key to remove the background, the purely white elements need to be slightly less white. otherwise all white elements will be interpreted as transparent and potentially illegible on stream. this is a great improvement in comparison with using chroma key, since it allows teams that use colors containing green. the visual difference is negligable.

### DIFF
--- a/html/views/overlay/index.css
+++ b/html/views/overlay/index.css
@@ -1,7 +1,7 @@
 /* General look */
 html { height: 100%; width: 100%; } 
 body { font-weight: bold; margin: 0; padding: 0; overflow: hidden; } 
-div { #f9f9f9-space: nowrap; display: block; padding: 0; margin: 0; overflow: hidden; } 
+div { white-space: nowrap; display: block; padding: 0; margin: 0; overflow: hidden; } 
 .Wrapper { clear: both; box-sizing: border-box; } 
 .Box { float: left; margin: 0; padding: 0; position: relative; display: block; } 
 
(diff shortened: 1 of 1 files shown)


html/views/overlay/index.css
```diff
@@ -1,7 +1,7 @@
 /* General look */
 html { height: 100%; width: 100%; } 
 body { font-weight: bold; margin: 0; padding: 0; overflow: hidden; } 
-div { white-space: nowrap; display: block; padding: 0; margin: 0; overflow: hidden; } 
+div { #f9f9f9-space: nowrap; display: block; padding: 0; margin: 0; overflow: hidden; }
```
queekusme: Please revert this to 'white' as 'white-space' is not associated with the colour white

queekusme (changes requested):

html/views/overlay/index.css
yetzt: oops, thanks for spotting this.

yetzt:
can we get this merged? i would really like to be able to have a green team on stream.

queekusme (approved):
Whilst I haven't had time to lest this locally, I can't see anymore obvious concerns

frank-weinberg:
My plan is to merge this along with the other changes for 2027 when I am ready to create the first Beta.

yetzt:
thank you! <3

PR #939, opened 2026-04-18
Fix issue #938.  Handle spaces in install path.

### LINKED ISSUE #938: [BUG] Spaces in install directory path will crash startup shell scripts.
**Describe the bug**
scoreboard.sh and scoreboard-mac.command can crash if they are installed into a directory path with whitespace. 

**To Reproduce**
1. Create a directory named `/tmp/my scoreboard`.  Note the blank space after `my`.
2. install CRG into that directory
3. cd into your home directory
4. Run the scoreboard `/tmp/CRG\ scoreboard/crg-scoreboard_v2025.9/scoreboard.sh`
6. See error**Expected behavior**

The shell script will display the following and crash

```
cts@nomad ~ % /tmp/my\ scoreboard/crg-scoreboard_v2025.9/scoreboard.sh 
/tmp/my scoreboard/crg-scoreboard_v2025.9/scoreboard.sh: line 3: cd: /tmp
scoreboard/crg-scoreboard_v2025.9: No such file or directory
Error: Unable to access jarfile lib/crg-scoreboard.jar
cts@nomad ~ % 
```

**Version:**
v2025.9

 - OS and version: Linux and MacOS (most likely any version)
 - Browser N/A
 - Version 2025.9 (probably any version since Nov. 2019)

**If this issue occurred on a smartphone, tablet, or other device, please complete the following information:**
 - Device: N/A
 - OS: N/A
 - Browser N/A
 - Version N/A

**Additional context**

craig-spannring:
The quoting of the output from dirname was incorrect.  If there were any spaces in the path to scoreboard.sh/scoreboard-mac.command, the script wouldn't set it's working directory correctly and would fail to find the .jar file.

In addition, the ${GUI} was quoted, even when empty and could result in an empty parameter being passed into the java program.

### DIFF
--- a/scoreboard.sh
+++ b/scoreboard.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-cd "$(dirname $0)"
+loc="$(dirname "${0}")"
+cd "${loc}" || exit 1
+
 
 GUI="--gui"
 
@@ -13,4 +15,4 @@ if [ -x /usr/libexec/java_home ]; then
   JAVA="/usr/libexec/java_home -exec java"
 fi
 
-exec $JAVA -Done-jar.silent=true -Dorg.eclipse.jetty.server.LEVEL=WARN -jar lib/crg-scoreboard.jar "$GUI" "$@"
+exec $JAVA -Done-jar.silent=true -Dorg.eclipse.jetty.server.LEVEL=WARN -jar "${loc}/lib/crg-scoreboard.jar" ${GUI} "$@"
--- a/scoreboard.sh
+++ b/scoreboard.sh
@@ -15,4 +15,4 @@ if [ -x /usr/libexec/java_home ]; then
   JAVA="/usr/libexec/java_home -exec java"
 fi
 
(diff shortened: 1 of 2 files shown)


scoreboard.sh
```diff
@@ -13,4 +15,4 @@ if [ -x /usr/libexec/java_home ]; then
   JAVA="/usr/libexec/java_home -exec java"
 fi
 
-exec $JAVA -Done-jar.silent=true -Dorg.eclipse.jetty.server.LEVEL=WARN -jar lib/crg-scoreboard.jar "$GUI" "$@"
+exec $JAVA -Done-jar.silent=true -Dorg.eclipse.jetty.server.LEVEL=WARN -jar "${loc}/lib/crg-scoreboard.jar" ${GUI} "$@"
```
frank-weinberg, line 18: Given the guarded cd above the relative path here should be fine. But the absolute path should also be fine.

frank-weinberg: But thinking about it some more - internally CRG uses relative paths for everything, so having a relative path here means we fail early and with a relatively clear message if there is some quirk that causes the working dir to be wrong. Having an absolute path here means that in that scenario we get more obscure errors during runtime.

PR #940, opened 2026-04-18
Fix potential freeze & improve debug capabilities

### LINKED ISSUE #929: [BUG] All operator buttons (Start Jam / Stop / Timeout) become unresponsive after Undo

frank-weinberg:
- avoid a race that could freeze the scoreboard on stats export
- Catch & Log exceptions in every thread

### DIFF
--- a/src/com/carolinarollergirls/scoreboard/core/game/GameImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/game/GameImpl.java
@@ -429,6 +429,7 @@ public final class GameImpl extends ScoreBoardEventProviderImpl<Game> implements
             }
             if (!"Never".equals(get(LAST_FILE_UPDATE))) { set(LAST_FILE_UPDATE, "Pre Game"); }
         }
+        if (prop == FILENAME && jsonSnapshotter != null) { jsonSnapshotter.setFilename((String) value); }
         if (prop == OFFICIAL_SCORE && (boolean) value && source == Source.WS) {
             Clock pc = getClock(Clock.ID_PERIOD);
             Clock ic = getClock(Clock.ID_INTERMISSION);
--- a/src/com/carolinarollergirls/scoreboard/Main.java
+++ b/src/com/carolinarollergirls/scoreboard/Main.java
@@ -100,7 +100,7 @@ public final class Main extends Logger {
     }
 
     private void stop(Throwable ex) {
-        if (ex != null) { Logger.printStackTrace(ex); }
+        if (ex != null) { Logger.printStackTrace("main thread", ex); }
         Logger.printMessage(LocalDateTime.now().toString() + ": Fatal error.   Exiting in 15 seconds.");
         try {
             Thread.sleep(15000);
@@ -212,8 +212,7 @@ public final class Main extends Logger {
                     }
                 }
             } catch (IOException e) {
-                Logger.printMessage("Error looking for instance to import from:");
-                Logger.printStackTrace(e);
+                Logger.printStackTrace("looking for instance to import from", e);
                 Logger.printMessage("Skipping import");
                 return;
             }
@@ -240,10 +239,7 @@ public final class Main extends Logger {
             copyDir(sourcePath, targetPath, Paths.get("html", "custom"));
             copyDir(sourcePath, targetPath, Paths.get("html", "images"));
             copyDir(sourcePath, targetPath, Paths.get("html", "videos"));
-        } catch (IOException e) {
-            Logger.printMessage("Exception during importing: ");
-            Logger.printStackTrace(e);
-        }
+        } catch (IOException e) { Logger.printStackTrace("importing", e); }
     }
 
     private void createGui() {
--- a/src/com/carolinarollergirls/scoreboard/core/admin/MediaImpl.java
+++ b/src/com/carolinarollergirls/scoreboard/core/admin/MediaImpl.java
@@ -21,6 +21,7 @@ import com.carolinarollergirls.scoreboard.core.interfaces.Media;
 import com.carolinarollergirls.scoreboard.core.interfaces.ScoreBoard;
 import com.carolinarollergirls.scoreboard.event.ScoreBoardEventProviderImpl;
 import com.carolinarollergirls.scoreboard.utils.BasePath;
+import com.carolinarollergirls.scoreboard.utils.Logger;
 
 /**
  * Maintain list of media files and provide notifications when list changes.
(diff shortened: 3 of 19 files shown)


src/com/carolinarollergirls/scoreboard/core/game/GameImpl.java
```diff
@@ -429,6 +429,7 @@ protected void valueChanged(Value<?> prop, Object value, Object last, Source sou
             }
             if (!"Never".equals(get(LAST_FILE_UPDATE))) { set(LAST_FILE_UPDATE, "Pre Game"); }
         }
+        if (prop == FILENAME && jsonSnapshotter != null) { jsonSnapshotter.setFilename((String) value); }
```
a1ly404, line 432: @frank-weinberg so is this where the race condition could happen? If the jsonSnapshotter is null and we have a filename where a value is being changed? Do snapshots happen on every changed value?

frank-weinberg: No. The race was:

Thread 1: 
- Export is triggered from UI
- WS.java l.162 calls runInBatch
- ScoreBoardEventProviderImpl l.118 acquires the coreLock
- some unrelated code runs

Meanwhile Thread 2: 
- regular data update for Game X triggered from JSONStateManger l.76 
- JSONStateSnapshotter l.37 calls synchronized method writeFile
- Thread gets the lock associated with the synchronized
- JSONStateSnapshotter l.47 called game.getFilename() -> this place was changed
- GameImpl l.1174 calls get
- ScoreBoardEventProviderImpl l.451 tries to acquire the coreLock and blocks because Thread 1 holds the lock

Thread 1:
- GameImpl l.534 calls jsonSnapshotter.writeFile and blocks because the method is synchronized and Thread 2 holds the lock.

And now both threads are blocked waiting for each other. And since one of them holds the coreLock, no other thread can perform any meaningful action.

PR #980, opened 2026-08-23
Fix macOS java detection for homebrew

### LINKED ISSUE #979: [BUG] macOS launcher can't find Homebrew installed OpenJDK
**Describe the bug**
On macOS (tahoe 26.4), scoreboard.sh and scoreboard-mac.command fail to launch when java/jre was installed via homebrew (`brew install openjdk`) and not symlinked. Both scripts resolve the java exclusively through `/usr/libexec/java_home -exec java` on macOS, and that call hard-fails with a default brew install because homebrew's openjdk is keg-only (not registered with the system `java_home` by default). The script never falls back to a working java on `PATH` nor does it respoect a value set in `$JAVA_HOME`

>The operation couldn't be completed. Unable to locate a Java Runtime.
>Please visit http://www.java.com for information on installing Java.

**To Reproduce**
1. Install homebrew && `brew install openjdk`
2. Add to your zshrc with the suggested echo > zshrc
3. run ./scoreboard-mac.command
4. See error about not being able to locate jre

**Expected behavior**
Find and use any available java runtime, probably following this order:
1. $JAVA_HOME/bin/java if JAVA_HOME is set
2. java_home on macOS
3. java via PATH (covers homebrew, manual installs, or anything not registered/symlinked to java_home)

**Version:**
Current dev branch. ~2027.0 / 2026-07-03

 - OS and version: macOS 26.4
- Java: Installed via homebrew, keg-only, not registered via symlinks
  - ❯ which java: /opt/homebrew/opt/openjdk/bin/java
     ❯ java --version:
        openjdk 26.0.2.1 2026-08-18
        OpenJDK Runtime Environment Homebrew (build 26.0.2.1)
        OpenJDK 64-Bit Server VM Homebrew (build 26.0.2.1, mixed mode, sharing)

**Additional context**
- /usr/libexec/java_home won't list homebrews jdk unless it's symlinked into /Library/Java/JavaVirtualMachines

Reasons for not wanting to use java_home: no admin on the machine you're using, running multiple versions of jdk.

**Potential solution**
1. Check $JAVA_HOME: Run a -n (var non empty) on JAVA_HOME and a -x (file exists) on the executable. 
2. Check libexec: Run a -x on java_home, run the executable with -F (fail fast) and -n the output (it'll either be empty or the actual path)
3. Check PATH: use command -v against java

chelming:
Fixes #979 

Java path resolution:
1. JAVA_HOME env var
2. java_home
3. PATH

### DIFF
--- a/scoreboard.sh
+++ b/scoreboard.sh
@@ -9,10 +9,22 @@ GUI="--gui"
 # If fd 0 (stdin) exists, this is an interactive shell, so don't use the gui
 test -t "0" && GUI=""
 
-JAVA=java
-if [ -x /usr/libexec/java_home ]; then
+JAVA=""
+if [ -n "${JAVA_HOME}" ] && [ -x "${JAVA_HOME}/bin/java" ]; then
+  # An explicitly configured JDK takes priority
+  JAVA="${JAVA_HOME}/bin/java"
+elif [ -x /usr/libexec/java_home ] && [ -n "$(/usr/libexec/java_home -F 2>/dev/null)" ]; then
   # We're on OS X, which has its own way of doing Java
   JAVA="/usr/libexec/java_home -exec java"
+elif command -v java >/dev/null 2>&1; then
+  # Fall back to java on the PATH (covers Homebrew/OpenJDK installs
+  # that aren't registered with java_home)
+  JAVA="$(command -v java)"
+fi
+
+if [ -z "$JAVA" ]; then
+  echo "Could not find a Java runtime. Install one (e.g. brew install openjdk) and try again." >&2
+  exit 1
 fi
 
 exec $JAVA -Done-jar.silent=true -Dorg.eclipse.jetty.server.LEVEL=WARN -jar lib/crg-scoreboard.jar ${GUI} "$@"
--- a/scoreboard-mac.command
+++ b/scoreboard-mac.command
@@ -9,10 +9,22 @@ GUI="--gui"
 # If fd 0 (stdin) exists, this is an interactive shell, so don't use the gui
 test -t "0" && GUI=""
 
-JAVA=java
-if [ -x /usr/libexec/java_home ]; then
+JAVA=""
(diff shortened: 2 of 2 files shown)


Anusien:
I saw your comment about using `command` because it's built-in, but they appear to have different semantics. The problem is that `command` will tell how your shell will execute a command, which doesn't necessarily point to a file. It is arguably more correct, but I think wrong in this example:
```
$ alias java='/no-such-path/java'
$ which java
/usr/bin/java
command -v java
alias java='/no-such-path/java'
```
You could say following the alias is better, but you'd have to *parse* the alias. I'd just use `which`.


Also while it's true that `which` is a separate binary, I'd suspect that it exists everywhere. If you really wanted to be paranoid, you could try `which` first and fall back to `command`.

scoreboard.sh
```diff
@@ -9,10 +9,22 @@ GUI="--gui"
...
   # We're on OS X, which has its own way of doing Java
   JAVA="/usr/libexec/java_home -exec java"
+elif command -v java >/dev/null 2>&1; then
+  # Fall back to java on the PATH (covers Homebrew/OpenJDK installs
+  # that aren't registered with java_home)
+  JAVA="$(command -v java)"
+fi
+
+if [ -z "$JAVA" ]; then
+  echo "Could not find a Java runtime. Install one (e.g. brew install openjdk) and try again." >&2
```
Anusien, line 26: Nit: Presumably `scoreboard.sh` is for Linux and shouldn't mention `brew`.

Anusien:
Also if you wanted to do a little extra, you could mention setting `$JAVA_HOME` in the error message for not finding Java to help people understand how to fix it. Especially because there's a quirk where presumably installing openjdk doesn't update `JAVA_HOME` or `$PATH` in the current session. I'm not sure how concise you could get that though.

scoreboard.sh
chelming: valid

chelming:
updated

